PR #83, opened 2026-02-02
fix: prune finalized blocks and attestations

MegaRedHand:
Implements automatic pruning of finalized data to prevent unbounded storage growth. Adds a LiveChain index table to optimize fork choice operations while enabling efficient cleanup of finalized blocks, attestations, and signatures.

## Summary

When slots become finalized, they no longer participate in fork choice. This PR:
1. Adds a LiveChain index for fast fork choice (avoids Block deserialization)
2. Automatically prunes finalized data when checkpoints advance
3. Moves signature-related types to storage module for proper encapsulation

### LiveChain Index
- New `LiveChain` table: `(slot, root) -> parent_root` mappings (40 bytes per block)
- Big-endian slot encoding enables efficient range scans and early termination
- Refactored `compute_lmd_ghost_head` to use `HashMap<H256, (u64, H256)>` instead of full `Block`
- Automatic pruning removes finalized entries while preserving Blocks table for historical queries

### Automatic Pruning
- `update_checkpoints()` triggers pruning when finalization advances
- **LiveChain**: Prunes index entries for slots < finalized_slot (keeps finalized block itself)
- **GossipSignatures**: Removes signatures for finalized attestations (no longer needed for block building)
- **AggregatedPayloads**: Removes aggregated proofs for finalized attestations

## Implementation Details

**Why pruning on finalization?**
- Finalized attestations can't contribute to future fork choice decisions
- Gossip signatures are only needed for building new blocks on live forks
- Aggregated payloads from finalized blocks won't be reused

## Testing

Ran a validator node for 1 hour with these changes. Storage reached ~1GB with a downward slope (actively pruning) compared to ~5GB before pruning was implemented.

Also tested via existing spec tests:
- `forkchoice_spectests.rs`: Fork choice with LiveChain index
- `signature_spectests.rs`: Block building with new signature types
- `stf_spectests.rs`: State transition with pruned storage

### DIFF
--- a/crates/blockchain/tests/forkchoice_spectests.rs
+++ b/crates/blockchain/tests/forkchoice_spectests.rs
@@ -9,7 +9,7 @@ use ethlambda_storage::{Store, backend::InMemoryBackend};
 use ethlambda_types::{
     attestation::Attestation,
     block::{Block, BlockSignatures, BlockWithAttestation, SignedBlockWithAttestation},
-    primitives::{H256, TreeHash, VariableList},
+    primitives::{H256, VariableList, ssz::TreeHash},
     state::State,
 };
 


crates/storage/src/store.rs
```diff
@@ -567,22 +768,46 @@ impl Store {
...
+    /// # Thread Safety
+    ///
+    /// This method uses a read-modify-write pattern that is NOT atomic:
+    /// 1. Read existing payloads
+    /// 2. Append new payload
+    /// 3. Write back
+    ///
+    /// Concurrent calls could result in lost updates. This method MUST be called
+    /// from a single thread. In ethlambda, the Store is owned by the BlockChain
+    /// actor which provides single-threaded access.
```
```suggestion
    /// from a single thread. In our case, that thread is the `BlockChain` `GenServer`
```

PR #114, opened 2026-02-11
fix: improve block fetch logic

### LINKED ISSUE #89: Improve pending block fetching from P2P
We currently retry BlocksByRoot requests 5 times before stopping in the P2P layer. If a block fails to be retrieved, eventually a child block will be received (unless that chain stops extending), which will cause `BlockChain` to fetch the block from P2P. This is inefficient, since we already have that block, and we should instead be fetching the missing link.

MegaRedHand:
Closes #89

This PR bundles some improvements to block fetch logic found during checkpoint-sync testing.

2be39051899196cad56ca831c8f932f385beedd8: includes an optimization to the unknown block fetch mechanism in `BlockChainServer`.
While testing checkpoint-sync we found the previous block search never ended, so we sped it up a bit by storing the oldest missing ancestor of each block.
Instead of going block by block and resetting after a failed request (the previous behavior of ethlambda), we skip the blocks we already fetched and fetch the missing ancestor.

cf75dcc6e9aa96d661f66f0b79face09b1abb5e2: fixes block fetch logic in the P2P module. Before, we considered empty responses as succeded, so we didn't retry. Now we follow the same logic as for error responses: we retry after a backoff.

7032bf1dda7037af3b988803cc850a822c651908: increases the number of retries to 10, with 5ms initial backoff and an exponent of 2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
+      # NOTE: we download the latest fixtures since generating them in CI fails
       - name: Download test fixtures
         env:
           GH_TOKEN: ${{ github.token }}


crates/blockchain/src/lib.rs
```diff
@@ -283,14 +294,24 @@ impl BlockChainServer {
         if !self.store.has_state(&parent_root) {
             info!(%slot, %parent_root, %block_root, "Block parent missing, storing as pending");
 
+            // Resolve the actual missing ancestor: if the parent is itself pending,
+            // reuse its cached missing ancestor. Otherwise the parent IS the missing block.
+            let missing_root = self
```
pablodeymo: I think this search should be recursive.

PR #116, opened 2026-02-11
feat: store pending blocks in db

MegaRedHand:
This PR addresses possible OOM issues during long chain syncs.

Previously, we stored pending blocks in memory.

Now, we store them in the DB, reducing the memory footprint. We still keep some metadata in memory, but it shouldn't bring problems on reasonable sync scenarios.


crates/blockchain/src/lib.rs
```diff
@@ -304,11 +304,37 @@ impl BlockChainServer {
...
+            // Walk up through DB: if missing_root is already stored from a previous
+            // session, the actual missing block is further up the chain.
+            while let Some(header) = self.store.get_block_header(&missing_root) {
+                if self.store.has_state(&header.parent_root) {
+                    // Parent state available — process from DB, cascade handles the rest
+                    let block = self
+                        .store
+                        .get_signed_block(&missing_root)
+                        .expect("signed block exists when header exists");
+                    self.on_block(block);
```
pablodeymo: We could change this recursion in favor of a different approach, to avoid theoretical stack overflows.

PR #141, opened 2026-02-23
Add --attestation-committee-count CLI parameter

pablodeymo:
## Motivation

`ATTESTATION_COMMITTEE_COUNT` was hardcoded to `1` in `crates/blockchain/src/lib.rs`, meaning all validators shared a single attestation subnet. Future devnets will use multiple committees (other clients like Zeam and Lantern already support this parameter), so the value needs to be configurable at runtime.

## Description

Add a `--attestation-committee-count` CLI argument (default `1`) that controls how many attestation subnets exist. Validators are assigned to subnets via `validator_id % attestation_committee_count`.

**Changes:**
- **`bin/ethlambda/src/main.rs`**: Add `--attestation-committee-count` CLI argument with default value `1`, pass it to `start_p2p()`
- **`crates/blockchain/src/lib.rs`**: Remove the `ATTESTATION_COMMITTEE_COUNT` constant (no longer needed since the value comes from CLI)
- **`crates/net/p2p/src/lib.rs`**: Accept `attestation_committee_count` as a function parameter instead of importing the constant, use it for subnet ID calculation

The hardcoded constant is removed entirely since clap provides the default. The blockchain crate doesn't store the value because it has no internal use for it — only the P2P layer needs it for subnet subscription.

## How to test

- Run with explicit value: `--attestation-committee-count 1` (should behave identically to before)
- Run without the flag (default `1`, same behavior)
- `make lint` and `make test` pass with no warnings

### DIFF
--- a/.claude/skills/test-pr-devnet/scripts/test-branch.sh
+++ b/.claude/skills/test-pr-devnet/scripts/test-branch.sh
@@ -62,7 +62,9 @@ if ! docker info &>/dev/null; then
     exit 1
 fi
 
-if [[ ! -d "$ETHLAMBDA_ROOT/.git" ]]; then
+# Use `git rev-parse` instead of `-d .git` to support git worktrees,
+# where .git is a file (not a directory) pointing to the main repo.
+if ! git -C "$ETHLAMBDA_ROOT" rev-parse --git-dir &>/dev/null; then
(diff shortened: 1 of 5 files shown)


pablodeymo:
@greptile-apps review it again please

PR #157, opened 2026-02-26
Split on_block() and on_gossip_attestation() into verified and unverified variants

pablodeymo:
## Motivation

PR #113 proposed extracting signature verification out of `on_block()` into a separate `precheck_block_signatures()` function called by the network layer before processing. Four automated reviewers flagged a critical security flaw: removing `verify_signatures()` from `on_block()` means any caller that forgets to call `precheck_block_signatures()` first can import blocks without cryptographic validation. There are 3 call sites for `on_block()` (production `process_block()`, and 2 test files), and only 1 would get the precheck.

PR #154 responded by taking the conservative approach: keep `on_block()` unchanged and document its validate-then-mutate structure. It argued against extraction.

This PR is the compromise: instead of removing verification from `on_block()`, we **keep `on_block()` as the safe default** (always verifies) and add `on_block_without_verification()` as an explicit opt-out. This follows Zeam's architecture where `ForkChoice.onBlock()` is signature-agnostic and verification happens at the chain layer, but inverted — verification is the default and opting out requires calling a function whose name makes the risk obvious.

This eliminates the `skip-signature-verification` compile-time feature flag entirely.

### `store.rs` — Split `on_block()`

Three functions replace the old `on_block()`:

| Function | Visibility | Verifies? | Stores sigs? | Used by |
|----------|-----------|-----------|-------------|---------|
| `on_block()` | public | Yes | Yes | Production (`lib.rs`), `signature_spectests` |
| `on_block_without_verification()` | public | No | No | `forkchoice_spectests` |
| `on_block_core()` | private | — | — | Shared logic for both |

`on_block_core()` takes `parent_state` by value (avoids the previous `.clone()` — minor perf win). The 4-line dedup + parent-state preamble is duplicated in both public functions to avoid double deserialization (verification needs the state, core needs the state).

### `store.rs` — Split `on_gossip_attestation()`

Same pattern:

| Function | Visibility | Verifies? | Stores sigs? |
|----------|-----------|-----------|-------------|
| `on_gossip_attestation()` | public | Yes | Yes |
| `on_gossip_attestation_without_verification()` | public | No | No |

Both share the validation + `on_attestation()` + logging logic. Some duplication is acceptable to keep each function self-contained.

### `Cargo.toml` — Remove feature flag

- Removed `[features]` section with `skip-signature-verification`
- Removed `required-features = ["skip-signature-verification"]` from forkchoice test entry

### `forkchoice_spectests.rs`

- Changed `store::on_block()` → `store::on_block_without_verification()`
- Fixed 4 pre-existing `collapsible_if` clippy warnings (now surfaced because the test no longer requires a feature flag to compile, so `clippy --all-targets` sees it)

### `Makefile`

- Removed `--features skip-signature-verification` from `test` target

### `CLAUDE.md` and `RELEASE.md`

- Removed all `skip-signature-verification` references from test commands, gotchas, and Docker build examples

## Relationship to #113 and #78

This PR supersedes #113's approach. Rather than extracting verification to a separate precheck (which creates the "forgotten precheck" security risk flagged by reviewers), it splits the function into safe and unsafe variants. The safe variant (`on_block()`) always verifies — callers can't accidentally skip verification. The unsafe variant (`on_block_without_verification()`) makes the risk explicit in its name. Issue #78's concern about signature verification is fully addressed: production paths always verify, and the compile-time feature flag is eliminated.

## How to test

```bash
make fmt
make lint
make test
cargo test -p ethlambda-blockchain --test forkchoice_spectests
cargo test -p ethlambda-blockchain --test signature_spectests --release
```

Confirm no feature flag remnants:
```bash
grep -r "skip-signature-verification" crates/
```

Note: `signature_spectests` have pre-existing failures (fixture deserialization issue with `ProposerSignature`). These failures are identical on `main` and are unrelated to this change.

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -473,7 +473,9 @@ fn on_block_core(
         );
     }
 
-    // Process proposer attestation (enters "new" stage, not "known")
+    // Process proposer attestation as if received via gossip
+    // The proposer's attestation should NOT affect this block's fork choice position.
+    // It is treated as pending until interval 3 (end of slot).
     // TODO: validate attestations before processing
     let _ = on_attestation(store, proposer_attestation, false)
         .inspect(|_| metrics::inc_attestations_valid("gossip"))


Makefile
```diff
@@ -9,10 +9,10 @@ fmt: ## 🎨 Format all code using rustfmt
 lint: ## 🔍 Run clippy on all workspace crates
 	cargo clippy --workspace --all-targets -- -D warnings
 
-test: leanSpec/fixtures ## 🧪 Run all tests, then forkchoice tests with skip-signature-verification
+test: leanSpec/fixtures ## 🧪 Run all tests, then forkchoice spec tests
 	# Tests need to be run on release to avoid stack overflows during signature verification/aggregation
 	cargo test --workspace --release
-	cargo test -p ethlambda-blockchain --features skip-signature-verification --test forkchoice_spectests
+	cargo test -p ethlambda-blockchain --test forkchoice_spectests
```
MegaRedHand: This was the case only due to the feature flag, so it's no longer needed.


```suggestion
```

crates/blockchain/src/store.rs
```diff
@@ -414,11 +461,7 @@ pub fn on_block(
     // to prevent the proposer from gaining circular weight advantage.
     update_head(store);
 
-    // Process proposer attestation as if received via gossip
-    // The proposer's attestation should NOT affect this block's fork choice position.
-    // It is treated as pending until interval 3 (end of slot).
```
MegaRedHand: We should keep this comment

PR #182, opened 2026-03-03
Replace aggregated payload RocksDB tables with in-memory circular buffers

pablodeymo:
## Motivation

When finalization stalls, ethlambda's attestation-related tables grow without bound, causing OOM crashes. In our devnet observations, memory climbs to ~11.5 GB and crashes in ~27-minute cycles. The root cause:

**`LatestKnownAggregatedPayloads` and `LatestNewAggregatedPayloads`** grow without bound — they are append-only with no eviction. Each gossip message with 9 validators creates 9 copies of the same ~256 KB proof, and the read-merge-write pattern in RocksDB keeps accumulating entries.

Pruning was gated on finalization advancing. When finalization stalls (which is exactly when memory pressure is highest), nothing gets cleaned up.

## Solution

In-memory circular buffers for aggregated payloads, inspired by [Lantern](https://github.com/Pier-Two/lantern) (C client).

Replace the two RocksDB aggregated payload tables with `VecDeque`-based buffers wrapped in `Arc<Mutex<...>>`, with FIFO eviction:

- **Known payloads** (fork-choice-active): capped at **4096 entries** (~30 min of history at 1 attestation/slot with 9 validators)
- **New payloads** (pending, drained every interval): capped at **512 entries**

**Why in-memory buffers are safe:**
- Aggregated payloads are **ephemeral** — rebuilt from gossip on restart. There's no durability requirement.
- Only the `BlockChain` actor reads/writes them (P2P and RPC clones never access these tables).

**Why `Arc<Mutex<...>>`:**
- `Store` is cloned 3× at startup (P2P, RPC, BlockChain actors). Wrapping buffers in `Arc<Mutex>` ensures all clones share the same buffer state, matching the existing `Arc<dyn StorageBackend>` sharing pattern.

**What changes for each method:**

| Method | Before (RocksDB) | After (buffer) |
|--------|-------------------|----------------|
| `insert_new_aggregated_payload` | Read-merge-write DB | `self.new_payloads.lock().unwrap().push(key, payload)` |
| `insert_known_aggregated_payload` | Read-merge-write DB | `self.known_payloads.lock().unwrap().push(key, payload)` |
| `iter_known_aggregated_payloads` | Full table scan + deserialize + `.collect()` | `self.known_payloads.lock().unwrap().grouped().into_iter()` |
| `iter_new_aggregated_payloads` | Full table scan + deserialize + `.collect()` | `self.new_payloads.lock().unwrap().grouped().into_iter()` |
| `promote_new_aggregated_payloads` | Read new → merge into known → delete new (40 lines) | `drain new → push_batch to known` (2 lines) |

### `crates/storage/src/store.rs`
- **Added `PayloadBuffer` struct** — `VecDeque<(SignatureKey, StoredAggregatedPayload)>` with configurable capacity. Methods: `push` (FIFO eviction), `push_batch`, `drain`, `grouped` (HashMap by key), `unique_keys` (HashSet).
- **Wrapped `new_payloads` / `known_payloads`** in `Arc<Mutex<PayloadBuffer>>` in the `Store` struct, so clones share buffer state.
- **Separate capacity constants**: `AGGREGATED_PAYLOAD_CAP` (4096) for known, `NEW_PAYLOAD_CAP` (512) for new.
- **Removed `Default` derive** from `PayloadBuffer` — a capacity-0 buffer violates the `entries.len() <= capacity` invariant.
- **Refactored all 9 public aggregated payload methods** to use in-memory buffers instead of RocksDB operations.
- **Removed 5 private DB helper methods** that are no longer needed: `iter_aggregated_payloads`, `iter_aggregated_payload_keys`, `insert_aggregated_payload`, `insert_aggregated_payloads_batch`, `prune_aggregated_payload_table`.
- **Removed aggregated payload pruning** from `update_checkpoints` (FIFO handles it now).
- **Added test helpers** `Store::test_store()` and `Store::test_store_with_backend()` to avoid exposing `PayloadBuffer` in block/state pruning tests.
- **Added 5 tests**: FIFO eviction, grouped(), drain(), promotion, and cloned store shares buffers.

### `crates/storage/src/api/tables.rs`
- Removed `LatestNewAggregatedPayloads` and `LatestKnownAggregatedPayloads` from `Table` enum (10 → 8 tables).

### `crates/storage/src/backend/rocksdb.rs`
- Removed `cf_name` match arms for the deleted table variants.

### `crates/blockchain/src/store.rs`
- Updated doc comments referencing removed table names.

## How to Test

```bash
make fmt    # Formatting passes
make lint   # Clippy with -D warnings passes
cargo test --workspace --release   # All 91 tests pass (including 5 new ones)
```

All 26 fork choice spec tests pass unchanged.

For production validation, deploy to a devnet and monitor memory usage during a finalization stall. Memory should plateau instead of climbing linearly.

### DIFF
--- a/crates/storage/src/store.rs
+++ b/crates/storage/src/store.rs
@@ -1,5 +1,5 @@
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::sync::{Arc, LazyLock};
+use std::sync::{Arc, LazyLock, Mutex};
 
 /// The tree hash root of an empty block body.
 ///
@@ -90,19 +90,21 @@ const _: () = assert!(
     "BLOCKS_TO_KEEP must be >= STATES_TO_KEEP"
 );
 
-/// Hard cap for each aggregated payload buffer (new and known).
+/// Hard cap for the known aggregated payload buffer.
 /// Matches Lantern's approach. With 9 validators, this holds
 /// ~455 unique attestation messages (~30 min at 1/slot).
 const AGGREGATED_PAYLOAD_CAP: usize = 4096;
 
-/// Attestation data retention window in slots (~4.3 min at 4s/slot).
-const ATTESTATION_RETENTION_SLOTS: u64 = 64;
+/// Hard cap for the new (pending) aggregated payload buffer.
+/// Smaller than known since new payloads are drained every interval (~4s).
+/// With 9 validators at 1 attestation/slot, one interval holds ~9 entries.
+const NEW_PAYLOAD_CAP: usize = 512;
 
 /// Fixed-size circular buffer for aggregated payloads.
 ///
 /// Entries are evicted FIFO when the buffer reaches capacity.
 /// This prevents unbounded memory growth when finalization stalls.
-#[derive(Clone, Default)]
+#[derive(Clone)]
 struct PayloadBuffer {
     entries: VecDeque<(SignatureKey, StoredAggregatedPayload)>,
     capacity: usize,
@@ -202,8 +204,8 @@ fn decode_live_chain_key(bytes: &[u8]) -> (u64, H256) {
 #[derive(Clone)]
 pub struct Store {
     backend: Arc<dyn StorageBackend>,
-    new_payloads: PayloadBuffer,
-    known_payloads: PayloadBuffer,
+    new_payloads: Arc<Mutex<PayloadBuffer>>,
+    known_payloads: Arc<Mutex<PayloadBuffer>>,
 }
 
 impl Store {
@@ -341,8 +343,8 @@ impl Store {
 
         Self {
             backend,
-            new_payloads: PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP),
-            known_payloads: PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP),
+            new_payloads: Arc::new(Mutex::new(PayloadBuffer::new(NEW_PAYLOAD_CAP))),
+            known_payloads: Arc::new(Mutex::new(PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP))),
         }
     }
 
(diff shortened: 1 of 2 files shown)


crates/storage/src/store.rs
```diff
@@ -505,40 +567,18 @@ impl Store {
...
-            }
-            if !deletes.is_empty() {
-                batch.delete_batch(table, deletes).expect("delete");
-            }
-            batch.commit().expect("commit");
+    /// Prune stale gossip signatures and attestation data by slot age.
+    ///
+    /// Independent of finalization — prevents unbounded growth when finalization stalls.
+    /// Returns (pruned_sigs, pruned_att_data).
+    pub fn prune_attestation_data_by_age(&mut self, current_slot: u64) -> (usize, usize) {
```
MegaRedHand: I think we don't need this. The circular buffer should be enough

crates/storage/src/store.rs
```diff
@@ -276,7 +339,11 @@ impl Store {
 
         info!(%anchor_state_root, %anchor_block_root, "Initialized store");
 
-        Self { backend }
+        Self {
+            backend,
+            new_payloads: PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP),
```
MegaRedHand: I think we should use a lower cap for new payloads

crates/storage/src/store.rs
```diff
@@ -1280,6 +1201,8 @@ mod tests {
         let backend = Arc::new(InMemoryBackend::new());
         let mut store = Store {
             backend: backend.clone(),
+            new_payloads: PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP),
+            known_payloads: PayloadBuffer::new(AGGREGATED_PAYLOAD_CAP),
```
MegaRedHand: Why do we need these changes? I think we should refactor this in another PR

PR #185, opened 2026-03-05
Migrate spawned-concurrency from 0.4 to 0.5.0-rc

pablodeymo:
## Motivation

The `spawned-concurrency` crate is releasing v0.5.0-rc with a new protocol/actor macro API that replaces the GenServer pattern. This migration adopts the new API, which provides:

- **Type-safe message dispatch** via `#[protocol]` traits instead of manual enum matching
- **Simpler handler signatures** with `#[send_handler]` instead of `handle_cast` + `CastResponse::NoReply`
- **Synchronous fire-and-forget sends** — `ActorRef::send()` returns `Result<(), ActorError>` immediately (no `.await`)
- **Cleaner timer API** — `send_after` takes `Context<A>` instead of cloning the handle

### Dependency changes

| File | Change |
|------|--------|
| `Cargo.toml` (workspace) | `spawned-concurrency` and `spawned-rt`: `"0.4"` → git dep at `lambdaclass/spawned.git` tag `v0.5.0-rc` |
| `bin/ethlambda/Cargo.toml` | Removed `spawned-concurrency` and `spawned-rt` (unused — binary only uses `BlockChain` from `ethlambda_blockchain`) |

### Actor migration (`crates/blockchain/src/lib.rs`)

**Before (0.4 — GenServer pattern):**
- `CastMessage` enum with 4 variants (`NewBlock`, `NewAttestation`, `NewAggregatedAttestation`, `Tick`)
- `impl GenServer for BlockChainServer` with `handle_cast` matching on the enum
- `GenServerHandle<BlockChainServer>` for the wrapper
- `handle.cast(CastMessage::X).await` for sending messages
- `send_after(duration, handle.clone(), CastMessage::Tick)` for timer scheduling

**After (0.5 — Protocol/Actor pattern):**
- `#[protocol] trait BlockChainProtocol` defining 4 send-only methods
- `#[actor(protocol = BlockChainProtocol)] impl BlockChainServer` with individual `#[send_handler]` methods
- `ActorRef<BlockChainServer>` for the wrapper
- `handle.new_block(block)` (sync, returns `Result<(), ActorError>`)
- `send_after(duration, ctx.clone(), block_chain_protocol::Tick)` for timer scheduling

The `#[protocol]` macro generates a `block_chain_protocol` module with message structs (`Tick`, `NewBlock`, `NewAttestation`, `NewAggregatedAttestation`) and implements the protocol methods directly on `ActorRef<BlockChainServer>`.

### Wrapper simplification

`notify_*` methods on `BlockChain` changed from `async fn(&mut self)` to `fn(&self)`:
- No `.await` needed — sends are non-blocking
- `&self` instead of `&mut self` — `ActorRef::send()` takes `&self`

### P2P call site updates

Removed `.await` from all 4 call sites in the P2P layer:

| File | Method |
|------|--------|
| `crates/net/p2p/src/gossipsub/handler.rs` | `notify_new_block` |
| `crates/net/p2p/src/gossipsub/handler.rs` | `notify_new_aggregated_attestation` |
| `crates/net/p2p/src/gossipsub/handler.rs` | `notify_new_attestation` |
| `crates/net/p2p/src/req_resp/handlers.rs` | `notify_new_block` |

## Migration guide improvements

While migrating, we identified several gaps in the upstream `spawned-concurrency` migration documentation. These are suggestions for the `spawned` maintainers:

### 1. `send_after` / `send_interval` API changes (undocumented)

The migration guide doesn't mention that `send_after` changed from taking `GenServerHandle<A>` to `Context<A>`. Every user of timers will hit this.

```rust
// 0.4: send_after takes GenServerHandle
send_after(duration, handle.clone(), MyMessage::Tick);

// 0.5: send_after takes Context
// Inside a handler (ctx available):
send_after(duration, ctx.clone(), Tick);
// Outside a handler (from ActorRef):
send_after(duration, actor_ref.context(), Tick);
```

### 2. Self-rescheduling tick pattern (undocumented)

A very common pattern is an actor that reschedules itself with variable delays. The guide should show this before/after:

```rust
// 0.4:
CastMessage::Tick => {
    self.do_work();
    send_after(next_delay, handle.clone(), CastMessage::Tick);
}

// 0.5:
#[send_handler]
async fn handle_tick(&mut self, _msg: Tick, ctx: &Context<Self>) {
    self.do_work();
    send_after(next_delay, ctx.clone(), Tick);
}
```

### 3. `ActorRef::context()` method (undocumented)

The guide shows `ActorRef` but doesn't document that `actor_ref.context()` converts an `ActorRef<A>` to a `Context<A>`. This is needed when scheduling timers from outside a handler (e.g., in `spawn()`).

### 4. `&mut self` → `&self` for send callers (undocumented)

Switching from `.cast().await` (async) to direct protocol method calls (sync `Result<(), ActorError>`) means callers no longer need `&mut self`. This can simplify wrapper structs and their call sites significantly.

### 5. Generated module naming convention (incomplete)

The guide mentions generated structs but doesn't explicitly state the naming rule:
- `#[protocol]` on `trait FooBarProtocol` generates module `foo_bar_protocol` (snake_case)
- Message structs: `foo_bar_protocol::MethodName` (PascalCase of method name)
- Type-erased ref: `FooBarProtocolRef = Arc<dyn FooBarProtocol>`

### 6. Mixed send + request protocols (missing example)

The guide shows send-only and request-only examples separately. A before/after for a protocol with **both** send and request methods would be helpful.

### 7. `#[handler]` vs `#[send_handler]` vs `#[request_handler]` (unclear)

The "Escape Hatches" section mentions `Handler<M>` but doesn't explain the difference between the three handler attributes:
- `#[request_handler]` — for request-response methods
- `#[send_handler]` — for fire-and-forget methods
- `#[handler]` — for manual `Message` types not in a protocol (?)

### 8. Wrapper pattern migration (missing)

A common pattern is wrapping `ActorRef` in a public struct to hide internals. The guide should show this before/after:

```rust
// 0.4:
pub struct MyService { handle: GenServerHandle<MyServer> }
impl MyService {
    pub async fn do_thing(&mut self, data: String) {
        let _ = self.handle.cast(MyMsg::DoThing(data)).await;
    }
}

// 0.5:
pub struct MyService { handle: ActorRef<MyServer> }
impl MyService {
    pub fn do_thing(&self, data: String) {
        let _ = self.handle.do_thing(data);
    }
}
```

### 9. `spawned-rt` changes (undocumented)

The migration guide focuses on `spawned-concurrency` but doesn't mention whether `spawned-rt` has breaking changes.

### 10. `#[started]` panic behavior (unclear)

In 0.4, `.start()` returned the handle directly. In 0.5, `ActorStart::start()` also returns `ActorRef<A>` directly, but what happens if the `#[started]` hook panics? The guide should clarify whether subsequent sends get `ActorError::ActorStopped`.

## How to test

```bash
make fmt    # cargo fmt --all
make lint   # clippy with -D warnings
make test   # all workspace tests + spec tests
```

All pass locally. For end-to-end verification, run a local devnet:
```bash
make run-devnet
```

Verify that:
- Blocks are produced and imported
- Attestations are gossiped and processed
- Justification and finalization advance
- Tick scheduling works correctly (800ms intervals)

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -55,6 +55,7 @@ impl BlockChain {
         validator_keys: HashMap<u64, ValidatorSecretKey>,
         is_aggregator: bool,
     ) -> BlockChain {
+        metrics::set_is_aggregator(is_aggregator);
         let genesis_time = store.config().genesis_time;
         let key_manager = key_manager::KeyManager::new(validator_keys);
         let handle = BlockChainServer {
--- a/crates/blockchain/src/key_manager.rs
+++ b/crates/blockchain/src/key_manager.rs
@@ -6,6 +6,8 @@ use ethlambda_types::{
     signature::{ValidatorSecretKey, ValidatorSignature},
 };
 
+use crate::metrics;
+
 /// Error types for KeyManager operations.
 #[derive(Debug, thiserror::Error)]
 pub enum KeyManagerError {
@@ -100,9 +102,13 @@ impl KeyManager {
             .get_mut(&validator_id)
             .ok_or(KeyManagerError::ValidatorKeyNotFound(validator_id))?;
 
-        let signature: ValidatorSignature = secret_key
-            .sign(slot, message)
-            .map_err(|e| KeyManagerError::SigningError(e.to_string()))?;
+        let signature: ValidatorSignature = {
+            let _timing = metrics::time_pq_sig_attestation_signing();
+            secret_key
+                .sign(slot, message)
+                .map_err(|e| KeyManagerError::SigningError(e.to_string()))
+        }?;
+        metrics::inc_pq_sig_attestation_signatures();
 
         // Convert ValidatorSignature to XmssSignature (FixedVector<u8, SignatureSize>)
         let sig_bytes = signature.to_bytes();
--- a/crates/blockchain/src/metrics.rs
+++ b/crates/blockchain/src/metrics.rs
@@ -79,35 +79,29 @@ pub fn set_node_start_time() {
 }
 
 /// Increment the valid attestations counter.
-pub fn inc_attestations_valid(source: &str) {
-    static LEAN_ATTESTATIONS_VALID_TOTAL: std::sync::LazyLock<IntCounterVec> =
+pub fn inc_attestations_valid() {
+    static LEAN_ATTESTATIONS_VALID_TOTAL: std::sync::LazyLock<IntCounter> =
         std::sync::LazyLock::new(|| {
-            register_int_counter_vec!(
+            register_int_counter!(
                 "lean_attestations_valid_total",
-                "Count of valid attestations",
-                &["source"]
+                "Total number of valid attestations"
             )
             .unwrap()
         });
-    LEAN_ATTESTATIONS_VALID_TOTAL
-        .with_label_values(&[source])
-        .inc();
+    LEAN_ATTESTATIONS_VALID_TOTAL.inc();
 }
 
 /// Increment the invalid attestations counter.
-pub fn inc_attestations_invalid(source: &str) {
-    static LEAN_ATTESTATIONS_INVALID_TOTAL: std::sync::LazyLock<IntCounterVec> =
+pub fn inc_attestations_invalid() {
+    static LEAN_ATTESTATIONS_INVALID_TOTAL: std::sync::LazyLock<IntCounter> =
         std::sync::LazyLock::new(|| {
-            register_int_counter_vec!(
+            register_int_counter!(
                 "lean_attestations_invalid_total",
-                "Count of invalid attestations",
-                &["source"]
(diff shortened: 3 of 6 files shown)


crates/blockchain/src/lib.rs
```diff
@@ -489,60 +484,61 @@ impl BlockChainServer {
...
 
-#[derive(Clone, Debug)]
-enum CastMessage {
-    NewBlock(SignedBlockWithAttestation),
-    NewAttestation(SignedAttestation),
-    NewAggregatedAttestation(SignedAggregatedAttestation),
-    Tick,
+#[protocol]
+pub(crate) trait BlockChainProtocol: Send + Sync {
+    fn new_block(&self, block: SignedBlockWithAttestation) -> Result<(), ActorError>;
```
MegaRedHand: Is this a call or cast?

PR #198, opened 2026-03-11
Remove aggregated signature verification bypass for small proof_data

pablodeymo:
## Motivation

During a security review, we identified that `verify_aggregated_signature()` in `crates/common/crypto/src/lib.rs` silently returned `Ok(())` for any `proof_data` under 10 bytes — including empty proofs — without performing any cryptographic verification.

This was reachable from two production P2P network paths:
1. **Gossip aggregated attestations** (`store.rs:459`) — any peer can send a `SignedAggregatedAttestation` with empty `proof_data`. No proposer key required.
2. **Block attestation signatures** (`store.rs:1197`) — a malicious proposer can include attestations with fabricated `aggregation_bits` and empty `proof_data`.

A malicious peer could forge attestation votes to manipulate fork choice (LMD-GHOST head selection) and influence justification/finalization thresholds.

Note: proposer individual XMSS signatures are NOT affected — they use `ValidatorSignature::is_valid()`, a separate code path with no size-based bypass.

### Origin of the bypass

`AggregatedSignatureProof::empty()` creates proofs with empty `proof_data` for the `verify=false` test path. The `< 10` guard was likely added to prevent SSZ deserialization errors when these empty proofs reached the verifier during development. However, the `verify=false` path never calls `verify_signatures`, so the guard was unnecessary — and it converted what should be an error into a silent success on the production `verify=true` path.

## Description

Remove the early-return bypass in `verify_aggregated_signature()`:

```rust
// REMOVED:
if proof_data.len() < 10 {
    return Ok(());
}
```

Without this guard, empty/small `proof_data` will naturally fail at the SSZ deserialization step (`Devnet2XmssAggregateSignature::from_ssz_bytes`) and return `VerificationError::DeserializationFailed`.

## How to test

All spec tests pass without the bypass:

```bash
cargo test -p ethlambda-blockchain --test forkchoice_spectests --release     # 26/26 passed
cargo test -p ethlambda-state-transition --test stf_spectests --release      # 14/14 passed
cargo test -p ethlambda-blockchain --test signature_spectests --release      #  8/8 passed
```

## Related

- `StoreError::ParticipantsMismatch` is defined at `store.rs:899` but never raised. Could be addressed in a follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,5 @@ Support for older devnet releases is discontinued when the next devnet version i
 Some features we are looking to implement in the near future, in order of priority:
 
 - [Add support for pq-devnet-4](https://github.com/lambdaclass/ethlambda/issues/155)
-- [Use spawned for our P2P event loop](https://github.com/lambdaclass/ethlambda/issues/165) - [WIP](https://github.com/lambdaclass/ethlambda/pull/188)
-- [Observability: more metrics from leanMetrics](https://github.com/lambdaclass/ethlambda/issues/76)
 - [RPC endpoints for chain data consumption](https://github.com/lambdaclass/ethlambda/issues/75)
 - [Add guest program and ZK proving of the STF](https://github.com/lambdaclass/ethlambda/issues/156)


MegaRedHand (approved):
Great catch!! 🎣

PR #203, opened 2026-03-12
fix: use updated latest_finalized.slot for justifiability checks

shariqnaiyer:
https://github.com/leanEthereum/leanSpec/pull/443

The above PR in leanSpec removed the use of the original_finalized_slot snapshot and changed it to using the updated latest_finalized.slot from state. A difference in this implementation within clients leads to some state root mismatches at certain blocks. 

I am making this PR so we can get consistency across clients prior to the devnet run in about 6-8 hours. Please feel to request changes, merge or close this PR.


MegaRedHand (approved):
Thanks!

PR #235, opened 2026-03-17
Add proto-array for incremental fork choice

### LINKED ISSUE #167: Cache fork-choice weights in `LiveChain`
We currently discard computed weights after each head update. We should instead cache them in our `LiveChain` and update it only with new attestations.

For reference, Lighthouse uses a "proto-array" implementation: https://github.com/sigp/lighthouse/tree/stable/consensus/proto_array/src
while Prysm uses a doubly-linked tree: https://github.com/OffchainLabs/prysm/tree/develop/beacon-chain/forkchoice/doubly-linked-tree

pablodeymo:
## Summary

- Introduce `ProtoArray` + `VoteTracker` in the fork_choice crate for incremental LMD-GHOST head computation
- Replace the O(validators × chain_depth) per-call recomputation in `update_head` with delta-based weight propagation: O(changed_votes) + O(nodes) single backward pass + O(depth) head lookup
- `update_safe_target` unchanged — runs once/slot with a different attestation pool, not worth optimizing in this phase
- Debug oracle (`#[cfg(debug_assertions)]`) runs both implementations in parallel and asserts equality

## Design

- `ForkChoiceState` struct wraps `ProtoArray` + `VoteTracker`, owned by `BlockChainServer`
- Initialized from live chain at startup via `ForkChoiceState::from_store()`
- `on_block_core` registers new blocks and prunes on finalization
- `update_head` uses proto-array; falls back to spec impl for tree visualization (`log_tree=true`, 1x/slot)
- 13 unit tests for proto-array (linear chains, forks, tiebreaks, pruning, vote changes)

## Test plan

- [x] `make fmt` — clean
- [x] `make lint` — clean
- [x] All 15 fork_choice unit tests pass (13 new + 2 existing)
- [x] All 27 fork choice spec tests pass (debug oracle validates every `update_head` call)
- [x] All 8 signature spec tests pass
- [ ] Devnet validation

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -1,6 +1,7 @@
 use std::collections::{HashMap, HashSet};
 
 use ethlambda_crypto::aggregate_signatures;
+use ethlambda_fork_choice::{ProtoArray, VoteTracker};
 use ethlambda_state_transition::{
     is_proposer, process_block, process_slots, slot_is_justifiable_after,
 };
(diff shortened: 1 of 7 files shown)


crates/blockchain/src/store.rs
```diff
@@ -346,12 +409,12 @@ pub fn on_tick(
                 }
             }
             3 => {
-                // Update safe target for validators
+                // Update safe target for validators (uses spec implementation, not proto-array)
```
MegaRedHand: Can we use proto-array here too?

MegaRedHand:
Closing this due to conflicts. We might continue this work in the future, though

PR #261, opened 2026-04-07
Advance XMSS key preparation window before signing

pablodeymo:
## Summary

- Expose `is_prepared_for(slot)` and `advance_preparation()` on `ValidatorSecretKey`, delegating to the leansig `SignatureSchemeSecretKey` trait
- Before signing in `KeyManager::sign_message()`, check if the target slot is within the prepared window and advance if needed
- Return a descriptive error if the key's activation interval is fully exhausted
- Add a timing test for `advance_preparation()` (526ms in release mode on Apple Silicon)

## Root Cause

XMSS keys use a Top-Bottom Tree Traversal scheme where only two consecutive bottom trees are loaded in memory at any time. Each bottom tree covers `sqrt(LIFETIME) = 2^16 = 65,536` slots, so the prepared window spans `131,072` slots (~6 days at 4s/slot).

The leansig library provides `advance_preparation()` to slide this window forward by computing the next bottom tree, but ethlambda's `KeyManager` never called it. When the devnet at `admin@ethlambda-1` reached slot 131,072, all 4 nodes panicked simultaneously:

```
Signing: key not yet prepared for this epoch, try calling sk.advance_preparation.
```

The fix checks the prepared interval before every sign operation and advances the window on demand. This is a lazy approach — `advance_preparation` is called at signing time rather than proactively in the tick loop — because:
- It happens once every ~3 days (65,536 slots)
- The computation (one bottom tree of hash leaves) takes ~526ms in release mode
- It keeps the change minimal and avoids tick-loop complexity

## Test plan

- [x] `make fmt` clean
- [x] `make lint` clean
- [x] `make test` passes (existing tests use small lifetimes or skip verification)
- [x] `test_advance_preparation_duration` passes (`cargo test -p ethlambda-types test_advance_preparation_duration --release -- --ignored --nocapture`)
- [ ] Deploy to devnet with fresh genesis and verify it runs past slot 131,072 without panic

### DIFF
--- a/crates/blockchain/src/key_manager.rs
+++ b/crates/blockchain/src/key_manager.rs
@@ -105,11 +105,13 @@ impl KeyManager {
 
         // Advance XMSS key preparation window if the slot is outside the current window.
         // Each bottom tree covers 65,536 slots; the window holds 2 at a time.
+        // Multiple advances may be needed if the node was offline for an extended period.
         if !secret_key.is_prepared_for(slot) {
             info!(validator_id, slot, "Advancing XMSS key preparation window");
             while !secret_key.is_prepared_for(slot) {
+                let before = secret_key.get_prepared_interval();
                 secret_key.advance_preparation();
-                if !secret_key.is_prepared_for(slot) {
+                if secret_key.get_prepared_interval() == before {
                     return Err(KeyManagerError::SigningError(format!(
                         "XMSS key exhausted for validator {validator_id}: \
                          slot {slot} is beyond the key's activation interval"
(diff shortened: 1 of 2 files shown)


crates/blockchain/src/key_manager.rs
```diff
@@ -102,6 +103,21 @@ impl KeyManager {
             .get_mut(&validator_id)
             .ok_or(KeyManagerError::ValidatorKeyNotFound(validator_id))?;
 
+        // Advance XMSS key preparation window if the slot is outside the current window.
+        // Each bottom tree covers 65,536 slots; the window holds 2 at a time.
+        if !secret_key.is_prepared_for(slot) {
```
MegaRedHand: #262

MegaRedHand (changes requested):
The while loop is redundant

PR #269, opened 2026-04-09
feat: formalize `slot_is_justifiable_after` in Lean4

MegaRedHand:
## Introduction

This PR introduces a Lean4 package under `formal/`, formalizing `slot_is_justifiable_after` and proving some properties like "there's always another justifiable slot" and "the distance between justifiable slots increases as the delta increases".

First step of #272

## Lean4 package structure

```text
formal/
    EthLambda/
    EthLambdaProofs/
    lean-ffi/
```

The new Lean4 package has two libraries: `EthLambda` and `EthLambdaProofs`. The first one contains the actual implementations called from Rust, using bounded types like `UInt64`. `EthLambdaProofs`, on the other hand, contains theorems proving some properties of the implementation.

Apart from those two, there's an additional directory with a Rust crate wrapping the Lean FFI glue.


formal/lean-ffi/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+    });
+}
+
+/// Check if `delta` is justifiable under 3SF-mini rules.
+///
+/// Calls the formally verified Lean4 implementation via FFI.
+pub fn justifiable(delta: u64) -> bool {
+    init_lean();
+    unsafe { lean_justifiable(delta) != 0 }
+}
```
MegaRedHand, line 46: Not sure we need this

MegaRedHand:
ZK proving the STF is in the roadmap, and moving to Lean4 would get in the way of that

PR #270, opened 2026-04-09
fix: self-deliver aggregator attestations to gossip_signatures

MegaRedHand:
## Summary

- Aggregator now calls `on_gossip_attestation()` locally before publishing to gossip, ensuring its own validator's attestation enters `gossip_signatures` and is included in aggregated proofs
- Adds `publish_best_effort()` to suppress the expected `NoPeersSubscribedToTopic` warning when the aggregator publishes to attestation subnets it subscribes to
- Threads `is_aggregator` from `SwarmConfig` through `BuiltSwarm` into `P2PServer` so the gossipsub handler can use best-effort publishing

## Context

Gossipsub does not deliver messages back to the sender. The aggregator subscribes to the attestation subnet and publishes via mesh, but in small devnets no other node subscribes to attestation subnets. This causes the aggregator's own attestation publish to fail silently with `NoPeersSubscribedToTopic`, meaning its vote never enters `gossip_signatures`, is never aggregated, and is never included in blocks.

The reference spec handles this explicitly in `_produce_attestations()`:

```python
# Process attestation locally before publishing.
#
# Gossipsub does not deliver messages back to the sender.
# Without local processing, the aggregator node never sees its own
# validator's attestation in gossip_signatures, reducing the
# aggregation count below the 2/3 safe-target threshold.
self.sync_service.store = self.sync_service.store.on_gossip_attestation(
    signed_attestation=signed_attestation,
    is_aggregator=is_aggregator_role,
)
```

## Test plan

- [x] `cargo build` passes
- [x] `cargo clippy --workspace -- -D warnings` passes
- [x] `cargo test --workspace --release --lib` passes (all 32 unit tests)
- [ ] Deploy to devnet and verify:
  - Validator 3's attestation appears in "Attestation processed" logs on the aggregator
  - Blocks from all proposers include aggregator's vote in aggregated proofs (higher attestation counts)
  - No more `NoPeersSubscribedToTopic` warnings on the aggregator node
  - Finalization continues normally

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -198,6 +198,17 @@ impl BlockChainServer {
                 signature,
             };
 
+            // Self-deliver: store our own attestation locally for aggregation.
+            // Gossipsub does not deliver messages back to the sender, so without
+            // this the aggregator never sees its own validator's signature in
+            // gossip_signatures and it is excluded from aggregated proofs.
+            if self.is_aggregator {
+                let _ = store::on_gossip_attestation(&mut self.store, signed_attestation.clone())
+                    .inspect_err(|err| {
+                        warn!(%slot, %validator_id, %err, "Self-delivery of attestation failed")
+                    });
+            }
+
             // Publish to gossip network
             if let Some(ref p2p) = self.p2p {
                 let _ = p2p.publish_attestation(signed_attestation).inspect_err(
--- a/crates/net/p2p/src/gossipsub/handler.rs
+++ b/crates/net/p2p/src/gossipsub/handler.rs
@@ -142,8 +142,16 @@ pub async fn publish_attestation(server: &mut P2PServer, attestation: SignedAtte
         .cloned()
         .unwrap_or_else(|| attestation_subnet_topic(subnet_id));
 
-    // Publish to the attestation subnet topic
-    server.swarm_handle.publish(topic, compressed);
+    // Publish to the attestation subnet topic.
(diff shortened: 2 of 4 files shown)


pablodeymo (commented):
**Style nit (non-blocking):** The `if let Err` pattern here works, but the codebase convention prefers `inspect_err` for side-effect-only error handling. One option:
```rust
let result = swarm.behaviour_mut().gossipsub.publish(topic, data);
if let Err(err) = &result {
    if !(ignore_no_peers && matches!(err, PublishError::NoPeersSubscribedToTopic)) {
        warn!(%err, "Swarm adapter: publish failed");
    }
}
```
Purely cosmetic — current version reads fine too.

crates/blockchain/src/lib.rs
```diff
@@ -198,6 +198,17 @@ impl BlockChainServer {
                 signature,
             };
 
+            // Self-deliver: store our own attestation locally for aggregation.
+            // Gossipsub does not deliver messages back to the sender, so without
+            // this the aggregator never sees its own validator's signature in
+            // gossip_signatures and it is excluded from aggregated proofs.
+            if self.is_aggregator {
+                let _ = store::on_gossip_attestation(&mut self.store, signed_attestation.clone())
```
pablodeymo: can we avoid this clone? we can do this in another PR

MegaRedHand: I think we can: https://github.com/lambdaclass/ethlambda/issues/271

pablodeymo (approved):

PR #275, opened 2026-04-10
Migrate from Poseidon2 to Poseidon1 signature scheme

pablodeymo:
## Motivation

Poseidon2 is not cryptographically safe. The upstream leanSpec, leansig, and leanMultisig libraries have already migrated to Poseidon1. This PR aligns ethlambda with the new scheme.

### Dependency changes
- Bump `leansig` to devnet4 branch (Poseidon1 internally)
- Bump `lean-multisig` to devnet4 HEAD (`fd88140`)
- Add `leansig_wrapper` for `XmssPublicKey`/`XmssSignature` types
- Remove `ethereum_ssz` from crypto crate (replaced by postcard+lz4 serialization)
- Bump `rand` from 0.9 to 0.10

### Signature scheme change
- Switch from `SIGTopLevelTargetSumLifetime32Dim64Base8` (V=64, TopLevelPoseidon) to `SchemeAbortingTargetSumLifetime32Dim46Base8` (V=46, AbortingHypercube)
- `SIGNATURE_SIZE` updated from 3112 to 2536 bytes

### API adaptation in ethlambda-crypto
- `xmss_aggregate_signatures` -> `xmss_aggregate` (with children + log_inv_rate params)
- `xmss_verify_aggregated_signatures` -> `xmss_verify_aggregation`
- `Devnet2XmssAggregateSignature` -> `AggregatedXMSS`
- SSZ serialization -> `AggregatedXMSS::serialize()`/`deserialize()`

### leanSpec bump
- Bump to latest HEAD (includes Poseidon1 migration + tiebreaker fix)
- Remove `patches/` directory (no longer needed)
- Adapt test harness for new fixture format:
  - Resolve `headRootLabel` from block registry when explicit root not provided
  - `justifiedSlots` changed from `Vec<u64>` to `Vec<bool>` (proper bitlist)
  - `SignedBlock.message` renamed to `SignedBlock.block` in fixtures
- CI: generate prod keys locally (upstream keys still Poseidon2)

### Cross-client tests
- Ream test vectors marked `#[ignore]` pending Poseidon1 vectors from ream team

## How to Test

```bash
make fmt
make lint
make test
```


.github/workflows/ci.yml
```diff
@@ -81,28 +81,6 @@ jobs:
...
-        id: cache-prod-keys
-        uses: actions/cache@v4
-        with:
-          path: leanSpec/packages/testing/src/consensus_testing/test_keys/prod_scheme
-          key: prod-keys-${{ steps.prod-keys-url.outputs.hash }}
-
-      - name: Download production keys
-        if: steps.cache-fixtures.outputs.cache-hit != 'true' && steps.cache-prod-keys.outputs.cache-hit != 'true'
-        working-directory: leanSpec
-        run: uv run python -m consensus_testing.keys --download --scheme prod
```
MegaRedHand, line 104: We should keep this

crates/common/crypto/src/lib.rs
```diff
@@ -307,10 +297,9 @@ mod tests {
     }
 
     // ── Cross-client XMSS compatibility tests (ream) ────────────────────
-    // Verify that signatures produced by the ream client (Rust) can be
-    // decoded and verified by ethlambda.
-    // Test vectors were produced by ream (slot 5, all-zeros message) and
-    // are also used in leanSpec PR #433 for Python-side cross-client testing.
+    // TODO: Update these test vectors with Poseidon1 signatures from the ream client.
+    // The current vectors were generated with Poseidon2 and are no longer valid.
```
MegaRedHand: Signature spectests should suffice, I think. We can remove these tests.

crates/common/crypto/Cargo.toml
```diff
@@ -12,13 +12,9 @@ version.workspace = true
...
-lean-multisig = { git = "https://github.com/leanEthereum/leanMultisig.git", rev = "e4474138487eeb1ed7c2e1013674fe80ac9f3165" }
-# rec_aggregation is needed to access the config types (LeanSigPubKey, LeanSigSignature)
-# which are not re-exported by lean-multisig
-rec_aggregation = { git = "https://github.com/leanEthereum/leanMultisig.git", rev = "e4474138487eeb1ed7c2e1013674fe80ac9f3165" }
-# ethereum_ssz is needed for Devnet2XmssAggregateSignature from lean-multisig,
-# which implements the old ethereum_ssz Encode/Decode traits
-ethereum_ssz = { package = "ethereum_ssz", version = "0.10.0" }
+lean-multisig = { git = "https://github.com/leanEthereum/leanMultisig.git", rev = "b09d5e7618cc5a32ef4a6c9009034a19778d7b39" }
+# leansig_wrapper provides XmssPublicKey/XmssSignature types used by lean-multisig's public API
+leansig_wrapper = { git = "https://github.com/leanEthereum/leanMultisig.git", rev = "b09d5e7618cc5a32ef4a6c9009034a19778d7b39" }
```
MegaRedHand: We should check if this is still needed

patches/fix-lexicographic-tiebreaker-test.patch
```diff
@@ -1,76 +0,0 @@
-diff --git a/tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py b/tests/consensus/devnet/fc/test_lexicographic_tiebreaker.py
```
MegaRedHand, line 1: What's this?

Cargo.toml
```diff
@@ -55,7 +55,7 @@ prometheus = "0.14"
 clap = { version = "4.3", features = ["derive", "env"] }
 
 # XMSS signatures
-leansig = { git = "https://github.com/leanEthereum/leanSig.git", rev = "73bedc26ed961b110df7ac2e234dc11361a4bf25" }
+leansig = { git = "https://github.com/leanEthereum/leanSig", branch = "devnet4" }
```
MegaRedHand, line 58: We should pin the commit

.github/workflows/ci.yml
pablodeymo: Restored the download + cache block. Added `continue-on-error: true` to the download step with a local key generation fallback, since the upstream keys are still Poseidon2. Once leansig-test-keys publishes Poseidon1 keys, the download will work and skip the fallback.

crates/common/crypto/src/lib.rs
pablodeymo: Agreed. Removed the cross-client ream tests entirely - signature spectests cover verification.

crates/common/crypto/Cargo.toml
pablodeymo: Yes, still needed. `lean-multisig` uses `XmssPublicKey`/`XmssSignature` from `leansig_wrapper` in its public API (`xmss_aggregate`, `xmss_verify_aggregation`) but doesn't re-export them. We need the direct dependency to access these types.

patches/fix-lexicographic-tiebreaker-test.patch
pablodeymo: This was a temporary patch to fix a leanSpec bug in the lexicographic tiebreaker test (multi-level forks validated at the wrong level). It was applied after cloning leanSpec at 9c30436. Now that we bumped leanSpec to HEAD, the fix is included upstream and the patch is deleted.

Cargo.toml
pablodeymo: We can't pin with `rev` here because `lean-multisig` internally depends on `leansig` via `branch = "devnet4"`. Cargo treats `?rev=X` and `?branch=devnet4` as different sources even when they resolve to the same commit, causing type incompatibility. Using `branch = "devnet4"` to match lean-multisig's specifier is the only way to unify them. The lockfile pins the actual commit (`5cc7e37`).

MegaRedHand (approved):

.github/workflows/ci.yml
```diff
@@ -102,6 +102,12 @@ jobs:
         if: steps.cache-fixtures.outputs.cache-hit != 'true' && steps.cache-prod-keys.outputs.cache-hit != 'true'
         working-directory: leanSpec
         run: uv run python -m consensus_testing.keys --download --scheme prod
+        continue-on-error: true
+
+      - name: Generate production keys (fallback if download unavailable)
+        if: steps.cache-fixtures.outputs.cache-hit != 'true' && steps.cache-prod-keys.outputs.cache-hit != 'true'
+        working-directory: leanSpec
+        run: uv run python -m consensus_testing.keys --scheme prod --count 12 --max-slot 100
```
```suggestion
```

PR #284, opened 2026-04-14
feat: add proposer proof aggregation

MegaRedHand:
This PR adds true attestation aggregation on the proposer block building pipeline. As part of this, the attestation byte cap we added was removed since we now have the 16 attestation+proof cap.

### DIFF
--- a/crates/common/crypto/src/lib.rs
+++ b/crates/common/crypto/src/lib.rs
@@ -35,11 +35,14 @@ pub enum AggregationError {
     #[error("public key count ({0}) does not match signature count ({1})")]
     CountMismatch(usize, usize),
 
-    #[error("aggregation panicked")]
-    AggregationPanicked,
-
     #[error("proof size too big: {0} bytes")]
     ProofTooBig(usize),
+
+    #[error("child proof deserialization failed at index {0}")]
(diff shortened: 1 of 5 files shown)


crates/common/crypto/src/lib.rs
```diff
@@ -92,19 +95,130 @@ pub fn aggregate_signatures(
...
+    if raw_public_keys.len() != raw_signatures.len() {
+        return Err(AggregationError::CountMismatch(
+            raw_public_keys.len(),
+            raw_signatures.len(),
+        ));
+    }
+
+    // Need at least one raw signature OR at least 2 children to merge.
+    if raw_public_keys.is_empty() && children.len() < 2 {
+        if children.is_empty() {
```
pablodeymo: this inner if is not necessary

pablodeymo (approved):

crates/common/crypto/src/lib.rs
MegaRedHand: Removed in [a9b8e06](https://github.com/lambdaclass/ethlambda/pull/284/commits/a9b8e06e92717e9da8df5d47cb19524aaa348c98)

PR #290, opened 2026-04-15
Bound gossip_signatures with FIFO eviction to prevent unbounded growth

### LINKED ISSUE #263: gossip_signatures grows unbounded during finalization stall
## Summary

The `gossip_signatures` HashMap in `crates/storage/src/store.rs:265` has no size cap. It is only pruned when finalization advances (`prune_gossip_signatures` at line 607, called from `update_checkpoints` at line 508). During a finalization stall, every gossip attestation from every validator at every slot is inserted and never removed.

## Observed Impact

In the devnet4 test_1 run (2026-04-07/08, ~18.5h), finalization stalled at slot 10733 for 6+ hours. During this period:
- Node 0 (aggregator) accumulated attestations indefinitely, with `attestation_count` in produced blocks growing monotonically from ~5 to **797**
- `gossip_signatures` entries piled up across all unfinalized slots
- This contributed to the RocksDB file descriptor exhaustion crash on node 0

## Root Cause

`gossip_signatures` relies entirely on finalization-based pruning:

```rust
pub fn prune_gossip_signatures(&mut self, finalized_slot: u64) -> usize {
    let mut gossip = self.gossip_signatures.lock().unwrap();
    gossip.retain(|_, entry| entry.data.slot > finalized_slot);
    // ...
}
```

When finalization stops, nothing is ever evicted. With 4-6 validators producing attestations every 4 seconds, this is ~5,400 entries/hour, each carrying a `ValidatorSignature` (~3 KB XMSS signature).

By contrast, `known_payloads` (cap=512) and `new_payloads` (cap=64) are correctly bounded with FIFO eviction via `PayloadBuffer`.

## Suggested Fix

Add a hard cap to `gossip_signatures` with FIFO or slot-based eviction, similar to the `PayloadBuffer` pattern used for aggregated payloads. This ensures bounded memory usage regardless of finalization progress.

pablodeymo:
## Motivation

Fixes #263. During a finalization stall, `gossip_signatures` grows without limit because it is only pruned when finalization advances. In the devnet4 test run (2026-04-07/08, ~18.5h), finalization stalled at slot 10733 for 6+ hours. During this period, every gossip attestation from every validator was inserted and never removed, with `attestation_count` in produced blocks growing from ~5 to 797. This contributed to the RocksDB file descriptor exhaustion crash on node 0.

## Description

Replaces the unbounded `HashMap<H256, GossipDataEntry>` with a new `GossipSignatureBuffer` struct that enforces a hard cap on total individual signatures, following the same FIFO eviction pattern already used by `PayloadBuffer` for aggregated payloads.

### New `GossipSignatureBuffer` struct

| Field | Type | Purpose |
|-------|------|---------|
| `data` | `HashMap<H256, GossipDataEntry>` | Signatures grouped by attestation data root |
| `order` | `VecDeque<H256>` | FIFO insertion order of data_roots |
| `capacity` | `usize` | Hard cap on `total_signatures` |
| `total_signatures` | `usize` | Running count (O(1), was O(n) sum) |

### Eviction algorithm

```
On insert(data_root, validator_id, signature):
  1. If (validator_id, data_root) already exists → skip (first-write-wins)
  2. Insert signature into BTreeMap (ascending order for XMSS)
  3. total_signatures += 1
  4. If new data_root → push to back of VecDeque
  5. While total_signatures > capacity:
       Pop front of VecDeque (oldest data_root)
       Remove entire entry from HashMap
       total_signatures -= removed.signatures.len()
```

### Capacity

`GOSSIP_SIGNATURE_CAP = 2048` individual signatures:
- With 4 validators: covers ~512 slots (~34 minutes)
- Worst-case memory: ~6 MB (2048 × 3KB XMSS signatures)
- Matches the scale of `AGGREGATED_PAYLOAD_CAP = 512`

### What stays the same

- **Finalization-based pruning** — still runs when finalization advances (complementary, not replaced)
- **Interval-2 deletion** — `delete_gossip_signatures()` still removes consumed sigs after aggregation
- **BTreeMap ordering** — ascending validator_id for XMSS aggregate correctness
- **First-write-wins dedup** — duplicate (validator_id, data_root) silently ignored
- **Public API** — all 5 `Store` methods keep their exact signatures and return types
- **Zero caller changes** — `blockchain/src/store.rs` compiles unchanged

### Performance note

The `delete()` method collects emptied data_roots during iteration and batch-cleans the VecDeque with a single `retain()` pass, avoiding per-root O(n) scans.

### Tests added

6 new unit tests covering:
1. FIFO eviction when over capacity
2. First-write-wins dedup
3. Multiple validators per data_root
4. Delete cleans up empty data_roots from both HashMap and VecDeque
5. Slot-based pruning with count verification
6. Eviction removes entire data_root (all its signatures)

## How to Test

```bash
make fmt    # passes
make lint   # passes (clippy clean)
cargo test -p ethlambda-storage --lib --release  # 22 tests pass (6 new + 16 existing)
```

Devnet validation: run a multi-client devnet, observe `lean_gossip_signatures` metric stays bounded even if finalization stalls.

## Closes

Closes #263

### DIFF
--- a/crates/storage/src/store.rs
+++ b/crates/storage/src/store.rs
@@ -270,7 +270,7 @@ impl GossipSignatureBuffer {
 
     /// Insert a gossip signature, FIFO-evicting oldest data_roots when over capacity.
     ///
-    /// First-write-wins: if (validator_id, data_root) already exists, the signature is ignored.
+    /// Last-write-wins: if (validator_id, data_root) already exists, the signature is overwritten.
     fn insert(
         &mut self,
         hashed: HashedAttestationData,
@@ -280,12 +280,11 @@ impl GossipSignatureBuffer {
         let (data_root, att_data) = hashed.into_parts();
 
         if let Some(entry) = self.data.get_mut(&data_root) {
-            // First-write-wins: skip if this validator already has a sig
-            if entry.signatures.contains_key(&validator_id) {
-                return;
-            }
+            let is_new = !entry.signatures.contains_key(&validator_id);
             entry.signatures.insert(validator_id, signature);
-            self.total_signatures += 1;
+            if is_new {
+                self.total_signatures += 1;
+            }
         } else {
             let mut signatures = BTreeMap::new();
             signatures.insert(validator_id, signature);
@@ -1802,7 +1801,7 @@ mod tests {
     }
 
     #[test]
-    fn gossip_buffer_dedup_first_write_wins() {
+    fn gossip_buffer_dedup_last_write_wins() {
         let mut buf = GossipSignatureBuffer::new(100);
         let data = make_att_data(1);
         let hashed = HashedAttestationData::new(data);
@@ -1810,6 +1809,7 @@ mod tests {
         buf.insert(hashed.clone(), 0, make_dummy_sig());
         buf.insert(hashed.clone(), 0, make_dummy_sig());
 
+        // Last-write-wins: overwrites the signature but count stays at 1
         assert_eq!(buf.total_signatures(), 1);
         assert_eq!(buf.len(), 1);
     }
--- a/crates/storage/src/store.rs
+++ b/crates/storage/src/store.rs
@@ -88,6 +88,11 @@ const AGGREGATED_PAYLOAD_CAP: usize = 512;
 /// Smaller than known since new payloads are drained every interval (~4s).
 const NEW_PAYLOAD_CAP: usize = 64;
 
+/// Hard cap for the gossip signature buffer (individual signatures, not distinct data_roots).
+/// With 4 validators and 4-second slots, 2048 signatures covers ~512 slots (~34 min).
+/// Each XMSS signature is ~3KB, so worst-case memory is ~6 MB.
+const GOSSIP_SIGNATURE_CAP: usize = 2048;
+
 /// An entry in the payload buffer: attestation data + set of proofs.
 #[derive(Clone)]
 struct PayloadEntry {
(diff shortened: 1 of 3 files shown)


crates/storage/src/store.rs
```diff
@@ -235,12 +240,150 @@ struct GossipDataEntry {
...
 
+/// Bounded buffer for gossip signatures with FIFO eviction.
+///
+/// Groups signatures by attestation data (via data_root). Each distinct
+/// attestation message stores the full `AttestationData` plus individual
+/// validator signatures in ascending order (required for XMSS aggregation).
+///
+/// Entries are evicted FIFO (by insertion order of the data_root) when
+/// total_signatures exceeds capacity, matching the `PayloadBuffer` pattern.
+struct GossipSignatureBuffer {
```
MegaRedHand, line 254: Don't we already have other FIFO buffers? Can we reuse that code?

crates/storage/src/store.rs
```diff
@@ -235,12 +240,150 @@ struct GossipDataEntry {
...
+        validator_id: u64,
+        signature: ValidatorSignature,
+    ) {
+        let (data_root, att_data) = hashed.into_parts();
+
+        if let Some(entry) = self.data.get_mut(&data_root) {
+            // First-write-wins: skip if this validator already has a sig
+            if entry.signatures.contains_key(&validator_id) {
+                return;
+            }
```
MegaRedHand: We should keep the latest sig of this validator

crates/storage/src/store.rs
pablodeymo: The APIs diverge enough that a generic would add more complexity than it saves — `PayloadBuffer` has `drain`/`extract_latest_attestations` while `GossipSignatureBuffer` needs `delete`/`prune`/`snapshot`, and the entry types are different (`Vec<Proof>` vs `BTreeMap<u64, Signature>`). The FIFO eviction loop itself is ~8 lines. Keeping them separate is intentional.

crates/storage/src/store.rs
pablodeymo: Done — switched to last-write-wins. Now overwrites the signature without double-counting `total_signatures`.

crates/storage/src/store.rs
```diff
@@ -235,12 +240,149 @@ struct GossipDataEntry {
...
+        &mut self,
+        hashed: HashedAttestationData,
+        validator_id: u64,
+        signature: ValidatorSignature,
+    ) {
+        let (data_root, att_data) = hashed.into_parts();
+
+        if let Some(entry) = self.data.get_mut(&data_root) {
+            let is_new = !entry.signatures.contains_key(&validator_id);
+            entry.signatures.insert(validator_id, signature);
```
```suggestion
            let is_new = entry.signatures.insert(validator_id, signature).is_none();
```

MegaRedHand (approved):

PR #299, opened 2026-04-16
fix: make signature aggregation asynchronous

### LINKED ISSUE #291: Cap aggregation on aggregator and proposer
We're seeing many skipped slots due to aggregation taking too long. This is exacerbated on the aggregator, which takes so long to aggregate that on many slots it fails to participate.

Some ideas:

- Aggregate signatures in the background
- Limit the number of proofs and signatures we aggregate at a time

MegaRedHand:
## Summary
- Moves committee-signature aggregation off the `BlockChainServer` actor thread onto a `tokio::task::spawn_blocking` worker. The actor no longer blocks for the duration of the XMSS proofs (previously 400-1200 ms per slot).
- Bounds aggregation with a `CancellationToken` fired by a 750 ms deadline (self-message via `send_after`). Aggregates produced inside the window are streamed back via `AggregateProduced` actor messages and published immediately; post-deadline aggregates still get applied locally and published.
- `#[stopped]` lifecycle hook cancels the in-flight worker and joins it (bounded at 2 s) so node shutdown is clean; new-session start joins any straggler from the previous slot and warns if it hadn't finished yet.

## Architecture
- `crates/blockchain/src/store.rs` now exposes pure helpers: `snapshot_aggregation_inputs`, `aggregate_job`, `apply_aggregated_group`, `finalize_aggregation_session`. The old `aggregate_committee_signatures` / `try_aggregate` (single inline function) are gone.
- `store::on_tick` no longer calls into aggregation — the actor drives it.
- New messages (`pub(crate)`): `AggregateProduced`, `AggregationDone`, `AggregationDeadline`. Late messages are fenced by a session id (the slot number).
- Per-aggregate work on the actor: insert into `new_payloads`, delete consumed gossip sigs, publish via gossipsub. End-of-session work (gauge refresh) is batched in `AggregationDone` to avoid 2N lock acquisitions per slot.

## Observations from devnet runs
Under a 4-node devnet with three induced partitions (via `docker pause`) across 200 slots:
- Finality lag: 3 slots (vs 5 under a previous synchronous cap)
- Sessions cancelled at deadline: 24 / 203 (11.8%) — only when `groups_considered > 1`
- Missed slots: 13, all inside partition windows
- No `Prior aggregation worker still running` warnings — worker wind-down is clean
- Chain recovers finalization after each partition

## Test plan
- [x] `cargo check --workspace`
- [x] `make fmt`, `make lint` (clippy clean)
- [x] `cargo test -p ethlambda-blockchain --test forkchoice_spectests` — 77/77 pass
- [x] 50-slot local devnet with 2 ethlambda + 1 zeam + 1 grandine — chain finalizes normally
- [x] 200-slot local devnet with induced partitions — consensus safety preserved, liveness recovers between partitions
- [ ] Review on CI

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2083,6 +2083,7 @@ dependencies = [
  "spawned-concurrency 0.5.0",
  "thiserror 2.0.18",
  "tokio",
+ "tokio-util",
  "tracing",
 ]
 
--- a/crates/blockchain/Cargo.toml
+++ b/crates/blockchain/Cargo.toml
@@ -22,6 +22,7 @@ ethlambda-types.workspace = true
 spawned-concurrency.workspace = true
 
 tokio.workspace = true
+tokio-util = { version = "0.7", default-features = false }
 
 rayon.workspace = true
 thiserror.workspace = true
--- a/crates/blockchain/src/aggregation.rs
+++ b/crates/blockchain/src/aggregation.rs
@@ -82,9 +82,6 @@ pub(crate) struct AggregationSession {
     /// Handle to the `spawn_blocking` worker. Held so `stopped()` / new-session
     /// start can await completion.
     pub(crate) worker: tokio::task::JoinHandle<()>,
-    /// Kept alive so the timer is implicitly cancelled when the field is
-    /// replaced or the actor stops (see `spawned_concurrency::tasks::time`).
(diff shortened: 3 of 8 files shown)


MegaRedHand:
Blocked by #304

pablodeymo (approved):

PR #302, opened 2026-04-17
fix: drop HashTreeRoot from XMSS-containing SSZ types

MegaRedHand:
## Summary

- Removes `HashTreeRoot` derive from `SignedAttestation`, `SignedBlock`, and `BlockSignatures` — the three types that embed `XmssSignature`.
- Routes those types through a new `run_serialization_only_test` helper in the SSZ spec-test harness. Encoding and round-trip are still enforced; the hash-tree-root check is skipped for them.

## Why

leanSpec's `Signature` container is SSZ-serialized as fixed bytes (via `is_fixed_size=True`, for cross-client wire compatibility with Ream) but its hash tree root is computed as a Container (merkleized fields). Our Rust `XmssSignature` is `SszVector<u8, SIGNATURE_SIZE>`, whose hash tree root uses `merkleize(pack_bytes(...))` — identical serialization, different root.

Since `SignedAttestation`, `SignedBlock`, and `BlockSignatures` are never hashed directly in production code (only their inner `Block` / `AttestationData` are), dropping the `HashTreeRoot` derive is cheaper than mirroring leanSpec's container layout. If any of them gets hashed later, we'll re-derive and match.

Fixes 4 failing `test_consensus_containers` SSZ tests after the leanSpec bump to `e9ddede`:

- `test_block_signatures_empty`
- `test_block_signatures_with_attestation`
- `test_signed_attestation_minimal`
- `test_signed_block_minimal`

## Test plan

- [x] `cargo test -p ethlambda-types --release --test ssz_spectests` passes (112/112)
- [ ] CI is green

## Follow-ups

- Two stacked PRs follow this one: #<PR2> adapts the fork-choice harness to the new fixture format, #<PR3> exercises real signature verification end-to-end.

### DIFF
--- a/crates/common/types/src/attestation.rs
+++ b/crates/common/types/src/attestation.rs
@@ -35,7 +35,11 @@ pub struct AttestationData {
 }
 
 /// Validator attestation bundled with its signature.
-#[derive(Debug, Clone, SszEncode, SszDecode, HashTreeRoot)]
+///
+/// `HashTreeRoot` is intentionally not derived: `XmssSignature` is a fixed-size
+/// byte vector for cross-client serialization but the spec Merkleizes it as a
+/// container, so roots would diverge. No code hashes `SignedAttestation`.
+#[derive(Debug, Clone, SszEncode, SszDecode)]
 pub struct SignedAttestation {
     /// The index of the validator making the attestation.
     pub validator_id: u64,
--- a/crates/common/types/src/attestation.rs
+++ b/crates/common/types/src/attestation.rs
@@ -36,9 +36,13 @@ pub struct AttestationData {
 
 /// Validator attestation bundled with its signature.
 ///
+/// <div class="warning">
+///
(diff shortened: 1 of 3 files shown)


crates/common/types/src/attestation.rs
```diff
@@ -35,7 +35,11 @@ pub struct AttestationData {
 }
 
 /// Validator attestation bundled with its signature.
-#[derive(Debug, Clone, SszEncode, SszDecode, HashTreeRoot)]
+///
+/// `HashTreeRoot` is intentionally not derived: `XmssSignature` is a fixed-size
```
pablodeymo, line 41: I would like to have a warning icon here
https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block

pablodeymo (approved):

crates/common/types/src/attestation.rs
MegaRedHand: Done

PR #309, opened 2026-04-20
feat(api): add runtime aggregator role toggle

MegaRedHand:
## Summary

Ports [leanSpec PR #636](https://github.com/leanEthereum/leanSpec/pull/636). Adds two admin endpoints that let operators rotate aggregation duties across nodes at runtime without a restart.

- `GET /lean/v0/admin/aggregator` returns `{"is_aggregator": bool}`
- `POST /lean/v0/admin/aggregator` with body `{"enabled": bool}` returns `{"is_aggregator": <new>, "previous": <old>}`

Returns `503` when the controller is not wired, `400` on malformed body or non-boolean `enabled` (JSON ints `0`/`1` are explicitly rejected, matching the spec).

## Design

The CLI flag `--is-aggregator` now seeds a shared `AggregatorController` (a thin `Arc<AtomicBool>` in `ethlambda-types`) instead of a plain bool. The blockchain actor reads the flag fresh on every tick and every gossip attestation, so runtime toggles take effect from the next tick with no protocol-message churn.

`axum::Extension` is used to thread the controller into admin handlers so existing store-backed routes don't need to change their signatures.

The `lean_is_aggregator` metric is updated from the actor's tick read (with a cached `last_seen_aggregator`) rather than from the HTTP handler, so the gauge reflects what the actor actually acted on.

## Scope limitations (intentional, matching spec)

| Scope | Status |
|---|---|
| Persistence across restart | ❌ in-process only |
| Gossip subnet re-subscription | ❌ frozen at startup |
| ENR advertisement | ❌ no ENR lifecycle |
| Authentication | ❌ consistent with rest of HTTP API |

**Operational model**: standby aggregators must boot with `--is-aggregator=true` so gossip subscriptions are in place, then use the admin endpoint to rotate duties (hot-standby model). A node booted with `--is-aggregator=false` and toggled on later will have no extra subnets to aggregate. This is documented in the CLI flag's docstring and the `CLAUDE.md` gotcha section.


bin/ethlambda/src/main.rs
```diff
@@ -154,8 +163,19 @@ async fn main() -> eyre::Result<()> {
...
+    // and the API server (which exposes GET/POST admin endpoints).
+    let aggregator = AggregatorController::new(options.is_aggregator);
+
+    let blockchain = BlockChain::spawn(store.clone(), validator_keys, aggregator.clone());
+
+    // Note: SwarmConfig.is_aggregator is intentionally a plain bool, not the
+    // AggregatorController — subnet subscriptions are decided once here and
+    // are not re-evaluated at runtime. Toggling via the admin API affects
+    // aggregation logic but not the gossip mesh. See crates/net/p2p/src/lib.rs
+    // for the invariant.
```
MegaRedHand, line 178:
```suggestion
    // for the invariant.
    // TODO: update ENR with the dynamic is_aggregator value
```

pablodeymo (approved):

PR #313, opened 2026-04-21
refactor(storage): dedupe PayloadBuffer proofs by subsumption

MegaRedHand:
## Summary

Replace the equality-only dedup in `PayloadBuffer::push` with a subset-based dedup that maintains each data_root's proof list as an antichain. On push, if the incoming proof's participants are a subset (including equal) of any existing proof, the push is skipped; otherwise every existing proof whose participants are a strict subset of the incoming proof is removed before the insert.

### Why

After an aggregator produces a new proof for a data_root, the child proofs merged into it are still in `new_payloads` with participant sets that are strict subsets of the new proof's. They add no coverage and just consume `PayloadBuffer` capacity, accelerating FIFO eviction of *other* data_roots. The same kind of redundancy accumulates in `known_payloads` through `drain_new_to_known` and block ingestion.

The leanSpec models this at `leanSpec/src/lean_spec/subspecs/forkchoice/store.py:1048-1062` by wholesale-replacing `latest_new_aggregated_payloads[D]` with `{new_proof}`. We can't mimic that synchronously because (a) aggregation runs on a `spawn_blocking` worker while the actor keeps writing `new_payloads` via gossip ingestion, and (b) `update_head` reads `known_payloads` only, so any known-side pruning has to be carefully timed to avoid a fork-choice coverage gap.

Dedupe-on-push sidesteps both constraints:

- Push never shrinks validator coverage: it only removes proofs whose validators are a strict subset of the added one's. Any `update_head` reading `known_payloads` between pushes sees monotonically-growing coverage, so there's no fork-choice timing risk.
- Concurrent gossip-received aggregates that are incomparable with our locally produced aggregate are preserved; ones that are subsumed are absorbed.
- No separate cleanup passes or wiring: `apply_aggregated_group`, `on_gossip_aggregated_attestation`, `on_block_core`, and `promote_new_aggregated_payloads` all benefit transparently.

### Cost

O(n·v) per push (n = proofs currently in the entry, v = participants per proof) via `HashSet<u64>` membership. `n` is bounded by the max-antichain size per data_root (typically 1 in steady state, small when concurrent aggregates overlap). If profiling later shows push in a hot path, a bitwise-native subset op on `AggregationBits` is a natural follow-up.

### Tests

- 8 new unit tests on `PayloadBuffer::push`: superset removes subset, subset skipped, equal-participants skipped, incomparable proofs coexist, superset absorbs multiple singletons in one push, mixed kept/removed under same data_root, cross-data-root independence, FIFO eviction still uses `total_proofs`.
- All 4 existing `PayloadBuffer` tests keep passing unchanged (they use pairwise-incomparable participant sets).

## Test plan

- [x] `cargo test -p ethlambda-storage` — 30 tests pass
- [x] `cargo test -p ethlambda-blockchain` — 20 unit tests pass
- [x] `cargo test -p ethlambda-blockchain --test forkchoice_spectests --release` — 70 spec tests pass
- [x] `cargo fmt --all` clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [ ] Local devnet smoke (post-review): with `--is-aggregator`, verify `lean_latest_new_aggregated_payloads` and `lean_latest_known_aggregated_payloads` stay bounded across interval 2 / interval 4 instead of growing monotonically

### DIFF
--- a/crates/common/types/src/attestation.rs
+++ b/crates/common/types/src/attestation.rs
@@ -86,12 +86,7 @@ pub fn validator_indices(bits: &AggregationBits) -> impl Iterator<Item = u64> + 
     })
 }
 
-/// Returns `true` iff every bit set in `a` is also set in `b` (i.e., participants(a) ⊆ participants(b)).
-///
-/// Operates byte-wise on the raw bitfield representation, avoiding the per-call
-/// `HashSet` allocation that index-iteration would require. Safe because SSZ
-/// decoding rejects bitlists with non-zero padding above the delimiter, so any
-/// bit not in `b` reliably reads as zero in `b.as_bytes()`.
+/// Returns `true` iff every bit set in `a` is also set in `b` (i.e., `a` is a subset of `b`).
 pub fn bits_is_subset(a: &AggregationBits, b: &AggregationBits) -> bool {
     let a_bytes = a.as_bytes();
     let b_bytes = b.as_bytes();
--- a/crates/common/types/src/attestation.rs
+++ b/crates/common/types/src/attestation.rs
@@ -146,3 +146,120 @@ impl From<AttestationData> for HashedAttestationData {
         Self::new(data)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// Build an `AggregationBits` of `len` bits with the indices in `set` flipped on.
+    fn bits(len: usize, set: &[usize]) -> AggregationBits {
+        let mut b = AggregationBits::with_length(len).unwrap();
+        for &i in set {
+            b.set(i, true).unwrap();
+        }
+        b
+    }
+
+    #[test]
+    fn subset_empty_bitlists() {
+        let empty = AggregationBits::new();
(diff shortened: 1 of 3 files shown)


crates/common/types/src/attestation.rs
```diff
@@ -86,6 +86,22 @@ pub fn validator_indices(bits: &AggregationBits) -> impl Iterator<Item = u64> +
     })
 }
 
+/// Returns `true` iff every bit set in `a` is also set in `b` (i.e., `a` is a subset of `b`).
+pub fn bits_is_subset(a: &AggregationBits, b: &AggregationBits) -> bool {
```
pablodeymo, line 90: it would be very nice to have tests for this function

MegaRedHand: Added in [ab80e2e](https://github.com/lambdaclass/ethlambda/pull/313/commits/ab80e2eb43a1d27407175993d8ce7b8fdd605b1d)

pablodeymo (approved):

PR #326, opened 2026-04-29
docs: render documentation site with mdbook

### LINKED ISSUE #122: Render documentation using mdbook
We currently have some documentation files under `docs/` that we could render into a website using [mdbook](https://rust-lang.github.io/mdBook/). For that, we'd also need to add a `SUMMARY.md`.

For an example, see how ethrex does it: https://github.com/lambdaclass/ethrex/blob/728dc7ded560c665ab1ff2cf7f3eeb197b5bc40f/Makefile#L201. They also use some plugins that might be useful.

MegaRedHand:
## Summary

Closes #122. Wires up [mdbook](https://rust-lang.github.io/mdBook/) so the existing `docs/` markdown can be rendered as a browsable site.

## Changes

- `book.toml`: mdbook config pointing `src` at `docs/`, with `git-repository-url` and an `edit-url-template` so each page links back to its source on `main`.
- `docs/SUMMARY.md`: required mdbook table of contents. Groups pages into two parts — **Consensus** (3SF-mini, LMD-GHOST) and **Operations** (Metrics, Checkpoint Sync, Fork Choice Visualization).
- `docs/introduction.md`: landing page that orients the reader and links to the existing `docs/infographics/*.html` (mdbook copies the HTML files into the rendered site as-is).
- `Makefile`: `docs-deps` (installs mdbook 0.4.51 via cargo), `docs` (one-shot build into `./book`), `docs-serve` (live-reload preview).
- `.gitignore`: excludes the `book/` build output.

No mdbook plugins are pulled in — none of the current docs use mermaid / katex / GFM alerts, so the bare renderer is enough. Plugins can be added later if/when content needs them, following the [ethrex setup](https://github.com/lambdaclass/ethrex/blob/728dc7ded560c665ab1ff2cf7f3eeb197b5bc40f/Makefile#L201) referenced in the issue.

## Test plan

- [x] `make docs-deps` installs mdbook (already cached locally; target verified)
- [x] `make docs` builds successfully and produces a complete `book/` site
- [x] `make docs-serve` renders Introduction + Consensus + Operations sections in the sidebar with correct ordering
- [x] Both infographics under `docs/infographics/` are reachable from the introduction page
- [ ] CI workflow for building / publishing the docs left as a follow-up

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help fmt lint docker-build run-devnet test
+.PHONY: help fmt lint docker-build run-devnet test docs docs-deps docs-serve
 
 help: ## 📚 Show help for each of the Makefile recipes
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
@@ -48,3 +48,12 @@ run-devnet: docker-build lean-quickstart ## 🚀 Run a local devnet using lean-q
 	@echo "Starting local devnet. Press Ctrl+C to stop all nodes."
 	@cd lean-quickstart \
 		&& NETWORK_DIR=local-devnet ./spin-node.sh --node all --generateGenesis --metrics > ../devnet.log 2>&1
+
+docs-deps: ## 📦 Install dependencies for generating the documentation
+	cargo install --version 0.4.51 mdbook
+
+docs: ## 📚 Generate the documentation site under ./book
+	mdbook build
+
+docs-serve: ## 📖 Serve the documentation locally with live reload
+	mdbook serve --open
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -48,3 +48,12 @@ run-devnet: docker-build lean-quickstart ## 🚀 Run a local devnet using lean-q
 	@echo "Starting local devnet. Press Ctrl+C to stop all nodes."
 	@cd lean-quickstart \
 		&& NETWORK_DIR=local-devnet ./spin-node.sh --node all --generateGenesis --metrics > ../devnet.log 2>&1
+
+docs-deps: ## 📦 Install dependencies for generating the documentation
+	cargo install --version 0.4.51 mdbook
```
pablodeymo: why do we use 0.4.51 and no the latest version (0.5.2)?

MegaRedHand: I didn't know there was a newer one. Let me update

MegaRedHand: Done. I also added the linkcheck plugin

pablodeymo (approved):

PR #332, opened 2026-04-30
feat(blockchain): advance XMSS preparation window preemtively

### LINKED ISSUE #262: Advance XMSS preparation window in the background
#261 calls `advance_preparation` synchronously on the block processing pipeline. This is OK for now, since the call takes ~500ms, and is only called once every ~65k slots, but we should move it to the background to avoid unnecessary instability.

conache:

This PR closes #262.

Every 65,536 slots, an XMSS signing key has to precompute its next bottom tree via leansig's [`advance_preparation`](https://github.com/leanEthereum/leanSig). The two most recently computed trees form a sliding "prepared window" of 131,072 slots - the range the key can sign for without doing more work. Once the wall-clock slot crosses out of that window, the precomputation has to run before the next signature.

PR #261 made `advance_preparation()` run synchronously on the `BlockChainServer` actor's tick handler. When the window has to slide forward, the actor blocks on the hash work long enough to stall other executions.

This PR moves that advance **off the signing path**, running it preemptively where blocking is cheap:

- **At startup**, in `BlockChain::spawn` (before the actor starts ticking) - catches each loaded key's prepared window up to the current wall-clock slot. Handles long offline gaps.
- **At the end of every tick**, after the interval's duties - advances each key's window to cover `slot + 1`, so the next tick's signing is always inside the prepared window.

The lazy advance loop inside `sign_with_*` is kept as a safety net, with added `elapsed_ms` timing logs so we'd see it if it ever fires.

## What Changed

**`crates/blockchain/src/key_manager.rs`**
- New `KeyManager::advance_keys_to(slot)` - iterates registered validators and advances both attestation and proposal keys to cover `slot`.
- New free helper `advance_key(...)` - synchronous advance loop with `Instant::now()` timing. Emits `info!` at start, `info!` with `elapsed_ms` at end, `warn!` on activation-interval exhaustion.
- Added matching `Instant::now()` + `elapsed_ms` `info!` to the pre-existing advance loops in `sign_with_attestation_key` / `sign_with_proposal_key`.

**`crates/blockchain/src/lib.rs`**
- `BlockChain::spawn`: computes current wall-clock slot and calls `key_manager.advance_keys_to(current_slot)` before the actor starts ticking.
- `on_tick`: at the very end, after metric updates, calls `self.key_manager.advance_keys_to((slot + 1) as u32)`.

- Signing is never delayed by an advance - preempt runs at the idle tail of the interval (or before any tick fires, at startup).
- Steady-state: the end-of-tick advance is a no-op in 65,535 out of every 65,536 slots.
- After a long offline gap: startup catch-up walks the window forward before any tick fires.
- Activation-interval exhaustion stays a hard error in the signing path; `advance_key` logs `warn!` and breaks so the next sign attempt surfaces the error.
- `ValidatorKeyPair` shape and the `KeyManagerError` enum are unchanged from main.

## Tests Added / Run

- [x] `make fmt` clean
- [x] `make lint` clean
- [x] `make test` passes
- [ ] Boundary-crossing verification on devnet - only fires every 65,536 slots, deferred to operational verification. Log lines `Preparing XMSS key for slot in background` (start) and `XMSS key advance complete` (success) signal the path firing.

## Related Issues / PRs

- Initially linked to #262 
- Follow-up to #261


crates/common/types/src/signature.rs
```diff
@@ -183,4 +197,18 @@ mod tests {
             result.err().map_or(String::new(), |e| e.to_string())
         );
     }
+
+    #[test]
+    fn test_advance_until_prepared_advances_then_detects_exhaustion() {
+        let key = generate_key_with_three_bottom_trees();
```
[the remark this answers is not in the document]
conache: Test ignored in https://github.com/lambdaclass/ethlambda/pull/332/commits/49a232b64e214182af8336d6042ba2b6046087db

crates/blockchain/src/lib.rs
```diff
@@ -263,14 +265,24 @@ impl BlockChainServer {
...
+                    ..
+                }) => {
+                    self.prepare_key_for_slot(validator_id, role, target_slot, ctx);
+                    continue;
+                }
+                Err(KeyManagerError::KeyUnavailable(_)) => continue,
+                Err(err) => {
+                    error!(%slot, %validator_id, %err, "Failed to sign attestation");
+                    continue;
+                }
```
MegaRedHand: This is functionally the same (or worse) than before. Currently we block until advancing the key, while this would skip attesting altogether when advancing is needed.

MegaRedHand (changes requested):
Thanks for the PR, but this approach isn't good for the following reasons:

- it skips signing when key preparation is needed
- it adds a new edge case of a key not being available
- changes are too many for what it offers

I realize it might be better to instead check at specific times if we can advance validator keys preemptively (i.e. check they're prepared for current slot + 1), along with an advance until the current slot when first initializing the node. Let's try with that first, doing it right after we finish duties for the current interval, in a blocking manner, and adding timing logs for it. We can see if a more complex approach is needed later down the line.

conache:
> Thanks for the PR, but this approach isn't good for the following reasons:
> 
> * it skips signing when key preparation is needed
> * it adds a new edge case of a key not being available
> * changes are too many for what it offers
> 
> I realize it might be better to instead check at specific times if we can advance validator keys preemptively (i.e. check they're prepared for current slot + 1), along with an advance until the current slot when first initializing the node. Let's try with that first, doing it right after we finish duties for the current interval, in a blocking manner, and adding timing logs for it. We can see if a more complex approach is needed later down the line.

@MegaRedHand thanks for the clear review! 🙏  Working on a simpler approach based on your feedback.

conache:
Hey @MegaRedHand , I just updated the PR based on your feedback. Now we're advancing the XMSS preparation window preemptively, in a blocking manner:
- at startup, so we can catch keys up to the current slot
- at the end of each tick, for slot + 1, so signing on the next tick shouldn't wait

As suggested, I added timing logs for all the cases, so we can see how long the blocking work takes. Looking forward to your review!

crates/blockchain/src/key_manager.rs
```diff
@@ -153,6 +176,28 @@ impl KeyManager {
...
+        key.advance_preparation();
+        if key.get_prepared_interval() == before {
+            warn!(validator_id, slot, "XMSS key activation interval exhausted");
+            break;
+        }
+    }
+    info!(
+        validator_id,
+        slot,
+        elapsed_ms = start.elapsed().as_millis() as u64,
```
MegaRedHand, line 196: We usually use the `Debug` implementation for printing durations:


```suggestion
        elapsed = ?start.elapsed(),
```

crates/blockchain/src/key_manager.rs
```diff
@@ -85,6 +94,7 @@ impl KeyManager {
         // Multiple advances may be needed if the node was offline for an extended period.
         if !key_pair.attestation_key.is_prepared_for(slot) {
             info!(validator_id, slot, "Advancing XMSS key preparation window");
+            let start = Instant::now();
```
MegaRedHand, line 97: Can we call `advance_keys_to` here, to avoid repeating code?

MegaRedHand (approved):
Looks good. Left some comments, but they can be addressed in another PR.

MegaRedHand:
@conache I opened https://github.com/lambdaclass/ethlambda/issues/383 for this. Feel free to start work on it.

conache:
> @conache I opened #383 for this. Feel free to start work on it.

@MegaRedHand thanks for the feedback! Sure, I'll add a follow-up PR for addressing #383.

PR #340, opened 2026-05-01
fix: make Store's public methods return `Result<T,Error>`, propagate and update every callers accordingly

### LINKED ISSUE #306: Store swallows DB errors with `.expect()`, propagate Result through storage layer
`crates/storage/src/store.rs` calls `StorageBackend` methods (`begin_read`, `begin_write`, `put_batch`, commit, etc.) but swallows all returned Results via `.expect()` at 50+ call sites. Any DB error (disk full, corruption, etc.) panics the node instead of being handled gracefully.

The fix is to make Store's public methods return `Result<T, Error>` and propagate with `?`, then update callers in `crates/blockchain/src/store.rs` accordingly. 

Related: #82 

Fixing this is also a prerequisite for #266. Without it, the "too many open files" error panics the node rather than being returned as a recoverable error.

d4m014:
All public Store methods that perform disk I/O previously panicked on failure via `.expect()`. This PR makes every such method return `Result<T, Error>` and propagates errors through every caller; the actor layer, RPC handlers, p2p handlers, fork-choice logic, and the binary entrypoint. Panicking on a recoverable storage error is unsafe in a long-running consensus node; this change ensures errors are surfaced and handled deliberately.

## What Changed
- `crates/storage/src/store.rs`:  `from_anchor_state/get_forkchoice_store/init_store → Result<Self, Error>;` `set_time`, `set_safe_target`, `insert_state, write_signed_block`, `prune_live_chain`, `prune_old_states`, `prune_old_blocks` all return `Result`; `head_slot/safe_target_slot` use `.ok_or_else(`) instead of `.unwrap()`

- `crates/storage/src/lib.rs` : re-exports `Error`                                                                                                                               

- `crates/blockchain/src/store.rs` : added `StoreError::Storage(#[source] ethlambda_storage::Error)` variant + `From impl`; all storage calls use `?` or two-step `?.ok_or()`;  affected functions updated to return `Result`                                                                                                                                 

- `crates/blockchain/src/lib.r`s : actor layer uses let `Ok(...) else { return }` for critical paths; non-critical reads use `if let Ok(...);` errors logged via `inspect_err`      

- `crates/net/rpc/src/lib.rs` : `get_latest_finalized_state/get_latest_justified_state` map storage errors to 500, missing resources to 404 

- `crates/net/rpc/src/fork_choice.rs` : local `try_store!` macro short-circuits on storage error with 500; non-critical header reads use `.ok().flatten()` 

- `crates/net/p2p/src/req_resp/handlers.rs` : block lookup uses `if let Ok(Some(b))`; `build_status` uses `.unwrap_or_default()` for graceful degradation

- `bin/ethlambda/src/checkpoint_sync.rs` : added `Storage(#[from] ethlambda_storage::Error)` to `CheckpointSyncError` 

- `bin/ethlambda/src/main.rs` : removed `Ok(...)` wrapper around `from_anchor_state;` uses `map_err` for error conversion

- `crates/blockchain/tests/forkchoice_spectests.rs` / `signature_spectests.rs` : all store calls updated to handle `Result` (`.unwrap()` in test context).

- No behavior changes on the happy path, all existing logic is preserved                                                                                                   - Storage errors that previously caused a panic now either propagate as Err, return a 500 HTTP response, or are logged and skipped depending on context
- Corrupt or missing entries in prune scans are skipped via `.filter_map` rather than aborting the entire prune 

## Tests Added / Run
- cargo test --workspace: All existing unit tests pass 
- cargo check --workspace passes clean

## Related Issues / PRs
- Closes #306

### DIFF
--- a/crates/blockchain/src/aggregation.rs
+++ b/crates/blockchain/src/aggregation.rs
@@ -127,7 +127,7 @@ pub fn snapshot_aggregation_inputs(store: &Store) -> Option<AggregationSnapshot>
         return None;
     }
 
-    let head_state = store.head_state();
+    let head_state = store.head_state().ok()?;
     let validators = &head_state.validators;
 
     let gossip_roots: HashSet<H256> = gossip_groups
--- a/crates/blockchain/src/aggregation.rs
+++ b/crates/blockchain/src/aggregation.rs
@@ -127,7 +127,9 @@ pub fn snapshot_aggregation_inputs(store: &Store) -> Option<AggregationSnapshot>
         return None;
     }
 
-    let head_state = store.head_state().ok()?;
+    let head_state = store
+        .head_state()
+        .expect("head state must exist for aggregation");
     let validators = &head_state.validators;
 
     let gossip_roots: HashSet<H256> = gossip_groups
--- a/crates/net/p2p/src/req_resp/handlers.rs
+++ b/crates/net/p2p/src/req_resp/handlers.rs
@@ -243,7 +243,13 @@ fn canonical_blocks_by_range(
         .filter_map(|index| {
             let slot = start_slot.checked_add(index.checked_mul(step)?)?;
             let root = roots_by_slot.get(&slot)?;
-            store.get_signed_block(root).ok().flatten()
+            store
+                .get_signed_block(root)
+                .inspect_err(|e| {
+                    tracing::warn!("Failed to get signed block for root {:?}: {:?}", root, e);
+                })
+                .ok()
+                .flatten()
         })
         .collect();
     Ok(blocks)
@@ -298,14 +304,18 @@ async fn handle_blocks_by_root_response(
 
 /// Build a Status message from the current Store state.
 pub fn build_status(store: &Store) -> Status {
-    let finalized = store.latest_finalized().unwrap_or_default();
-    let head_root = store.head().unwrap_or_default();
+    let finalized = store
+        .latest_finalized()
+        .expect("store should always contain a finalized checkpoint");
+    let head_root = store
+        .head()
+        .expect("store should always contain a head block root");
     let head_slot = store
         .get_block_header(&head_root)
         .ok()
         .flatten()
-        .map(|h| h.slot)
-        .unwrap_or(0);
+        .expect("head block exists")
+        .slot;
     Status {
         finalized,
         head: Checkpoint {
--- a/crates/net/p2p/src/req_resp/handlers.rs
+++ b/crates/net/p2p/src/req_resp/handlers.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashMap, HashSet};
 
+use ethlambda_storage::Error;
 use ethlambda_storage::Store;
 use libp2p::{PeerId, request_response};
 use rand::seq::SliceRandom;
@@ -133,10 +134,11 @@ async fn handle_blocks_by_root_request(
 
     let mut blocks = Vec::new();
     for root in request.roots.iter() {
-        if let Some(signed_block) = server.store.get_signed_block(root) {
-            blocks.push(signed_block);
+        match server.store.get_signed_block(root) {
+            Ok(Some(signed_block)) => blocks.push(signed_block),
+            Ok(None) => {} // missing blocks are skipped per spec
+            Err(e) => warn!(%root, %e, "DB error fetching block by root"),
         }
-        // Missing blocks are silently skipped (per spec)
     }
 
     let found = blocks.len();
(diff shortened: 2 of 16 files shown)


MegaRedHand (changes requested):
@d4m014 please fix the conflicts

d4m014:
> @d4m014 please fix the conflicts

done, there were some new errors across different files merged to main that called Store function which i already made to return Result in this PR (i've now resolved them), i suggest we review and merge this PR before new ones got merged to main, because any changes that call Store's functions without handling the new Result changes will break this one and we will have to go through each files for errors again.

cc @MegaRedHand @pablodeymo

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -182,43 +202,51 @@ async fn handle_blocks_by_range_request(
...
     }
 
-    (start_slot..=end_slot)
-        .filter_map(|slot| {
+    let blocks = (0..count)
+        .filter_map(|index| {
+            let slot = start_slot.checked_add(index.checked_mul(step)?)?;
             let root = roots_by_slot.get(&slot)?;
-            store.get_signed_block(root)
+            store.get_signed_block(root).ok().flatten()
```
MegaRedHand: Let's add an `inspect_err` here, to log a warning when this fails.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -270,12 +298,14 @@ async fn handle_blocks_by_root_response(
...
+    let finalized = store.latest_finalized().unwrap_or_default();
+    let head_root = store.head().unwrap_or_default();
     let head_slot = store
         .get_block_header(&head_root)
-        .expect("head block exists")
-        .slot;
+        .ok()
+        .flatten()
+        .map(|h| h.slot)
+        .unwrap_or(0);
```
MegaRedHand: Let's keep the expect here

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -270,12 +298,14 @@ async fn handle_blocks_by_root_response(
 
 /// Build a Status message from the current Store state.
 pub fn build_status(store: &Store) -> Status {
-    let finalized = store.latest_finalized();
-    let head_root = store.head();
+    let finalized = store.latest_finalized().unwrap_or_default();
+    let head_root = store.head().unwrap_or_default();
```
MegaRedHand: Let's `expect` here, to keep the old behavior.

crates/net/rpc/src/fork_choice.rs
```diff
@@ -30,26 +30,41 @@ pub struct ForkChoiceNode {
 pub async fn get_fork_choice(
     axum::extract::State(store): axum::extract::State<Store>,
 ) -> impl IntoResponse {
-    let blocks = store.get_live_chain();
+    use axum::http::StatusCode;
+
+    macro_rules! try_store {
```
MegaRedHand: No macros!

~Let's move parts of this (or the whole function) into something that returns `Result`, and forwards errors with `?`. Then we can have this match for that function only.~

Actually, let's just panic on any DB error. This PR's focus is on changing the storage module only. We can remove panics in a later PR.

crates/storage/src/store.rs
```diff
@@ -472,15 +475,22 @@ impl Store {
...
     pub fn get_forkchoice_store(
         backend: Arc<dyn StorageBackend>,
-        mut anchor_state: State,
+        anchor_state: State,
         anchor_block: Block,
-    ) -> Self {
-        assert!(
-            anchor_pair_is_consistent(&mut anchor_state, &anchor_block),
-            "anchor block does not match anchor state"
+    ) -> Result<Self, Error> {
```
MegaRedHand, line 485: Let's make this return another `Error` type that wraps the current API `Error`. That'll let us return nice errors for some additional checks we have here. We can put this new `Error` inside an error module.

crates/storage/src/store.rs
```diff
@@ -532,74 +542,61 @@ impl Store {
...
-        let view = self.backend.begin_read().expect("read view");
+    fn get_metadata<T: SszDecode>(&self, key: &[u8]) -> Result<T, Error> {
+        let view = self.backend.begin_read()?;
         let bytes = view
-            .get(Table::Metadata, key)
-            .expect("get")
-            .expect("metadata key exists");
-        T::from_ssz_bytes(&bytes).expect("valid encoding")
+            .get(Table::Metadata, key)?
+            .ok_or("Metadata key not found")?;
```
MegaRedHand: This one, for example, could be a `ethlambda_storage::error::Error` variant

crates/blockchain/src/aggregation.rs
```diff
@@ -127,7 +127,7 @@ pub fn snapshot_aggregation_inputs(store: &Store) -> Option<AggregationSnapshot>
         return None;
     }
 
-    let head_state = store.head_state();
+    let head_state = store.head_state().ok()?;
```
MegaRedHand: Let's keep the panic here.

MegaRedHand (commented):
Thanks for the big contribution! I left some comments. We'll try to merge this once addressed, so no more conflicts appear.

pablodeymo:
@d4m014 can you please solve conflicts? so then we can re-review the PR.
Thanks!

MegaRedHand:
Closing this PR due to inactivity and conflicts

PR #342, opened 2026-05-01
refactor(p2p): move node name registry into P2PServer

### LINKED ISSUE #106: Move node name registry to each module
We currently have a name registry mapping the peer ID to its name (loaded from the configuration at startup) in a static variable with an `RwLock`. We should instead keep this hashmap inside the P2P struct, for simplicity. This reduces global state and would also allow us to more easily add node names to logs (instead of the opaque peer IDs).

After doing this, we can replicate it in other modules, for example, to map from validator ID to node name.

conache:

The `PeerId -> node_name` lookup lived in a global `RwLock<HashMap<PeerId, &'static str>>` populated once at startup. This PR moves it into `P2PServer` so the actor owns its own state.

## What Changed

- **`crates/net/p2p/src/lib.rs`**: the registry is now a field on the P2P actor's state struct, with a small helper to resolve a peer's name.
- **`crates/net/p2p/src/metrics.rs`**: no longer owns the registry — its peer-event helpers take the resolved name directly.
- **`bin/ethlambda/src/main.rs`**: the loader now returns the map and hands it off to `P2P::spawn` at startup; it also logs how many entries it loaded.

- Metric label values (`lean_connected_peers{client="..."}`) are unchanged for valid configs.
- Bad rows in `validator-config.yaml` (unparseable secp256k1 privkeys) used to be silently dropped;  they now emit a `warn!` per drop and are still skipped. No new panics, no new failure paths.
- No consensus-layer impact (fork choice, attestations, STF, signatures untouched).

## Tests Added / Run

  - [x] `make fmt`
  - [x] `make lint`
  - [x] `make test`

## Related Issues / PRs

  - Closes #106


pablodeymo:
@conache can you please solve the conflicts?

pablodeymo (changes requested):
Request solve the conflicts

conache:
> @conache can you please solve the conflicts?

@pablodeymo thanks for the heads-up! Sure, just solved the conflicts.

pablodeymo (approved):

PR #343, opened 2026-05-04
feat(metrics): add lean_gossip_mesh_peers gauge

MegaRedHand:

Implements the `lean_gossip_mesh_peers` metric introduced in [leanMetrics PR #35](https://github.com/leanEthereum/leanMetrics/pull/35).

This is a `Gauge` that exposes the number of peers currently in our gossipsub mesh, broken down by `client` label. It complements the existing `lean_connected_peers` gauge: connected peers are nodes we have an active libp2p connection to, while mesh peers are the smaller subset gossipsub has selected as full-message neighbours.

The `client` label uses the same `<name>_<N>|unknown` resolution as `lean_connected_peers` (handled by the existing `resolve()` helper in `crates/net/p2p/src/metrics.rs`), so PR #35's label-format clarification needed no code changes — only the new gauge.

## What Changed

- `crates/net/p2p/src/metrics.rs`
  - Register `LEAN_GOSSIP_MESH_PEERS: IntGaugeVec` with the `client` label.
  - Add `update_gossip_mesh_peers(peers)`: groups peers by resolved client name (acquiring the `NODE_NAME_REGISTRY` read lock once for the batch). To avoid the scrape-window gap that an `IntGaugeVec::reset()` + repopulate would leave between writes, the function instead enumerates currently-registered children via `Collector::collect()` and `set(0)` for any client labels no longer in the mesh, then writes the new counts. Each `set()` is a single registry write, so a concurrent Prometheus scrape always observes a coherent label set.
- `crates/common/metrics/src/lib.rs`
  - Re-export `prometheus::core::Collector` so consumers can enumerate existing label children.
- `crates/net/p2p/src/swarm_adapter.rs`
  - Add a `tokio::time::interval` (10 s, `MissedTickBehavior::Skip`) arm to the `swarm_loop` `tokio::select!`. On each tick it reads `swarm.behaviour().gossipsub.all_mesh_peers()` and pushes counts into the gauge.

- **No state changes on the network/consensus path.** The new code only reads `gossipsub.all_mesh_peers()` (a `BTreeSet` iteration) and writes to a Prometheus registry; it cannot affect block propagation, fork choice, or attestation handling.
- **Scrape-stable updates.** Departed client labels are written as `set(0)` rather than removed via `reset()`, so a Prometheus scrape between writes always sees a complete (non-empty) label set. Alerting rules like `lean_gossip_mesh_peers > 0` see continuous time series instead of vanishing labels.
- **No starvation of swarm/cmd events.** The new arm shares the `tokio::select!` with the swarm event and command receiver, with `MissedTickBehavior::Skip` to avoid backlogging metric refreshes if the swarm loop is busy. Polling cost is one short iterator pass per 10 s over the mesh peer set (typically <50 peers).
- **Polling rationale.** Mesh membership changes inside gossipsub heartbeats (700 ms) and is not surfaced as a `BehaviourEvent`, so an event-driven update is not possible without forking the gossipsub behaviour. 10 s polling bounds scrape staleness while keeping overhead negligible.

## Tests Added / Run

The change is operational (Prometheus gauge update): no new unit tests were added because there's no observable behaviour outside the metrics endpoint and the gauge update is exercised by every devnet run.

Verification commands:

- `cargo check -p ethlambda-p2p` — clean
- `make fmt` — clean
- `make lint` (clippy with `-D warnings`) — clean
- `cargo test --workspace --release` — all passing

Devnet verification (4 ethlambda nodes, local image):

- All 4 nodes published `lean_gossip_mesh_peers{client="ethlambda_N"} = 1` for each peer once mesh formed.
- After `docker stop ethlambda_3`, surviving nodes' `lean_gossip_mesh_peers{client="ethlambda_3"}` dropped to `0` within ~6 s instead of disappearing — confirming scrape-stable behavior.

## Related Issues / PRs

- Implements the `lean_gossip_mesh_peers` metric from leanEthereum/leanMetrics#35

### DIFF
--- a/crates/net/p2p/src/swarm_adapter.rs
+++ b/crates/net/p2p/src/swarm_adapter.rs
@@ -12,10 +12,7 @@ use tracing::{error, warn};
 use crate::{Behaviour, BehaviourEvent, metrics, req_resp::Request, req_resp::Response};
 
 /// Interval between gossipsub mesh peer metric refreshes.
-///
-/// Slightly slower than the gossipsub heartbeat (700ms) so the gauge
-/// reflects post-heartbeat mesh state with minimal polling overhead.
-const MESH_METRIC_REFRESH_INTERVAL: Duration = Duration::from_secs(1);
+const MESH_METRIC_REFRESH_INTERVAL: Duration = Duration::from_secs(10);
 
 pub enum SwarmCommand {
     Publish {
--- a/crates/net/p2p/src/swarm_adapter.rs
+++ b/crates/net/p2p/src/swarm_adapter.rs
@@ -1,13 +1,21 @@
+use std::time::Duration;
+
 use libp2p::{
     Multiaddr, PeerId, StreamProtocol,
     futures::StreamExt,
     request_response::{self, OutboundRequestId},
     swarm::SwarmEvent,
 };
-use tokio::sync::mpsc;
+use tokio::{sync::mpsc, time::MissedTickBehavior};
 use tracing::{error, warn};
 
-use crate::{Behaviour, BehaviourEvent, req_resp::Request, req_resp::Response};
+use crate::{Behaviour, BehaviourEvent, metrics, req_resp::Request, req_resp::Response};
+
+/// Interval between gossipsub mesh peer metric refreshes.
+///
+/// Slightly slower than the gossipsub heartbeat (700ms) so the gauge
+/// reflects post-heartbeat mesh state with minimal polling overhead.
+const MESH_METRIC_REFRESH_INTERVAL: Duration = Duration::from_secs(1);
 
 pub enum SwarmCommand {
     Publish {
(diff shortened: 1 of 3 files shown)


crates/net/p2p/src/swarm_adapter.rs
```diff
@@ -1,13 +1,21 @@
...
 use tracing::{error, warn};
 
-use crate::{Behaviour, BehaviourEvent, req_resp::Request, req_resp::Response};
+use crate::{Behaviour, BehaviourEvent, metrics, req_resp::Request, req_resp::Response};
+
+/// Interval between gossipsub mesh peer metric refreshes.
+///
+/// Slightly slower than the gossipsub heartbeat (700ms) so the gauge
+/// reflects post-heartbeat mesh state with minimal polling overhead.
+const MESH_METRIC_REFRESH_INTERVAL: Duration = Duration::from_secs(1);
```
MegaRedHand:
```suggestion
/// Interval between gossipsub mesh peer metric refreshes.
const MESH_METRIC_REFRESH_INTERVAL: Duration = Duration::from_secs(10);
```

pablodeymo (approved):

PR #348, opened 2026-05-07
feat(p2p): add inbound BlocksByRange req/resp support

### LINKED ISSUE #346: Add inbound `BlocksByRange` request support
https://github.com/leanEthereum/leanSpec/pull/691 was recently merged in the spec, adding a new `BlocksByRange` reqresp protocol, with a similar function as `BlocksByRoot`.

We need to support it so other clients can use it. Implementation should be similar to `BlocksByRoot`.

dicethedev:

This PR adds inbound `BlocksByRange` request-response support to the P2P req/resp protocol implementation.

The change follows the recently merged spec update:
- leanEthereum/leanSpec#691

This is needed so peers can request canonical blocks by slot range, similar to the existing `BlocksByRoot` protocol.

The implementation:
- registers the new protocol
- adds SSZ request/response handling
- supports serving canonical blocks from local storage
- validates malformed requests

This improves interoperability with other clients implementing the updated spec.

---
### Req/Resp Protocol
- Added `BlocksByRangeRequest`
- Added `BlocksByRange` response payload variant
- Added protocol ID:
  - `/leanconsensus/req/blocks_by_range/1/ssz_snappy`

### Codec
- Updated request/response codec read paths
- Updated request/response codec write paths

### Behaviour Registration
- Registered `BlocksByRange` in the libp2p request-response behaviour

### Inbound Request Handling
- Added inbound request handler for `BlocksByRange`
- Serves canonical blocks by walking backward from the current fork-choice head
- Skips:
  - empty slots
  - side forks

### Validation
Added validation for:
- `step == 0`
- `count > 1024`

Invalid requests return protocol error responses.

### Tests
- Added unit test for canonical range selection and ordering
---

### Preserved Invariants
- Only canonical blocks are returned
- Returned blocks preserve requested slot ordering
- Empty slots are skipped
- Non-canonical side forks are ignored

### Behavior Notes
- Invalid requests are rejected early with error responses
- Maximum request size is capped at `1024` blocks
- Implementation mirrors existing `BlocksByRoot` handling patterns for consistency

---
### Added
- `blocks_by_range_returns_canonical_blocks_in_requested_order`
### Verified With
```bash
cargo fmt --check
cargo check -p ethlambda-p2p
cargo test -p ethlambda-p2p blocks_by_range_returns_canonical_blocks_in_requested_order
git diff --check
```
---
## Related Issues / PRs
- Closes #346
- Related to leanEthereum/leanSpec#691


crates/net/p2p/src/req_resp/codec.rs
```diff
@@ -76,6 +82,9 @@ impl libp2p::request_response::Codec for Codec {
     {
         let label = protocol_label(protocol.as_ref());
         match protocol.as_ref() {
+            STATUS_PROTOCOL_V1 => decode_status_response(io).await,
+            BLOCKS_BY_ROOT_PROTOCOL_V1 => decode_blocks_by_root_response(io).await,
```
MegaRedHand: These were duplicated, I think 🤔

crates/net/p2p/src/req_resp/messages.rs
```diff
@@ -89,6 +92,7 @@ impl std::fmt::Debug for ResponseCode {
 pub enum ResponsePayload {
     Status(Status),
     BlocksByRoot(Vec<SignedBlock>),
+    BlocksByRange(Vec<SignedBlock>),
```
MegaRedHand: Do we need to differentiate between both of these? Maybe we can unify them into a single `Blocks` variant

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -140,6 +157,97 @@ async fn handle_blocks_by_root_request(
...
+
+    let Some(last_offset) = count
+        .checked_sub(1)
+        .and_then(|value| value.checked_mul(step))
+    else {
+        return Vec::new();
+    };
+    let Some(end_slot) = start_slot.checked_add(last_offset) else {
+        return Vec::new();
+    };
```
MegaRedHand: These two can be combined:

```suggestion
    let Some(end_slot) = count
        .checked_sub(1)
        .and_then(|value| value.checked_mul(step))
        .and_then(|last_offset| start_slot.checked_add(last_offset)) else {
        return Vec::new();
    };
```

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -140,6 +157,97 @@ async fn handle_blocks_by_root_request(
...
+
+    let mut roots_by_slot = HashMap::new();
+    let mut current_root = store.head();
+
+    while !current_root.is_zero() {
```
MegaRedHand, line 213: We should create an index block number -> block root for this. Maybe even change our blocks table to be indexed by (number, root).

I'll create an issue for it

MegaRedHand (commented):
Left some comments and questions

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: Alright then.

crates/net/p2p/src/req_resp/messages.rs
dicethedev: > Do we need to differentiate between both of these? Maybe we can unify them into a single `Blocks` variant

Good point. We don’t need to differentiate them at the response payload level because both protocols return the same payload shape: a stream/vector of `SignedBlock`s.

I unified them into a single `ResponsePayload::Blocks(Vec<SignedBlock>)` variant and made both `BlocksByRoot` and `BlocksByRange` decode through the same block response path. The request variants remain separate because their request payloads differ.

dicethedev:
@MegaRedHand You can review again

MegaRedHand:
CI's failing due to a linter error

crates/net/p2p/src/req_resp/handlers.rs
MegaRedHand: https://github.com/lambdaclass/ethlambda/issues/355

dicethedev:
> CI's failing due to a linter error

let me see if I can fix from my end. @MegaRedHand You review issue again.

MegaRedHand (approved):
LGTM

PR #351, opened 2026-05-08
feat(p2p): use BlocksByRange for long-range sync

### LINKED ISSUE #347: Use `BlocksByRange` for long range syncing
https://github.com/leanEthereum/leanSpec/pull/691 was recently merged in the spec, adding a new `BlocksByRange` reqresp protocol, with a similar function as `BlocksByRoot`.

Once #346 adds support for it, we can start using it in place of `BlocksByRoot` for downloading long ranges of blocks.

dicethedev:
This PR closes #347 by wiring the `BlocksByRange` protocol added in #348 into the status-response sync path.

Previously, when a peer's `Status` response revealed it was ahead of our local head, we had no mechanism to backfill the gap. Now, when the gap exceeds a configurable threshold (`LONG_RANGE_SYNC_THRESHOLD = 2 slots`), we request the missing range using `BlocksByRange` instead of relying on gossip or individual `BlocksByRoot` fetches.

For small gaps (1–2 slots), we defer to the existing `FetchBlock` path since roots are typically already available from gossip and `BlocksByRoot` is more precise for that case.

---
## What Changed
**`lib.rs`**
- Added `LONG_RANGE_SYNC_THRESHOLD: u64 = 2` constant

**`req_resp/handlers.rs`**
- Updated `handle_status_response` to branch on gap size:
  - `gap > LONG_RANGE_SYNC_THRESHOLD` → `request_blocks_by_range_from_peer`
  - `gap ≤ threshold` → defers to gossip / `FetchBlock`

---
- `request_blocks_by_range_from_peer` already batches internally at `MAX_REQUEST_BLOCKS` (1024), so nodes thousands of slots behind are handled correctly across multiple requests with no additional changes
- `handle_blocks_by_range_response` (added in #348) already forwards each block to the blockchain layer — the response path is complete
- `BlocksByRoot` behavior for individual missing blocks (`FetchBlock`, retry/backoff logic) is unchanged

---
## Tests Added / Run
No new tests required. The range response handling and canonical block selection are covered by the test added in #348 (`blocks_by_range_returns_canonical_blocks_in_requested_order`).

## Related Issues / PRs
- Closes [Use BlocksByRange for long range syncing #347](https://github.com/lambdaclass/ethlambda/issues/347)
- Depends on / builds on [feat(p2p): add inbound BlocksByRange req/resp support #348](https://github.com/lambdaclass/ethlambda/pull/348)
- Related to [feat: add BlocksByRange req/resp protocol leanEthereum/leanSpec#691](https://github.com/leanEthereum/leanSpec/pull/691)

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


dicethedev:
@MegaRedHand do we still need this PR now that #355 is planned? Since  #355  adds a slot/block-number index for efficient BlocksByRange lookups, I’m wondering if this PR should wait for that, or if it’s still useful to keep this implementation for now and let #355 optimize the lookup path later.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -207,6 +234,12 @@ fn canonical_blocks_by_range(
         return Vec::new();
     };
 
+    // Avoid expensive lookups if the requested range is too far in the past (beyond recent gossip history).
+    let head_slot = store.head_slot();
+    if head_slot.saturating_sub(end_slot) > MAX_SLOT_LOOKBACK {
+        return Vec::new();
+    }
```
MegaRedHand: Does the spec mention something like this? Otherwise, we should just remove it, since we already keep only the latest 30000 blocks, and would just return nothing in those cases.

MegaRedHand:
Let's keep this PR and #355 separate. It's easier to review this way

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -118,8 +127,26 @@ async fn handle_status_request(
...
+async fn handle_status_response(server: &mut P2PServer, status: Status, peer: PeerId) {
     info!(finalized_slot=%status.finalized.slot, head_slot=%status.head.slot, "Received status response from peer {peer}");
+
+    let our_head_slot = server.store.head_slot();
+    if status.head.slot <= our_head_slot {
+        return;
+    }
+    let gap = status.head.slot - our_head_slot;
+
+    if gap > LONG_RANGE_SYNC_THRESHOLD {
```
MegaRedHand: I think it's simpler here to just always do the long range sync. We don't usually receive many status messages

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -380,6 +443,61 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+            warn!(
+                %peer,
+                start_slot = next_slot,
+                count = batch_count,
+                "Failed to send BlocksByRange request (swarm adapter closed)"
+            );
+            return false;
+        };
+
+        server.range_request_ids.insert(request_id);
```
MegaRedHand: We should deduplicate requests, to avoid asking all our peers for the same range

MegaRedHand (commented):
Left some comments.

dicethedev:
> Left some comments.

You can review again.

MegaRedHand (approved):
LGTM. We'll run some tests before merging

MegaRedHand:
@dicethedev the CI is failing. The step field was removed in #365

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -88,6 +94,8 @@ pub async fn handle_req_resp_message(
             // Check if this was a block fetch request
             if let Some(root) = server.request_id_map.remove(&request_id) {
                 handle_fetch_failure(server, root, peer, ctx).await;
+            } else if server.range_request_ids.remove(&request_id) {
+                warn!(%peer, ?request_id, "BlocksByRange request failed");
```
MegaRedHand: We should look into adding retry logic here.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -366,6 +414,68 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+            count = batch_count,
+            "Sending BlocksByRange request"
+        );
+
+        let Some(request_id) = server
+            .swarm_handle
+            .send_request(
+                peer,
+                Request::BlocksByRange(request),
+                libp2p::StreamProtocol::new(BLOCKS_BY_RANGE_PROTOCOL_V1),
```
MegaRedHand: We should space these in time. We can send a single request and wait for a response before requesting more.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -88,6 +94,8 @@ pub async fn handle_req_resp_message(
             // Check if this was a block fetch request
             if let Some(root) = server.request_id_map.remove(&request_id) {
                 handle_fetch_failure(server, root, peer, ctx).await;
+            } else if server.range_request_ids.remove(&request_id) {
```
MegaRedHand: `range_request_ids` should store the requested range for validation

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -366,6 +414,68 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+        return true;
+    }
+    let end_slot = start_slot.saturating_add(count).saturating_sub(1);
+
+    // Deduplicate: skip if we already have this range in-flight
+    if server.pending_sync_ranges.contains(&(start_slot, end_slot)) {
+        info!(%peer, start_slot, end_slot, "BlocksByRange already in-flight, skipping duplicate");
+        return true;
+    }
+    server.pending_sync_ranges.insert((start_slot, end_slot));
```
MegaRedHand: `pending_sync_ranges` is never cleared. We should remove entries once we get a response.

Also, we should look into ways to deduplicate ranges. If we receive (122, 189) and (122, 190), we should only make a request for (189, 190).

crates/net/p2p/src/lib.rs
```diff
@@ -302,6 +303,8 @@ impl P2P {
             connected_peers: HashSet::new(),
             pending_requests: HashMap::new(),
             request_id_map: HashMap::new(),
+            range_request_ids: HashSet::new(),
+            pending_sync_ranges: HashSet::new(),
```
MegaRedHand: We should name these similarly, maybe `pending_root_requests` and `pending_range_requests`, `request_ids_to_root` and `request_ids_to_range`.

Also, maybe we should merge `request_id_map` with `range_request_ids`, making the value an enum. That way we can do a single map get and then match on the map value, instead of one get and map per response type.

MegaRedHand (commented):
Left some more comments, but they can be addressed in another PR.

Also, there's an issue with `decode_payload` (#384), which makes response decoding fail for BlocksByRange. We need to fix that before merging this. Feel free to look at that if you have time.

dicethedev:
> Also, there's an issue with `decode_payload` (#384), which makes response decoding fail for BlocksByRange. We need to fix that before merging this. Feel free to look at that if you have time.

@MegaRedHand  Thanks! So regarding the `decode_payload` issue (#384), would you prefer I fix that on this branch as part of the current PR, or should I open a separate PR for it?

MegaRedHand:
A separate PR would be best. We need to add tests for it anyway, so this PR isn't actually needed for testing.

MegaRedHand (changes requested):
Blocked until #384 is fixed

dicethedev:
@MegaRedHand still waiting for reviews

MegaRedHand:
PR looks good, but it requires #384 to be fixed first

dicethedev:
> PR looks good, but it requires #384 to be fixed first

I will push a fix soon for #384

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -86,8 +114,29 @@ pub async fn handle_req_resp_message(
...
+                        .remove(&(start_slot, end_slot));
+                    send_after(
+                        Duration::from_millis(500),
+                        ctx.clone(),
+                        p2p_protocol::RetryRangeSync {
+                            peer_id: peer,
+                            start_slot,
+                            end_slot: total_end_slot, // retry the full remaining range
+                        },
+                    );
```
MegaRedHand: Let's make this similar to `handle_fetch_failure`. In fact, I think we can remove this match altogether, something like before:

```rust
if let Some(pending_request) = server.outbound_requests.remove(&request_id) {
    handle_fetch_failure(server, pending_request, peer, ctx).await;
}
```

And then have the match inside `handle_fetch_failure`. Also, we should add retry tracking to the range requests too.

crates/net/p2p/src/lib.rs
```diff
@@ -403,6 +422,31 @@ impl P2PServer {
...
+
+        // safety check: if already synced, skip retry
+        let still_needed = !self
+            .pending_range_requests
+            .contains(&(start_slot, end_slot));
+
+        if still_needed {
+            tracing::trace!(%peer, start_slot, end_slot, "Skipping retry, range already resolved");
+            return;
+        }
```
MegaRedHand: We remove the entry from `pending_range_requests` right before sending the `RetryRangeSync` message. That would make this always land on this skip on error.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -353,15 +463,105 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+        if effective_start > total_end_slot {
+            info!(
+                %peer,
+                start_slot,
+                total_end_slot,
+                "BlocksByRange fully covered by in-flight requests, skipping"
+            );
+            return true;
+        }
+    }
```
MegaRedHand: We should look for a better way to handle deduplication. I'm thinking we can simplify this by having a single `BlocksByRange` request in progress, then we can just merge all seen ranges.

When we receive the first new range, we initialize our total range to the start and end of that range. Since the start can't possibly change unless we advance our head, we can add to this range by bumping the end slot whenever we find a newer range from peer status responses. Then, whenever we receive a response for a `BlocksByRange` message, we'd increase the start slot to the end of the received range and request the next chunk. Once we reach the end slot, we can stop requesting blocks, and represent that by setting our "queue" to `None`.

For peer tracking, we can have a list of peers with the latest slot we saw in their status message. We can then request ranges from peers that we know have them, while dropping peers with an old head.

```rust
// We can also use std::range::Range<u64> instead
range_sync_state: Option<RangeSyncState>

struct RangeSyncState {
    current_range: Range<u64>,
    peer_set: HashMap<PeerId, u64>,
}
```

MegaRedHand (commented):
Retry functionality is not working, currently. Let's remove it before merging this PR.

We can add the retry functionality in another PR. I left some comments regarding that.

crates/net/p2p/src/lib.rs
```diff
@@ -514,8 +586,8 @@ async fn handle_swarm_event(
         } => {
             let direction = connection_direction(&endpoint);
             let reason = match cause {
-                None => "remote_close",
-                Some(err) => {
+                Option::None => "remote_close",
+                Option::Some(err) => {
```
MegaRedHand: Why is the fully-qualified syntax necessary?

dicethedev: > Why is the fully-qualified syntax necessary?

It isn’t necessary. I used the fully qualified syntax to work around a temporary rust-analyzer false positive that treated `None` as a variable while the dependency graph was incomplete. Should I revert back?

MegaRedHand: Yes please.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -353,26 +471,102 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
     // Map request_id to root for failure handling
-    server.request_id_map.insert(request_id, root);
+    server
+        .outbound_requests
+        .insert(request_id, PendingRequestKind::Root(root));
 
     true
 }
 
+async fn request_next_range_batch(server: &mut P2PServer) -> bool {
```
MegaRedHand, line 489: The returned value isn't used anywhere

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -118,8 +163,30 @@ async fn handle_status_request(
...
 }
 
-async fn handle_status_response(status: Status, peer: PeerId) {
+async fn handle_status_response(server: &mut P2PServer, status: Status, peer: PeerId) {
     info!(finalized_slot=%status.finalized.slot, head_slot=%status.head.slot, "Received status response from peer {peer}");
+
+    let our_head_slot = server.store.head_slot();
+    if status.head.slot <= our_head_slot {
+        return;
+    }
```
MegaRedHand, line 172: We should ignore status reponses with checkpoints for future slots. If someone tells us they have a block for a slot that hasn't arrived yet, we should ignore them.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -268,6 +333,56 @@ async fn handle_blocks_by_root_response(
...
+
+    let Some(ref blockchain) = server.blockchain else {
+        server.range_sync_state = None;
+        warn!(%peer, "No blockchain handler available");
+        return;
+    };
+
+    for block in blocks {
```
MegaRedHand, line 357: Just in case, we should sort blocks in slot-increasing order, so the blockchain actor processes them efficiently.

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -353,26 +471,102 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+
+    info!(
+        %peer,
+        start_slot = batch.start,
+        count,
+        total_end_slot = server
+            .range_sync_state
+            .as_ref()
+            .map_or(batch.end, |state| state.current_range.end)
+            .saturating_sub(1),
```
MegaRedHand, line 512: This should be moved to a variable

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -353,26 +471,102 @@ pub async fn fetch_block_from_peer(server: &mut P2PServer, root: H256) -> bool {
...
+        "Sending BlocksByRange request (single batch)"
+    );
+
+    let Some(request_id) = server
+        .swarm_handle
+        .send_request(
+            peer,
+            Request::BlocksByRange(request),
+            libp2p::StreamProtocol::new(BLOCKS_BY_RANGE_PROTOCOL_V1),
```
MegaRedHand, line 521: This should be moved to a variable

MegaRedHand (approved):
LGTM

pablodeymo (approved):

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: is it worth fixing and open another pr?

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: is it worth fixing and open another pr?

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: is it worth fixing and open another pr?

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: is it worth fixing and open another pr?

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: is it worth fixing and open another pr?

PR #353, opened 2026-05-11
feat(rpc): add new `/lean/v0/blocks/finalized` endpoint

MegaRedHand:

https://github.com/leanEthereum/leanSpec/pull/713 added a new endpoint to the spec, to be used during checkpoint-sync, similar to how the beacon chain uses it.

New checkpoint-sync functionality is still missing, tracked in #354

## What Changed

The `rpc` crate was updated with the new endpoint.

## Tests Added / Run

Unit tests were added for the base case and for when a signed block is unavailable.

## ✅ Verification Checklist

- [X] Ran `make fmt` — clean
- [X] Ran `make lint` (clippy with `-D warnings`) — clean
- [X] Ran `cargo test --workspace --release` — all passing


pablodeymo (approved):
LGTM

PR #361, opened 2026-05-11
refactor(types): port leanSpec Type-1 / Type-2 aggregation envelope

pablodeymo:

Ports the typed two-level multi-signature envelope introduced by contributor commit
[`anshalshukla/leanSpec@0ab09dd`](https://github.com/anshalshukla/leanSpec/commit/0ab09dd166efb398be0373d32cdac3ec18f2b071)
("dummy type 1 and type 2 aggregation with block proofs") to ethlambda:

- `TypeOneMultiSignature` — single-message N-signer proof; replaces `AggregatedSignatureProof` on the `SignedAggregatedAttestation` gossip wire.
- `TypeTwoMultiSignature` — merged multi-message proof binding every per-attestation Type-1 plus a singleton proposer Type-1 over the block root.
- `SignedBlock.signature: BlockSignatures` → `SignedBlock.proof: ByteListMiB` carrying the SSZ-encoded merged Type-2.

The upstream commit is WIP (verify functions are explicit stubs, not yet in canonical `leanethereum/leanSpec`). ethlambda leads the wire-shape migration so the type plumbing is in place when canonical absorbs the refactor and real `lean_multisig` bindings land. **Opening as draft until canonical catches up.**

## What Changed

Landed as three commits, one per phase. Each phase compiled and passed `make test` independently.

### Phase 1 — `f2d0fb5` — additive type plumbing
- `crates/common/types/src/block.rs` — added `TypeOneInfo`, `TypeOneMultiSignature`, `TypeOneInfos` (SSZ-list limit `MAX_ATTESTATIONS_DATA + 1`), `TypeTwoMultiSignature`, and `BytecodeClaim` (typed alias for `H256`, placeholder until `lean_multisig` defines the trusted evaluation).
- SSZ round-trip + capacity unit tests.
- Pure additive: no consumers yet.

### Phase 2 — `18a60b5` — gossip-layer pipeline
- `crates/common/types/src/attestation.rs` — `SignedAggregatedAttestation.proof` → `TypeOneMultiSignature`.
- `crates/blockchain/src/aggregation.rs` — `AggregatedGroupOutput.proof`, `aggregate_job`, `resolve_child_pubkeys`, `select_proofs_greedily` all carry/read Type-1.
- `crates/storage/src/store.rs` — `PayloadEntry.proofs: Vec<TypeOneMultiSignature>`; subsumption logic reads `info.participants`.
- Block-builder helpers (`compact_attestations`, `extend_proofs_greedily`, `build_block`) operate on Type-1 throughout.
- Temporary `to_legacy` / `from_legacy` boundary at block assembly + block-body ingestion so `SignedBlock` wire stayed legacy through Phase 2.

### Phase 3 — `fc9ce1f` — block wire + storage
- `SignedBlock.signature: BlockSignatures` → `SignedBlock.proof: ByteListMiB`. Legacy `BlockSignatures` / `AttestationSignatures` / `AggregatedSignatureProof` removed.
- `crates/blockchain/src/lib.rs::propose_block` wraps the proposer XMSS as a singleton Type-1, calls `aggregate_type_2`, SSZ-encodes the merged proof, and stashes it on `SignedBlock.proof`.
- `crates/blockchain/src/store.rs::verify_signatures` rewritten as a structural-only check (mirrors upstream `verify_type_2` stub): decode the merged proof, assert `info.len() == attestations.len() + 1`, validate per-attestation `(message, slot, participants)` alignment and the trailing proposer entry; no per-Type-1 crypto.
- `crates/storage/src/store.rs::write_signed_block` / `get_signed_block` now store `ByteListMiB` blobs in the existing `BlockSignatures` column family (renaming deferred to avoid a CF migration).
- `aggregate_type_2` is a no-crypto stub today: it preserves the full `TypeOneInfos` metadata list but leaves `proof: ByteListMiB::default()`. Real merging arrives when `lean_multisig` exposes a merged-proof primitive — the existing `aggregate_proofs` only handles single-message merging.
- Test fixtures regenerated from canonical leanSpec (`make leanSpec/fixtures`). The regen also cleared three pre-existing forkchoice spec failures on `main` (`AttestationTooFarInFuture` ×2, `AggregateVerificationFailed(InvalidProof)` on `test_valid_gossip_aggregated_attestation`) — they were stale-fixture artifacts.

**Verified at gossip:** `on_gossip_aggregated_attestation` continues to run real `ethlambda_crypto::verify_aggregated_signature` on every `SignedAggregatedAttestation`. Invalid aggregates are rejected at the gossip boundary just like before.

**Block-level becomes structural:** Block-level verification no longer crypto-verifies the merged proof. The merged proof bytes can't be split client-side (the type-2 merging primitive doesn't exist in `lean-multisig` yet — the existing `aggregate_proofs` is single-message only). `verify_signatures` enforces:
- `info.len() == attestations.len() + 1`,
- each `info[i]` matches the corresponding `block.body.attestations[i]` on `participants`, `slot`, and `message`,
- the trailing `info[N]` has `message == block_root`, `slot == block.slot`, single-bit `participants` set to `block.proposer_index`,
- all participant indices fit within the validator registry.

This is the conscious "mirror upstream stubs" trade-off agreed during planning. When `lean_multisig` ships a real `verify_type_2`, the structural stub is swapped for the real call.

**Block-body ingestion preserves fork-choice LMD GHOST inputs:** since the merged proof can't be split, `process_new_block` inserts info-only Type-1 entries (real `(message, slot, participants)`, empty proof bytes) into the payload buffer. `extract_latest_known_attestations` works unchanged. Empty-bytes entries never get fed back into `aggregate_proofs` (that path is only hit when multiple proofs share the same `AttestationData`, in which case at least one came from gossip with real bytes).

**Storage:** Table name kept (`BlockSignatures`) to avoid a RocksDB CF migration; doc comment updated. Renaming to `Table::BlockProof` is a follow-up.

**Skipped tests, all behind `TODO(type1-type2)`:**
- `ssz_spectests.rs`: `SignedBlock`, `BlockSignatures`, `AggregatedSignatureProof`, `SignedAggregatedAttestation` — on-disk SSZ bytes still use the legacy schema since canonical leanSpec hasn't absorbed the refactor.
- `signature_spectests.rs`: `test_invalid_proposer_signature` — relies on block-level proposer-signature crypto, which is now a structural stub.

Attempted to bump `LEAN_SPEC_COMMIT_HASH` to `anshalshukla/leanSpec@0ab09dd` to regenerate fixtures against the new schema. Reverted: the upstream testing harness in that commit (`leanSpec/packages/testing/src/consensus_testing/keys.py`) still imports `AttestationSignatures`, which the same commit removes — `fill` crashes on module load. Documented in a `NOTE(type1-type2)` in the Makefile.

## Tests Added / Run

- Added: SSZ round-trip and capacity unit tests for the new Type-1/Type-2 containers in `crates/common/types/src/block.rs`.
- Updated: `verify_signatures_rejects_participants_mismatch`, `build_block_caps_attestation_data_entries`, `on_block_rejects_duplicate_attestation_data`, the `compact_attestations` and `extend_proofs_greedily` tests, all `forkchoice_spectests.rs` step builders, `signature_types.rs` fixture converter, and the `rpc::test_get_latest_finalized_block` test — all rebuilt to construct the new merged-proof shape.
- Verified locally:
  - `make fmt` — clean
  - `cargo clippy --workspace --all-targets -- -D warnings` — clean
  - `cargo test --workspace --release` — green (84 forkchoice spec tests, 7 signature spec tests with 1 expected skip, all unit tests pass)

## Related Issues / PRs

- Upstream commit being ported: [`anshalshukla/leanSpec@0ab09dd`](https://github.com/anshalshukla/leanSpec/commit/0ab09dd166efb398be0373d32cdac3ec18f2b071)
- Follow-ups when canonical absorbs the refactor:
  - Swap the structural `verify_type_2` stub for the real `lean_multisig` primitive.
  - Revert `LEAN_SPEC_COMMIT_HASH` skip markers in `ssz_spectests.rs` and `signature_spectests.rs`.
  - Consider renaming `Table::BlockSignatures` → `Table::BlockProof`.

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


crates/blockchain/src/aggregation.rs
```diff
@@ -387,6 +395,12 @@ pub(crate) fn aggregation_bits_from_validator_indices(bits: &[u64]) -> Aggregati
     aggregation_bits
 }
 
+// Type-1 / Type-2 envelope builders moved to
+// [`ethlambda_types::block::TypeOneMultiSignature::for_proposer`] and
+// [`ethlambda_types::block::TypeTwoMultiSignature::from_type_1s`] so the
+// `ethlambda-test-fixtures` crate can share them without depending on the
+// blockchain layer.
+
```
MegaRedHand:
```suggestion
```

crates/blockchain/tests/signature_spectests.rs
```diff
@@ -13,6 +13,19 @@ use ethlambda_test_fixtures::verify_signatures::VerifySignaturesTestVector;
...
+/// Phase 3 of the Type-1 / Type-2 aggregation migration replaces the per-
+/// attestation `verify_aggregated_signature` plus standalone proposer-signature
+/// verification with a structural check on the merged Type-2 proof; the real
+/// safety net is gossip-time per-attestation verification. Tests that only
+/// fail on the *crypto* leg accordingly pass when run against the structural
+/// stub, so they are skipped pending the `lean_multisig`-backed real
+/// `verify_type_2` primitive.
+///
+/// TODO(type1-type2): re-enable once block-level crypto verification returns.
+const SKIP_TESTS: &[&str] = &["test_invalid_proposer_signature"];
```
MegaRedHand, line 27: We need to review why this fails

crates/common/test-fixtures/src/fork_choice.rs
```diff
@@ -9,14 +9,22 @@ use crate::{
...
 use std::collections::HashMap;
 use std::path::Path;
 
+/// Per-block cap from `ethlambda_blockchain::MAX_ATTESTATIONS_DATA`, mirrored
+/// here so the test-fixtures crate stays free of a blockchain-layer dep. Used
+/// by [`BlockStepData::to_blank_signed_block`] to fall back to an empty proof
+/// for oversized-block test cases (the merged Type-2 metadata list caps at
+/// `MAX_ATTESTATIONS_DATA + 1`, which is exactly the proposer + every block
+/// attestation; oversized blocks are rejected upstream of proof decoding).
+const MAX_ATTESTATIONS_DATA: usize = 16;
```
MegaRedHand: We should move this to `ethlambda-types`

crates/common/types/src/block.rs
```diff
@@ -4,118 +4,187 @@ use libssz_derive::{HashTreeRoot, SszDecode, SszEncode};
...
+
+impl TypeTwoMultiSignature {
+    /// Merge a list of Type-1 single-message proofs into a single Type-2
+    /// multi-message proof. Mirrors upstream leanSpec's `aggregate_type_2`
+    /// stub: the metadata list (`TypeOneInfos`) is faithfully preserved so a
+    /// verifier can re-derive the per-message binding inputs, but the merged
+    /// `proof` bytes are left empty until the `lean_multisig_py` bindings ship
+    /// real cryptographic merging. Block-level signature verification stays
+    /// structural-only in the meantime, and per-attestation crypto verification
+    /// continues to run at gossip ingestion.
```
MegaRedHand, line 179: So we still lack the crypto primitives for this. Can we try using the devnet5 branch in leanMultisig?

crates/common/types/tests/ssz_types.rs
```diff
@@ -129,87 +124,11 @@ impl From<SignedAttestation> for DomainSignedAttestation {
...
-    }
-}
+// NOTE: After Phase 3 the legacy `BlockSignatures` / `AttestationSignatures` /
+// `AggregatedSignatureProof` containers are removed from the domain, and
+// `SignedBlock` now carries a single `proof: ByteListMiB` field. The pinned
+// leanSpec fixtures still use the old shape, so SSZ-byte and root assertions
+// for `SignedBlock`, `BlockSignatures`, `AggregatedSignatureProof`, and
+// `SignedAggregatedAttestation` are intentionally skipped in
+// `ssz_spectests.rs::run_ssz_test` until the fixture commit is bumped to the
+// Type-1/Type-2 schema.
```
MegaRedHand, line 134: Can we not remove the types yet? It sounds like this is only temporary

crates/blockchain/tests/signature_spectests.rs
pablodeymo: Phase 3 replaced per-attestation `verify_aggregated_signature` + standalone proposer-signature crypto at block import with a structural check on the merged Type-2 envelope. The skipped `test_invalid_proposer_signature` flips proposer signature bytes and expects rejection on the crypto leg — that leg moved to gossip-time per-attestation verification, so the test passes against the structural stub and is misleading as-is. Re-enabled in the same follow-up that swaps the structural stub for the real `verify_type_2` call (see thread on `block.rs:178`).

crates/common/types/src/block.rs
pablodeymo: Yes — checked `leanethereum/leanMultisig` `devnet5`. It re-exports everything we need from `rec_aggregation`: `TypeOneInfo`, `TypeOneMultiSignature`, `TypeTwoMultiSignature`, `aggregate_type_1`, `merge_many_type_1`, `split_type_2`, `verify_type_1`, `verify_type_2` (see [`src/lib.rs`](https://github.com/leanethereum/leanMultisig/blob/devnet5/src/lib.rs#L1-L10)).

Two integration considerations for the follow-up PR:
- Their `TypeTwoMultiSignature` uses postcard + lz4 (not SSZ) and carries `ExecutionProof` + `Evaluation<EF>`, so we keep our outer SSZ `proof: ByteListMiB` and stuff the `compress()` bytes inside.
- `setup_prover()` / `setup_verifier()` must run once at boot.

Tracking this as a follow-up rather than expanding the scope of this PR — OK to keep the structural stub here in the meantime?

crates/common/types/tests/ssz_types.rs
pablodeymo: The temporariness is real, but I think the right cure is wiring up `lean-multisig` `devnet5` (see thread on `block.rs:178`) rather than keeping the legacy containers around. Reasons:

- The legacy `BlockSignatures` / `AttestationSignatures` / `AggregatedSignatureProof` shape is incompatible with how the merged Type-2 proof actually works on the wire — they assume per-attestation signatures travel inside the block, while Type-2 collapses them into a single merged blob, so a compat shim would be a no-op.
- Keeping dead types invites accidental use of the wrong shape during the migration.
- The follow-up PR is small once the dep is in: mostly the `from_type_1s` and `verify_block_signatures` swaps.

Happy to revisit if you want a belt-and-braces approach.

crates/common/types/src/block.rs
MegaRedHand: Yes. Let's integrate in another PR

crates/common/types/src/block.rs
```diff
@@ -4,118 +4,194 @@ use libssz_derive::{HashTreeRoot, SszDecode, SszEncode};
...
-}
+// ============================================================================
+// Type-1 / Type-2 multi-signature model
+// ============================================================================
+//
+// New typed multi-signature surface introduced by leanSpec commit
+// `anshalshukla/leanSpec@0ab09dd` ("dummy type 1 and type 2 aggregation with
+// block proofs"). Defined alongside the legacy `AggregatedSignatureProof` /
+// `BlockSignatures` types during the phased migration; consumers will switch
+// over in later phases (gossip layer first, then block wire).
```
MegaRedHand:
```suggestion
```

MegaRedHand (approved):
LGTM

PR #370, opened 2026-05-14
feat: integrate leanMultisig devnet5 + leanSpec PR #717 wire format

MegaRedHand:
Branched from #378

## Summary

- Bump `lean-multisig` / `leansig_wrapper` to devnet5 HEAD (`0242c909`) and rewrite `ethlambda-crypto` on the new Type-1 / Type-2 API.
- Align the on-wire block proof with [leanSpec PR #717](https://github.com/leanEthereum/leanSpec/pull/717) — `SignedBlock.proof` carries the SSZ-encoded `TypeTwoMultiSignature { proof: ByteList512KiB }` container, which collapses to `[4-byte LE offset = 4][type2_wire]` on the wire.
- Port leanSpec PR #717's `SyncService._deconstruct_block_into_store` to the actor: imported blocks are SNARK-split per attestation, merged with local partial Type-1s, and (on aggregators) re-published on gossip.
- Re-fixture against leanSpec `d9d2e67` (just past PR #717) and migrate the prod_scheme key JSON shape so signature and forkchoice spec tests cover the real cryptographic verifier end-to-end.

## Branch commit list

| sha | what |
|--|--|
| `1cd80dd` | Crate-level integration: new Type-1 / Type-2 wrappers, real merge in `propose_block`, real `verify_type_2` in `verify_block_signatures`. |
| `2c9dec0` | First pass at PR #717 envelope: `TypeOneInfo { participants, proof }`, `TypeTwoMultiSignature { info, proof }`, drop `bytecode_claim`. `split_type_2_signature(index)` → `split_type_2_by_message(message)`. |
| `5361136` | Strip per-component Type-1 bytes when packing the Type-2 envelope — real Type-1s are ~225 KiB, N+1 copies blow the (old) 1 MiB `ByteListMiB` cap. |
| `3199e7d` → `70c7cdb` | Experimental `--crypto-merge-t1-into-t2` flag, **reverted**. The merge runs synchronously on the actor today; moving it off-thread is a follow-up. |
| `604ea4c` | Plan B: flatten `TypeOneMultiSignature` to `{ participants, proof }`, delete `TypeOneInfo` / `TypeOneInfos` / Rust `TypeTwoMultiSignature` wrappers, rename `ByteListMiB` → `ByteList512KiB`. |
| `cc3df59` | Plan A: `reaggregate_from_block` module + actor hook. Caps `MAX_REAGGREGATIONS_PER_BLOCK = 4`, skips attestations behind the store's justified checkpoint, runs only when the chain is in sync. Aggregator-only republish on gossip. |
| `4238a94` | Bump pinned `LEAN_SPEC_COMMIT_HASH` to `d9d2e67`, switch fixture generation to `--fork Lstar --scheme=test/prod`, port fixture parsers to the PR #717 schema (`signedBlock.proof.data` blob, `attestation.proof.proof.data` for gossip aggregates). |
| `961aba4` | Restore the thin SSZ container header in front of the merged proof: `SignedBlock::wrap_merged_proof` / `merged_proof_bytes` helpers; lower `MAX_ATTESTATIONS_DATA` from 16 to 8 to match leanSpec PR #717. |

## Crypto crate API

| function | wraps | notes |
|--|--|--|
| `aggregate_signatures(pks, sigs, msg, slot)` | `aggregate_type_1([], raw_xmss, …)` | Type-1 from raw XMSS only |
| `aggregate_mixed(children, raw_pks, raw_sigs, msg, slot)` | `aggregate_type_1(children, raw_xmss, …)` | mixed Type-1 children + raw XMSS |
| `aggregate_proofs(children, msg, slot)` | `aggregate_type_1(children, [], …)` | recursive Type-1 merge |
| `verify_aggregated_signature(proof, pks, msg, slot)` | `verify_type_1` | Type-1 SNARK verify + explicit binding check |
| `merge_type_1s_into_type_2(parts)` | `merge_many_type_1` | bundle N Type-1s into a Type-2 |
| `verify_type_2_signature(proof_bytes, pks_per_component, expected_bindings)` | `verify_type_2` | Type-2 SNARK verify + per-component binding check; takes `&[u8]` after envelope strip |
| `split_type_2_by_message(proof_bytes, pks_per_component, message)` | `split_type_2` (after locating index by message) | disaggregate to one Type-1; mirrors leanSpec `split_by_msg` |

Type-1 / Type-2 proof bytes are `compress_without_pubkeys()` form throughout. `verify_type_2_signature` and `split_type_2_by_message` take `&[u8]` so callers feed the raw bytes (post-envelope-strip) directly.

## Wire format

```
TypeOneMultiSignature  { participants: AggregationBits, proof: ByteList512KiB }
SignedBlock.proof bytes: [4-byte LE offset = 4][raw lean-multisig Type-2 wire]
```

The 4-byte prefix is the SSZ Container-with-one-varlen-field offset header — the spec's `TypeTwoMultiSignature { proof: ByteList512KiB }` container. `SignedBlock::merged_proof_bytes()` / `SignedBlock::wrap_merged_proof()` keep the magic number off the call sites. No Rust struct for `TypeTwoMultiSignature` — per-component participants come from `block.body.attestations[i].aggregation_bits` and `block.proposer_index`, not the envelope.

`MAX_ATTESTATIONS_DATA = 8` (down from 16, matching leanSpec PR #717). The merged Type-2 binds `MAX_ATTESTATIONS_DATA + 1 = 9` components, within lean-multisig's `MAX_RECURSIONS = 16`.

## Reaggregate-from-block

New `crates/blockchain/src/reaggregate.rs`. After `process_block` succeeds and the chain is in sync, the actor:

1. Selects up to 4 attestations whose target outruns the store's justified checkpoint AND whose participants extend the local coverage.
2. `split_type_2_by_message`-splits each one out of the block's merged Type-2 proof.
3. Merges with locally-held partial Type-1s via `aggregate_proofs`.
4. Writes the combined proof into `latest_new_aggregated_payloads`. Aggregator-role nodes republish on gossip.

5 unit tests cover the candidate-selection rules without paying SNARK cost (target-slot gate, participant subset gate, hard cap, priority ordering).

Each `split_type_2_by_message` runs a fresh SNARK; it currently executes synchronously on the actor thread. Moving to an off-thread worker mirroring `aggregation::run_aggregation_worker` is a natural follow-up if profiling shows it bleeding into the slot budget.

## Test coverage

| Suite | Result |
|---|---|
| `signature_spectests` | 13 / 13 |
| `forkchoice_spectests` | 84 / 84 |
| `stf_spectests` | 35 / 35 |
| `ssz_spectests` | 51 / 51 |
| `test_driver_e2e` (Hive) | 8 / 8 |
| `ethlambda-blockchain` unit | 24 / 24 (incl. 5 new reaggregate-from-block) |
| Workspace total | 419 / 0 |

Fixture regeneration: `LEAN_SPEC_COMMIT_HASH = d9d2e67`, generated with `make leanSpec/fixtures` (`uv run fill --fork Lstar --scheme=prod -o fixtures`). The prod_scheme key JSON shape upstream still has the pre-#725 flat layout (`attestation_public` / `attestation_secret` at top level) which the post-#725 `keys.py:395` no longer reads; this branch carries a one-shot local migration to the nested shape (`attestation_keypair.public_key` etc.) so the prod-scheme fixture filler runs. A small upstream PR converting the 12 `prod_scheme/*.json` files would let us drop the local step.

## Devnet validation

Pending — re-run on a multi-node devnet now that `verify_type_2` actually executes on the import path. Expected to surface latency cliffs that the `--crypto-merge-t1-into-t2` flag previously hid.

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -8,8 +8,12 @@ use ethlambda_types::{
     ShortRoot,
     aggregator::AggregatorController,
     attestation::{SignedAggregatedAttestation, SignedAttestation},
-    block::{ByteListMiB, SignedBlock, TypeOneMultiSignature, TypeTwoMultiSignature},
+    block::{
+        ByteListMiB, BytecodeClaim, SignedBlock, TypeOneInfo, TypeOneInfos, TypeOneMultiSignature,
+        TypeTwoMultiSignature,
+    },
     primitives::{H256, HashTreeRoot as _},
+    signature::{ValidatorPublicKey, ValidatorSignature},
 };
 use libssz::SszEncode as _;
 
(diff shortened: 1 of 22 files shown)


pablodeymo:
Reviewed locally: `cargo check` ✅, `cargo clippy -- -D warnings` ✅ (with the new `lean-multisig 0242c909` fetched).

The PR splits cleanly into three layers and reads well:

1. **crypto crate** — clean wrappers around `aggregate_type_1` / `merge_many_type_1` / `verify_type_1` / `verify_type_2` / `split_type_2`, with binding pre-checks before each SNARK verify.
2. **wire format** — slim envelope matching leanSpec PR #717, drops `message` / `slot` / `bytecode_claim`.
3. **integration** — real `verify_type_2` on the verify path, plus the `--crypto-merge-t1-into-t2` flag (default off) for the still-too-slow prover path.

## Worth raising (rough priority)

**1. Off-default path skips proposer-signature verification entirely.** With `crypto_merge_t1_into_t2=false` (the default), `propose_block` ships `merged.proof = empty`, and `verify_block_signatures` (`crates/blockchain/src/store.rs:1248`) early-returns `Ok` after the structural check. A peer can broadcast a block claiming any `proposer_index` and it will be accepted. Per-attestation crypto still runs at gossip ingestion, so attestations inside the block are real — but the block itself is unauthenticated. This is documented as intentional, but worth:
   - Emitting a one-shot `warn!` at startup when the flag is false (so operators see the tradeoff in logs)
   - Optionally a feature gate / explicit devnet marker so a default-off path can't accidentally ship to a non-devnet build

**2. Doc/wire-format inconsistency on `TypeOneInfo::proof`.** `crates/common/types/src/block.rs:71-74` says `info.proof` is "Standalone Type-1 proof bytes ... Used by split-by-msg and by re-broadcast paths." And `TypeOneMultiSignature::new` (`block.rs:121-129`) clones the bytes into both `info.proof` and outer `proof`, with a doc comment claiming they're kept aligned. But the production block builder (`crates/blockchain/src/lib.rs:466-471`) **always** strips `info[i].proof` to empty before serialization — so on the wire `info[i].proof` is never populated. The "`info[i].proof` always empty on-wire to fit the 1 MiB cap" invariant is in the PR description and the `5361136` commit message but not in any source comment. Worth a sentence on `TypeTwoMultiSignature` (or the wire-format header block) so future readers don't trust `info[i].proof` as a usable field.

**3. `propose_block` resolves the proposer pubkey twice.** `crates/blockchain/src/lib.rs:378` and `lib.rs:413` both call `proposer_validator.get_proposal_pubkey()` in the `crypto_merge_t1_into_t2=true` path. Symmetric error handling but redundant work — hoist the resolution out of the two branches.

**4. Dead `let _ = idx;` in `verify_type_2_signature`.** `crates/common/crypto/src/lib.rs:360` — either drop it or actually use `idx` in a richer error. The "reserved for diagnostics" comment doesn't pull its weight.

**5. `proposer_proof_bytes.clone()` at `crates/blockchain/src/lib.rs:410`.** In the `crypto_merge_t1_into_t2=true` path this is a ~225 KiB Type-1 SNARK; the second use at `lib.rs:450` moves it. Easy to pass-by-move and skip the clone (or restructure to compute once and use in both `for_proposer` + `merge_inputs`).

**6. Unrealistic Type-1 size in the payload-buffer test.** `crates/blockchain/src/store.rs:1465-1467` uses `PROOF_SIZE = 50 KiB` with a comment saying that's "roughly what a real lean-multisig devnet5 Type-1 SNARK weighs in at" — but the PR description elsewhere says real Type-1s are ~225 KiB (the very reason `info[i].proof` is stripped). Reconcile: either bump to 225 KiB and confirm the envelope-stripping path holds, or rephrase the comment to "artificial size chosen so packed bytes fit inline for this structural test."

**7. Minor cleanup.** Dropped-but-recomputed locals in test fixtures (`crates/common/test-fixtures/src/fork_choice.rs:170-173`, `crates/blockchain/src/store.rs:1437-1439` — `let _ = block_root;` patterns). Either delete the unused calculations or push them past their last use.

## Things that look right

- `verify_aggregated_signature` and `verify_type_2_signature` both re-derive bindings from caller-supplied `(message, slot)` and check them against what the proof binds **before** running the SNARK — the correct ordering.
- `verify_block_signatures` builds `expected_bindings` from the block body, not from anything the proof envelope carries — correct given the slim wire format.
- `split_type_2_by_message` distinguishes `UnknownMessage` from `MultipleMessages`.
- `aggregate_mixed` requires at least 1 raw sig OR at least 2 children (no useless one-child re-aggregation).
- `BytecodeClaim` is fully removed (no stragglers).
- `signature_spectests.rs` un-skips `test_invalid_proposer_signature` now that real `verify_type_2` runs.

## Test-plan gap

The "Local single-node devnet: finalizes one slot per slot for 360+ slots" checkbox was done with the default flag off — so it doesn't exercise the block-level crypto path. The `test_type_2_merge_verify_split_round_trip` ignored unit test is the only end-to-end exercise of the prover. Fine for now, but worth flagging that the multi-node devnet checkbox (still unchecked) is the real gate before flipping the default to `true`.

MegaRedHand:
Tests are failing due to some key format issues in leanSpec. I migrated the keys locally and spectests are passing, so ignore the failing `CI / Test` job

crates/blockchain/src/lib.rs
```diff
@@ -350,22 +351,106 @@ impl BlockChainServer {
...
-        let merged = TypeTwoMultiSignature::from_type_1s(all_proofs);
-        let proof_bytes = ByteListMiB::try_from(merged.to_ssz())
-            .expect("merged Type-2 proof fits in ByteListMiB");
+        // Assemble SignedBlock: wrap the proposer's raw XMSS signature into a
+        // singleton Type-1 SNARK, then merge it with every attestation Type-1
+        // into the block's single Type-2 proof.
+        //
+        // Both proofs run synchronously on the actor thread, so propose_block
+        // currently dominates the slot budget; see PR #370 for the off-thread
+        // refactor follow-up.
```
```suggestion
```

pablodeymo (approved):

PR #374, opened 2026-05-15
feat: accept multiple URLs for checkpoint sync

### LINKED ISSUE #111: Accept multiple URLs for checkpoint sync
We currently accept a URL with `--checkpoint-sync-url`, to fetch the latest finalized state from a server. We could accept multiple such URLs for redundancy. For backwards compatibility, we should still accept the current `--checkpoint-sync-url <url>` form, however, we want to also accept `--checkpoint-sync-url <url-1>,<url-2>` or `--checkpoint-sync-url <url-1> --checkpoint-sync-url <url-2>` forms.

Aliemeka:
Adds redundancy to checkpoint sync by letting operators supply more than one peer URL via `--checkpoint-sync-url`. The flag now accepts either a single URL (existing behavior), a comma-separated list, or multiple repeated occurrences. URLs are tried sequentially: the first peer that successfully serves a valid anchor wins, and we only fail startup if every URL fails. It is backward compatible. PR closes #111 

## What Changed
- `bin/ethlambda/src/main.rs`
  - `--checkpoint-sync-url` is now `Option<Vec<String>>` with`value_delimiter = ','`, so `u1,u2` and repeated `--checkpoint-sync-url` flags both populate the list. A single URL still works unchanged (backwards-compatible).
  - Extracted `try_checkpoint_url`, which wraps the existing per-URL`AnchorPairingMismatch` retry (3 attempts, 1s backoff). No behavior change for the single-URL case.
  - `fetch_initial_state` now takes `&[String]`, iterates URLs in order, logs a `warn!` with the URL and underlying error on each failure, and moves on to the next. Returns the last error only if every URL fails.
  - Updated CLI help and function doc comments to describe the failover semantics.

- Empty URL list ⇒ genesis init, identical to before.
- Single URL ⇒ same retry/error behavior as before (the per-URL retry was moved into `try_checkpoint_url`, not changed).
- Multi-URL ⇒ first-success-wins failover. Operators can see which peers failed via the warn logs; the surfaced error on total failure is the last URL's error.
- No changes to the anchor verification rules in `checkpoint_sync.rs` — every peer's anchor is still validated against the local genesis config and internal state/block consistency before it is accepted.

## Tests Added / Run
- No new unit tests — all multi-URL logic lives in `main.rs` orchestration code that drives the already-tested `fetch_finalized_anchor` / `verify_checkpoint_state` paths. Existing `checkpoint_sync` tests (verification + URL normalization) still pass.
- Commands run:
  - `cargo clippy -p ethlambda --all-targets -- -D warnings`
  - `cargo test --workspace --release` 

## Related Issues / PRs
- Closes #111 

## ✅ Verification Checklist
- [x] Ran `make fmt` — clean


bin/ethlambda/src/main.rs
```diff
@@ -558,12 +564,60 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
+    reqwest::Url::parse(url)
+        .map(|mut u| {
+            let _ = u.set_username("");
+            let _ = u.set_password(None);
+            u.set_query(None);
+            u.set_fragment(None);
+            u.to_string()
+        })
+        .unwrap_or_else(|_| "<unparseable url>".to_string())
+}
```
MegaRedHand: Let's not do URL redacting. I don't know any cases where this is used in practice in the beacon chain.

bin/ethlambda/src/main.rs
```diff
@@ -581,51 +635,53 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
-    const ANCHOR_FETCH_RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(1);
+    let mut result = try_checkpoint_url(first_url, genesis.genesis_time, &validators).await;
+    if let Err(err) = &result {
+        let url = redact_url(first_url);
+        if !rest_urls.is_empty() {
+            warn!(%url, %err, "Checkpoint sync failed for this peer; trying next URL");
+        } else {
+            warn!(%url, %err, "Checkpoint sync failed for this peer; no more URLs to try");
+        }
+    }
```
MegaRedHand: This is repeated with the next `for`

bin/ethlambda/src/main.rs
Aliemeka: Alright I will remove that

bin/ethlambda/src/main.rs
Aliemeka: I will refactor.

pablodeymo:
@Aliemeka commits must be signed in order to be able to merge the PR. Can you sign them? thanks!

bin/ethlambda/src/main.rs
```diff
@@ -558,12 +564,45 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
     bytes
 }
 
+/// Fetch the finalized anchor from a single checkpoint URL, retrying transient
+/// races where the peer advances finalization between the state and block
+/// fetches.
+async fn try_checkpoint_url(
+    url: &str,
+    genesis_time: u64,
+    validators: &[ethlambda_types::state::Validator],
```
MegaRedHand: We should import this instead:

```suggestion
    validators: &[Validator],
```

bin/ethlambda/src/main.rs
```diff
@@ -558,12 +564,45 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
+) -> Result<(State, ethlambda_types::block::SignedBlock), checkpoint_sync::CheckpointSyncError> {
```
MegaRedHand: Same


```suggestion
) -> Result<(State, SignedBlock), checkpoint_sync::CheckpointSyncError> {
```

bin/ethlambda/src/main.rs
```diff
@@ -558,12 +564,45 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
+    const MAX_ANCHOR_FETCH_ATTEMPTS: u32 = 3;
+    const ANCHOR_FETCH_RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(1);
```
MegaRedHand: Same


```suggestion
    const ANCHOR_FETCH_RETRY_DELAY: Duration = Duration::from_secs(1);
```

bin/ethlambda/src/main.rs
```diff
@@ -558,12 +564,45 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
     bytes
 }
 
+/// Fetch the finalized anchor from a single checkpoint URL, retrying transient
+/// races where the peer advances finalization between the state and block
+/// fetches.
+async fn try_checkpoint_url(
```
MegaRedHand: Can we move this to the checkpoint sync module?

bin/ethlambda/src/main.rs
```diff
@@ -581,49 +620,43 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
+    // Log only the count — URLs may carry basic-auth credentials or token query
+    // parameters; per-URL log lines below redact those before emission.
+    info!(
+        url_count = checkpoint_urls.len(),
+        "Starting checkpoint sync"
+    );
 
-    let mut attempt = 1;
+    let mut iter = checkpoint_urls.iter().peekable();
     let (state, signed_block) = loop {
```
MegaRedHand: Can we move this loop to the checkpoint sync module?

MegaRedHand:
Commit [9cf6482](https://github.com/lambdaclass/ethlambda/pull/374/commits/9cf64825ee2b404c6f253ceea86312a5d7f56475) is still unsigned

bin/ethlambda/src/main.rs
```diff
@@ -581,49 +620,43 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
-    info!(%checkpoint_url, "Starting checkpoint sync");
-
-    // The state and block are fetched in parallel; if the peer advances
-    // finalization between the two requests the pair won't match. Retry a
-    // small number of times so this transient race doesn't fail node startup.
-    const MAX_ANCHOR_FETCH_ATTEMPTS: u32 = 3;
-    const ANCHOR_FETCH_RETRY_DELAY: std::time::Duration = std::time::Duration::from_secs(1);
+    // Checkpoint sync path: try URLs in order, fail over to the next on error.
+    // Log only the count — URLs may carry basic-auth credentials or token query
+    // parameters; per-URL log lines below redact those before emission.
```
pablodeymo: **The comment promises redaction that isn't implemented — URLs are logged verbatim.**

This comment says per-URL log lines "redact those before emission," but none of them do. Every site logs the raw URL via `%url`:
- L589 (`try_checkpoint_url` retry): `warn!(%url, attempt, max, "…retrying")`
- L650: `info!(%url, "Checkpoint sync successful with this peer")`
- L655 / L657: `warn!(%url, %err, …)`

So a URL carrying credentials (`https://user:pass@peer:5052`) or a token query param (`?token=…`) ends up in logs in plaintext — exactly the case this comment says is handled. Two options:

1. **Implement the redaction** the comment claims: sanitize before logging (strip userinfo and sensitive query params), e.g. a small `redact(url)` helper used at all four sites. The `info!(url_count = …)` summary line below can stay as the extra-cautious default.
2. **Drop the claim** if redaction isn't in scope for this PR — remove the "per-URL log lines below redact those before emission" sentence so the comment doesn't describe behavior that doesn't exist.

Leaning toward (1) since you've already identified the risk, but either is fine as long as the comment and the code agree.

bin/ethlambda/src/main.rs
```diff
@@ -581,49 +620,43 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
-            {
-                warn!(
-                    attempt,
-                    max = MAX_ANCHOR_FETCH_ATTEMPTS,
-                    "Anchor state and block disagree (peer likely advanced finalization mid-fetch); retrying"
-                );
-                tokio::time::sleep(ANCHOR_FETCH_RETRY_DELAY).await;
-                attempt += 1;
+        let Some(url) = iter.next() else {
+            return Err(checkpoint_sync::CheckpointSyncError::AllPeersFailed);
```
pablodeymo: **`AllPeersFailed` discards the underlying error, which contradicts the PR description and changes single-URL behavior.**

The description says *"the surfaced error on total failure is the last URL's error,"* but this returns the generic `AllPeersFailed`. The real cause survives only in the `warn!` lines — the value propagated to the call site (`fetch_initial_state(...).inspect_err(|err| error!(%err, "Failed to initialize state"))?`) loses it.

This is also a behavior change for the single-URL backward-compat path: previously a single failing URL surfaced its specific error (`AnchorPairingMismatch`, an HTTP/verification error, …); now it always surfaces `"all checkpoint peers failed"`. The PR notes single-URL error behavior is unchanged, but the propagated error type does change here.

Suggest preserving the cause — either return the last error directly, or have the variant carry it:

```rust
#[error("all checkpoint peers failed; last error: {0}")]
AllPeersFailed(Box<CheckpointSyncError>),
```

and track the last error in the loop:

```rust
let mut last_err = None;
let (state, signed_block) = loop {
    let Some(url) = iter.next() else {
        return Err(CheckpointSyncError::AllPeersFailed(
            Box::new(last_err.expect("loop runs at least once for a non-empty list")),
        ));
    };
    match try_checkpoint_url(url, genesis.genesis_time, &validators).await {
        Ok(pair) => { info!(%url, "Checkpoint sync successful with this peer"); break pair; }
        Err(err) => {
            if iter.peek().is_some() {
                warn!(%url, %err, "Checkpoint sync failed for this peer; trying next URL");
            } else {
                warn!(%url, %err, "Checkpoint sync failed for this peer; no more URLs to try");
            }
            last_err = Some(err);
        }
    }
};
```

That keeps the new variant (nice for the multi-URL case) while restoring the specific cause in the surfaced error and the top-level `error!` log. The `expect` is safe because the `is_empty()` guard above guarantees ≥1 iteration. If you'd rather keep it simple, just `return Err(err)` on the last URL instead — but then `AllPeersFailed` is unused and the description should drop the "last URL's error" wording.

bin/ethlambda/src/main.rs
MegaRedHand: We already dropped the redaction before, we should drop the comment too

bin/ethlambda/src/main.rs
```diff
@@ -76,14 +76,20 @@ struct CliOptions {
...
-    #[arg(long)]
-    checkpoint_sync_url: Option<String>,
+    ///
+    /// Multiple URLs may be supplied for redundancy, either comma-separated
+    /// (`--checkpoint-sync-url u1,u2`) or by repeating the flag
+    /// (`--checkpoint-sync-url u1 --checkpoint-sync-url u2`). URLs are tried
+    /// in order; the first one that succeeds is used and any failures fall
+    /// over to the next URL. Startup only aborts if every URL fails.
+    #[arg(long, value_delimiter = ',')]
+    checkpoint_sync_url: Option<Vec<String>>,
```
pablodeymo: **Two small non-blocking polish notes on this flag (both optional).**

**1 — `Option<Vec<String>>` is redundant.** With clap, a `Vec<String>` arg already defaults to empty when the flag is absent, and downstream you collapse `None` and `Some(vec![])` to the same thing anyway (`options.checkpoint_sync_url.as_deref().unwrap_or(&[])` at L216, then `fetch_initial_state` branches on `is_empty()`). Dropping the `Option` removes the redundant layer:

```rust
#[arg(long, value_delimiter = ',')]
checkpoint_sync_url: Vec<String>,
```

and the call site becomes `&options.checkpoint_sync_url`.

**2 — empty entries slip through the delimiter.** `--checkpoint-sync-url u1,`, `u1,,u2`, or `--checkpoint-sync-url ""` all produce empty-string URLs in the Vec. Each one becomes a wasted failover attempt that fails deep inside `fetch_finalized_anchor` and emits a confusing `warn!` for an empty URL. Trimming + filtering before the list is used avoids that, e.g.:

```rust
options.checkpoint_sync_url
    .iter()
    .map(|u| u.trim())
    .filter(|u| !u.is_empty())
```

Neither blocks merge — just tidies the type and guards against trailing/empty commas.

bin/ethlambda/src/main.rs
Aliemeka: OK

bin/ethlambda/src/main.rs
Aliemeka: Noted.

Aliemeka:
> Commit [9cf6482](https://github.com/lambdaclass/ethlambda/pull/374/commits/9cf64825ee2b404c6f253ceea86312a5d7f56475) is still unsigned

Please @pablodeymo can I have some help with this.

MegaRedHand:
@Aliemeka Have you tried doing a `git rebase` of the whole branch?

Aliemeka:
> @Aliemeka Have you tried doing a `git rebase` of the whole branch?

Yes. Had some difficulty signing [9cf6482](https://github.com/lambdaclass/ethlambda/pull/374/commits/9cf64825ee2b404c6f253ceea86312a5d7f56475)  because it's a merge commit. Let me retry though

bin/ethlambda/src/main.rs
```diff
@@ -586,51 +596,32 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
-                );
-                tokio::time::sleep(ANCHOR_FETCH_RETRY_DELAY).await;
-                attempt += 1;
-            }
-            Err(err) => return Err(err),
-        }
-    };
+    // Checkpoint sync path: try URLs in order, fail over to the next on error.
+    // Log only the count — URLs may carry basic-auth credentials or token query
+    // parameters; per-URL log lines below redact those before emission.
```
MegaRedHand: Let's remove these

```suggestion
```

bin/ethlambda/src/main.rs
```diff
@@ -586,51 +596,32 @@ fn read_hex_file_bytes(path: impl AsRef<Path>) -> Vec<u8> {
...
+    info!(
+        url_count = checkpoint_urls.len(),
+        "Starting checkpoint sync"
+    );
+
+    let (state, signed_block) = checkpoint_sync::fetch_anchor_from_peers(
```
MegaRedHand: The name is confusing:

```suggestion
    let (state, signed_block) = checkpoint_sync::fetch_anchor_block_and_state(
```

bin/ethlambda/src/checkpoint_sync.rs
```diff
@@ -269,6 +274,73 @@ fn verify_checkpoint_state(
...
+            Err(err) => return Err(err),
+        }
+    }
+}
+
+/// Try each checkpoint URL in order, returning the first successful anchor
+/// pair. Logs per-peer success/failure. On total failure, returns
+/// `AllPeersFailed` wrapping the last peer's error so the underlying cause
+/// is preserved in the propagated error (not just the per-peer `warn!` logs).
+pub async fn fetch_anchor_from_peers(
```
MegaRedHand:
```suggestion
pub async fn fetch_anchor_block_and_state(
```

bin/ethlambda/src/checkpoint_sync.rs
```diff
@@ -269,6 +274,73 @@ fn verify_checkpoint_state(
...
+            }
+            Err(err) => return Err(err),
+        }
+    }
+}
+
+/// Try each checkpoint URL in order, returning the first successful anchor
+/// pair. Logs per-peer success/failure. On total failure, returns
+/// `AllPeersFailed` wrapping the last peer's error so the underlying cause
+/// is preserved in the propagated error (not just the per-peer `warn!` logs).
```
MegaRedHand: Let's remove some info from here, since most of it are implementation details:

```suggestion
/// the last fetch error.
```

bin/ethlambda/src/checkpoint_sync.rs
```diff
@@ -269,6 +274,73 @@ fn verify_checkpoint_state(
...
+    checkpoint_urls: &[String],
+    genesis_time: u64,
+    validators: &[Validator],
+) -> Result<(State, SignedBlock), CheckpointSyncError> {
+    let mut iter = checkpoint_urls.iter().peekable();
+    let mut last_err: Option<CheckpointSyncError> = None;
+    loop {
+        let Some(url) = iter.next() else {
+            return Err(match last_err {
+                Some(err) => CheckpointSyncError::AllPeersFailed(Box::new(err)),
```
MegaRedHand: Do we need to wrap this error? We could also return the plain error here, understanding that any error returned from this function is from the last fetch attempt.

bin/ethlambda/src/checkpoint_sync.rs
```diff
@@ -58,6 +59,10 @@ pub enum CheckpointSyncError {
     BlockHeaderJustifiedRootMismatch,
     #[error("anchor block does not match anchor state")]
     AnchorPairingMismatch,
+    #[error("all checkpoint peers failed; last error: {0}")]
+    AllPeersFailed(Box<CheckpointSyncError>),
+    #[error("no checkpoint peers configured")]
+    NoPeersConfigured,
```
MegaRedHand: Let's not call them peers. In production, they're usually a trusted checkpoint sync server.

```suggestion
    #[error("no checkpoint urls configured")]
    NoCheckpointUrls,
```

bin/ethlambda/src/checkpoint_sync.rs
```diff
@@ -269,6 +274,73 @@ fn verify_checkpoint_state(
...
+/// Fetch the finalized anchor from a single checkpoint URL, retrying transient
+/// races where the peer advances finalization between the state and block
+/// fetches.
+async fn try_checkpoint_url(
+    url: &str,
+    genesis_time: u64,
+    validators: &[Validator],
+) -> Result<(State, SignedBlock), CheckpointSyncError> {
+    const MAX_ANCHOR_FETCH_ATTEMPTS: u32 = 3;
+    const ANCHOR_FETCH_RETRY_DELAY: Duration = Duration::from_secs(1);
```
MegaRedHand: These should be at the top of the file, right after the imports

MegaRedHand (approved):
Left some comments. I think we can merge and address them after.

MegaRedHand:
@Aliemeka There are some conflicts.

Also, can we remove the error variant `CheckpointSyncError::AllPeersFailed`? (https://github.com/lambdaclass/ethlambda/pull/374#discussion_r3320862460)

pablodeymo:
@Aliemeka can you solve the conflict please?

bin/ethlambda/src/checkpoint_sync.rs
Aliemeka: Please checkout the change I made for this. I'm logging out an error message and returning the AnchorPairingError

MegaRedHand (approved):
LGTM

PR #386, opened 2026-05-20
feat: add attestation aggregate coverage metrics

pablodeymo:
## Description / Motivation

Ports two upstream changes that together add attestation aggregate coverage observability:
- [leanSpec #735](https://github.com/leanEthereum/leanSpec/pull/735): metric registration (mirrors [zeam #898](https://github.com/blockblaz/zeam/pull/898)).
- [zeam #876](https://github.com/blockblaz/zeam/pull/876): per-slot coverage computation and emission.

After this PR, all 14 coverage series receive real per-slot updates from chain activity.

### Registration (`crates/blockchain/src/metrics.rs`)

- Two `pub const &[&str]` label-set constants — single source of truth for sections and directions.
- Three `IntGaugeVec` statics:
  - `lean_attestation_aggregate_coverage_validators{section, subnet}` — `subnet="combined"` is the section total; `subnet="subnet_N"` is per-subnet coverage (lazy).
  - `lean_attestation_aggregate_coverage_subnets{section}` — covered-subnet count per section.
  - `lean_attestation_aggregate_coverage_diff_validators{direction}` — symmetric difference between block-included and locally-aggregated pre-merge aggregates.
- `init()` forces the statics and seeds 14 default zero-valued series.

**Sections:** `timely`, `late`, `block`, `combined`, `agg_start_new`, `proposal_combined`.
**Directions:** `block_only`, `timely_only`.

### Emission (4 sites)

Every per-slot section is keyed by the attestation `data.slot` (the voting round being reported), so `timely`/`late`/`block`/`combined` describe the **same cohort** (matching zeam #876).

- **`accept_new_attestations` (blockchain/store.rs)** — before each promote, snapshots `new_payloads` participant bits, tagged **per entry with their `data.slot`**, stashed on the Store. Read at the next slot boundary to populate the `timely` section ("prev_new" in zeam).
- **`on_tick` interval 1 (blockchain/lib.rs)** — emits the post-block report for `slot - 1`. Fired at **interval 1, not 0**, so the block carrying that round's votes (proposed at interval 0 of this slot) has typically been received and processed, letting the `block` section observe the same round. Computes `timely` (pre-merge snapshot filtered to the round), `late` (current `new_payloads` for the round), `block` (attestations in the **canonical head block** filtered to the round — canonical by construction, so a fork block can't poison it), and `combined` (their union); then emits `diff_validators` as the symmetric difference between `block` and `timely`. **Emits nothing** when no section has coverage for the round, so a missed slot doesn't surface as a misleading `block_only=0, timely_only=N`.
- **`start_aggregation_session` (blockchain/lib.rs)** — emits `agg_start_new` from current `new_payloads` right before fork-choice aggregation at interval 2.
- **`propose_block` (blockchain/lib.rs)** — emits `proposal_combined` (the full set of validators included across the block's aggregated attestations) after the block is built. (A payload-vs-gossip split was considered but dropped: ethlambda builds blocks exclusively from `known_aggregated_payloads`, so every included validator is payload-sourced by construction — `proposal_gossip` would always be 0 and `proposal_payloads` would always equal `proposal_combined`.)

### Supporting changes

- The three emitters live as private free `fn`s at the bottom of `blockchain/src/lib.rs`. They use `Vec<bool>` locals (`seen` for validators, `has_subnet` for subnets) plus three small helpers (`cov_add`, `cov_record`, `or_into`). Subnet derivation is `vid % committee_count`, matching the gossip subnet assignment in `crates/net/p2p/src/lib.rs`.
- `Store` gets a single `CoverageSnapshot` field (`pre_merge_new_coverage`) holding `(data.slot, AggregationBits)` entries captured before each promote — only participant bits, no proof duplication. The `block` section needs no stored state: it is read from the canonical head block at report time.
- `BlockChain::spawn` now takes `attestation_committee_count` (resolved in `main.rs` from CLI > validator-config.yaml > 1; previously only known to P2P for subnet subscriptions).

## Operator interpretation of `diff_validators`

The aggregation pipeline produces two pools for the same slot: `timely` (locally aggregated pre-merge) and `block` (what the proposer included). The diff counts validators in the symmetric difference:

- `block_only` persistently high → this node was slow to receive/aggregate via gossip; proposer had a better view.
- `timely_only` persistently high → proposer omitted attestations the network had time to gossip.
- Both near zero → local aggregation tracks proposers; the network is converging.

- **No fork-choice or state-transition change.** Coverage snapshots are pure observability and do not feed back into block processing.
- The `diff_validators` help text **intentionally diverges** from upstream's terse phrasing to spell out the symmetric-difference semantics. Metric name, labels, and values are unchanged; dashboards built against any client's schema work as-is.
- Cardinality: 14 series at registration; per-subnet series appear lazily when written. Even with full per-subnet instrumentation at 64 subnets, the validators gauge tops out at `6 × 65 = 390` series.

## Tests Added / Run

No new unit tests. The earlier draft of this PR included unit tests for an internal `Coverage` bitset wrapper; once the wrapper was inlined as `Vec<bool>` locals, those tests were exercising trivial iterator ops rather than the emission contract, so they were dropped.

Commands run:

- `cargo fmt --all -- --check` — clean
- `make lint` (clippy `-D warnings`) — clean
- `cargo test -p ethlambda-blockchain --release --lib --bins` — 21 passed
- `cargo test -p ethlambda-storage --release` — 38 passed
- `cargo test -p ethlambda-blockchain --release --test forkchoice_spectests` — 62 passed, 8 failed (all `AttestationTooFarInFuture`, pre-existing fixture flakes on `main`, unrelated to this PR).

## Related Issues / PRs

- Ports: leanEthereum/leanSpec#735
- Mirrors: blockblaz/zeam#898 + blockblaz/zeam#876

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
- [x] Ran `cargo test --workspace --release` — passing modulo 8 pre-existing forkchoice fixture flakes documented above (unrelated to this PR)


MegaRedHand (changes requested):
We still need to emit the metrics

pablodeymo:
Instrumentation added in 855f56d. Five emission sites now write to the 18 series registered in e9a04f7:

- `timely` / `late` / `block` / `combined` / `diff_validators` — emitted at the slot boundary (`on_tick` interval 0) for `slot - 1`, sourcing from a pre-merge snapshot of `new_payloads`, current late arrivals, and the last-imported block's aggregation bits.
- `agg_start_new` — emitted right before fork-choice aggregation runs (interval 2).
- `proposal_payloads` / `proposal_gossip` / `proposal_combined` — emitted from `propose_block` after `produce_block_with_signatures`, classifying validators in the proposed block as known-payload-covered vs. gossip-only.

`Coverage` lives in a new `crates/blockchain/src/coverage.rs` (~280 lines, 6 unit tests covering `add_bits`, `merge_from`, `diff_counts`, plus empty/zero/out-of-range edge cases). Subnet derivation uses `vid % committee_count` to match the gossip subnet assignment in `crates/net/p2p/src/lib.rs:241`.

Store gets a small `CoverageSnapshot` type and two `Arc<Mutex<Option<…>>>` fields — no proofs duplicated, only `AggregationBits`. The pre-merge capture happens inside `accept_new_attestations` just before `promote_new_aggregated_payloads`, so consumer-side timing concerns stay in the existing tick path.

`BlockChain::spawn` gains an `attestation_committee_count: u64` parameter — only `bin/ethlambda/src/main.rs` calls it, and the value was already resolved there (CLI > validator-config.yaml > 1).

Ready for another look.

pablodeymo:
@MegaRedHand please re-review the PR

crates/storage/src/store.rs
```diff
@@ -507,6 +521,12 @@ pub struct Store {
     known_payloads: Arc<Mutex<PayloadBuffer>>,
     /// In-memory gossip signatures, consumed at interval 2 aggregation.
     gossip_signatures: Arc<Mutex<GossipSignatureBuffer>>,
+    /// Snapshot of `new_payloads` participant bits captured right before each
+    /// promote-to-known. Observability-only.
+    pre_merge_new_coverage: Arc<Mutex<Option<CoverageSnapshot>>>,
+    /// Snapshot of the most-recently-imported block's aggregated attestation
+    /// participant bits. Reset on each imported block. Observability-only.
+    last_block_coverage: Arc<Mutex<Option<CoverageSnapshot>>>,
```
MegaRedHand: I think we should move these out of the storage crate

MegaRedHand (changes requested):

crates/blockchain/src/lib.rs
```diff
@@ -766,3 +807,194 @@ impl Handler<AggregationDeadline> for BlockChainServer {
...
+/// Pre-merge snapshot of `new_payloads` participant bits, used by the
+/// attestation aggregate coverage report.
+///
+/// Each entry is tagged with its attestation `data.slot` (the voting round) so
+/// the consumer can filter to a single round at emit time — `new_payloads` may
+/// hold entries spanning more than one slot. Holds raw participant bits; the
+/// consumer constructs coverage bitsets at emit time using the current
+/// validator and committee counts.
+#[derive(Debug, Clone)]
+pub(crate) struct CoverageSnapshot {
```
MegaRedHand: Let's move all this new code to another module

crates/blockchain/src/lib.rs
```diff
@@ -766,3 +807,194 @@ impl Handler<AggregationDeadline> for BlockChainServer {
...
+    pub(crate) entries: Vec<(u64, AggregationBits)>,
+}
+
+/// Observability-only handle holding the most recent pre-merge `new_payloads`
+/// snapshot. Owned by the blockchain actor; written during attestation
+/// promotion (`accept_new_attestations`) and read once per slot by the
+/// post-block coverage report. Kept out of the storage `Store` since it carries
+/// no fork-choice or state-transition state.
+#[derive(Clone, Default)]
+pub struct PreMergeCoverage(Arc<Mutex<Option<CoverageSnapshot>>>);
```
MegaRedHand: Is the mutex still needed?

crates/blockchain/src/store.rs
```diff
@@ -17,22 +17,45 @@ use ethlambda_types::{
...
+    if new_payloads.is_empty() {
+        return;
+    }
+    let mut entries: Vec<(u64, AggregationBits)> = Vec::new();
+    for (data, proofs) in new_payloads.values() {
+        for proof in proofs {
+            entries.push((data.slot, proof.participants.clone()));
+        }
+    }
+    coverage.save(CoverageSnapshot { entries });
```
MegaRedHand: Should this go with the rest of the snapshot functions?

crates/blockchain/src/store.rs
```diff
@@ -192,7 +215,12 @@ fn validate_attestation_data(store: &Store, data: &AttestationData) -> Result<()
 /// 800ms interval. Slot and interval-within-slot are derived as:
 ///   slot     = store.time() / INTERVALS_PER_SLOT
 ///   interval = store.time() % INTERVALS_PER_SLOT
-pub fn on_tick(store: &mut Store, timestamp_ms: u64, has_proposal: bool) {
+pub fn on_tick(
+    store: &mut Store,
+    timestamp_ms: u64,
+    has_proposal: bool,
+    coverage: Option<&PreMergeCoverage>,
+) {
```
MegaRedHand: I don't like this change. Can't we move the `snapshot_pre_merge_new_coverage` call outside this module? We already have some interval-specific processing inside `BlockChainServer::on_tick`, we can add an if there to compute this coverage when proposing or at the end of the last interval.

pablodeymo:
@MegaRedHand addressed all the review comments in c751426 — one refactor that pulls the coverage observability out of the storage/STF path:

- **Moved all the coverage code into a new `crates/blockchain/src/coverage.rs` module** — `CoverageSnapshot`, the `cov_*`/`or_into` helpers, the three `emit_*` functions, and the snapshot fn (renamed `snapshot_new_payloads`). Nothing coverage-related lives at the bottom of `lib.rs` anymore.
- **Removed the `coverage` parameter from `store::on_tick` / `accept_new_attestations` / `get_proposal_head` / `produce_block_with_signatures`.** The pre-merge snapshot is now taken in `BlockChainServer::on_tick` itself, gated on `interval == 4 || (interval == 0 && proposing)` — i.e. when proposing or at the end of the slot, exactly as you suggested. It runs right before `store::on_tick`, so the captured `new_payloads` set is identical to the old in-store capture.
- **Dropped the `Arc<Mutex<…>>`.** Now that the snapshot is owned by the actor and only touched from the single-threaded message loop, it's a plain `Option<CoverageSnapshot>` field.

No behavior change — coverage emission stays observability-only. `fmt`/`clippy -D warnings` clean; blockchain/storage/rpc/signature tests green, forkchoice unchanged (same 8 pre-existing `AttestationTooFarInFuture` fixture flakes as `main`).

crates/blockchain/src/lib.rs
```diff
@@ -190,6 +205,16 @@ impl BlockChainServer {
             .then(|| self.get_our_proposer(slot))
             .flatten();
 
+        // Snapshot the pre-merge `new_payloads` set before the store may promote
+        // it on this tick, so the post-block coverage report can see the
+        // "timely" cohort. Promotion happens when proposing (interval 0) or at
+        // the end of the slot (interval 4); skip empty snapshots so a missed
+        // round keeps the last set we saw. Pure observability.
+        let will_promote = interval == 4 || (interval == 0 && proposer_validator_id.is_some());
+        if will_promote && let Some(snapshot) = coverage::snapshot_new_payloads(&self.store) {
```
MegaRedHand: **Finding 1 (Medium): proposer interval-0 snapshot can corrupt the `timely` cohort it's about to report.**

Walking the timeline for a slot `S` this node proposes:

1. Slot `S-1` interval 4: `snapshot_new_payloads` captures round `S-1` aggregates into `pre_merge_coverage`, then `store::on_tick` promotes (drains `new_payloads`).
2. Between then and slot `S` interval 0, gossiped aggregates arrive — `store.rs` `insert_new_aggregated_payload` pushes into `new_payloads` at arbitrary network-driven times.
3. Slot `S` interval 0: since we're proposer, `will_promote` fires and `snapshot_new_payloads` returns `Some` (non-empty due to those stragglers), **overwriting** the good `S-1` snapshot with a partial set.
4. Slot `S` interval 1: `emit_post_block_coverage` filters the overwritten snapshot to `data.slot == S-1` and finds only the stragglers.

Result: `timely` under-reports and `diff_validators{block_only}` is inflated, but only on slots this node proposes, so the bias is intermittent and proposer-correlated. The interval-0 snapshot doesn't actually serve the round it's reporting; consider only refreshing `pre_merge_coverage` when the new snapshot covers the round not yet reported, or skip the interval-0 capture entirely.

crates/blockchain/src/coverage.rs
```diff
@@ -0,0 +1,201 @@
...
+    let mut combined_s = timely_s.clone();
+    or_into(&mut combined_v, &late_v);
+    or_into(&mut combined_s, &late_s);
+    or_into(&mut combined_v, &block_v);
+    or_into(&mut combined_s, &block_s);
+
+    // Emit nothing when there is no coverage for this round, rather than
+    // pushing a misleading all-zero report (e.g. `block_only=0, timely_only=N`
+    // on a missed slot). Gauges retain their previous value.
+    if !combined_v.iter().any(|&b| b) {
```
MegaRedHand: **Finding 2 (Medium-low): this guard doesn't suppress the case the comment promises.**

The comment above says it emits nothing to avoid "a misleading all-zero report (e.g. `block_only=0, timely_only=N` on a missed slot)". But the guard keys on `combined`, and `combined = timely | late | block`.

On a missed slot, `store.get_block(store.head())` returns an older block whose `att.data.slot != reporting_slot`, so `block_v` is all-false — but the `timely` snapshot for `S-1` is populated, so `combined` is non-empty and the report **is** emitted, pushing `block_only=0, timely_only=N`: exactly the case the comment claims to suppress.

To match the stated intent, gate on `block_v` (or require both `block_v` and `timely_v` non-empty) rather than `combined_v`.

MegaRedHand (commented):
Review of the attestation aggregate coverage metrics. All findings are in observability-only code (no fork-choice or state-transition impact), but two of them make the new metrics misleading under specific conditions. Inline comments below. Findings 1 and 2 are the ones worth addressing before merge; 3 is an efficiency cleanup and 5 is a schema/doc gap.

| # | Finding | Severity |
|---|---------|----------|
| 1 | Proposer interval-0 snapshot overwrites the reported round's `timely` cohort | Medium |
| 2 | Emit-gate doesn't suppress the missed-slot case its comment promises | Medium-low |
| 3 | `new_aggregated_payloads` clones 1 MiB proof blobs to keep only bits (×2/slot) | Medium (perf) |
| 5 | `subnet=subnet_N` series never written despite help text | Low (doc) |

crates/storage/src/store.rs
```diff
@@ -1251,6 +1251,20 @@ impl Store {
...
+    ///
+    /// Mirrors [`known_aggregated_payloads`]. Used by the attestation aggregate
+    /// coverage report to compute coverage from `new_payloads` before promote.
+    pub fn new_aggregated_payloads(
+        &self,
+    ) -> HashMap<H256, (AttestationData, Vec<AggregatedSignatureProof>)> {
+        let buf = self.new_payloads.lock().unwrap();
+        buf.data
+            .iter()
+            .map(|(root, entry)| (*root, (entry.data.clone(), entry.proofs.clone())))
```
MegaRedHand: **Finding 3 (Medium, perf): clones full proof blobs only to keep the participant bits.**

`entry.proofs.clone()` deep-clones `Vec<AggregatedSignatureProof>`, and each proof carries `proof_data: ByteList<1_048_576>` (`block.rs:88`). The only consumers — `snapshot_new_payloads` and the `late` loop in `emit_post_block_coverage` — keep just `proof.participants` and drop the bytes. Worse, `emit_post_block_coverage` calls `new_aggregated_payloads()` a second time within the same tick, so a proposer-slot tick deep-clones the entire pending proof pool twice purely to read bitfields.

A participants-only accessor (e.g. returning `Vec<(u64 /*data.slot*/, AggregationBits)>`) eliminates both clones.

crates/blockchain/src/coverage.rs
```diff
@@ -0,0 +1,201 @@
...
+            seen[vid] = true;
+            has_subnet[vid % cc] = true;
+        }
+    }
+}
+
+fn cov_record(section: &str, seen: &[bool], has_subnet: &[bool]) {
+    metrics::set_attestation_aggregate_coverage_validators(
+        section,
+        "combined",
```
MegaRedHand, line 61: **Finding 5 (Low, doc/schema): per-subnet `subnet=subnet_N` series are never written.**

`cov_record` only ever calls `set_attestation_aggregate_coverage_validators(section, "combined", ...)`. The `has_subnet` bitsets are computed but only feed the subnet-*count* gauge, never the per-subnet validator series. Yet the `lean_attestation_aggregate_coverage_validators` help text advertises "subnet=subnet_N is per-subnet coverage", and the gauge carries a `subnet` label dimension for it.

As written, querying `subnet=subnet_N` returns nothing — the schema overstates what's emitted. Either wire up per-subnet emission or trim the help text/label so dashboards don't expect a series that never appears.

pablodeymo:
Thanks for the thorough pass @MegaRedHand — all four addressed in 6fcefec (rebased onto the latest `main` merge):

**Finding 1 (proposer interval-0 snapshot corrupts `timely`).** Spot on, and it was worse than "catch-up only" — it biased every slot this node proposes. Dropped the interval-0 capture entirely; the pre-merge snapshot is now taken only at the interval-4 promote, which is the one that actually serves the reported round. The stragglers that used to clobber the snapshot now show up in the `late` section, where they belong. Gate simplified to `if interval == 4`.

**Finding 2 (emit-gate doesn't suppress the missed-slot case).** Right — `combined` includes `timely`, so a missed slot still emitted `block_only=0, timely_only=N`. Now gating on `block_v` (the round's votes appearing in the canonical head block), so we only report a round once there's actually a block carrying it. I went with `block_v` alone rather than "both non-empty" so the genuine `block_only=N, timely_only=0` signal (node was slow to aggregate) still gets reported.

**Finding 3 (1 MiB proof clones to read bits).** Replaced `new_aggregated_payloads()` with `new_aggregated_payload_participants() -> Vec<(u64, AggregationBits)>`, which clones only the bitfields and never touches `proof_data`. All three consumers (`snapshot_new_payloads`, the `late` loop, `agg_start_new`) use it now, so no proof blob is deep-copied anywhere in the coverage path.

**Finding 5 (`subnet=subnet_N` never written).** Per-subnet validator emission isn't wired (only the subnet *count* gauge uses `has_subnet`), so I corrected the help text to state only `subnet=combined` is currently populated and the per-subnet breakdown is reserved. Kept the `subnet` label for schema parity with the upstream/zeam dashboards rather than dropping a dimension they expect.

`fmt`/`clippy -D warnings` clean; blockchain/storage/rpc/signature tests green, forkchoice unchanged (same 8 pre-existing `AttestationTooFarInFuture` fixture flakes).

MegaRedHand (approved):

PR #402, opened 2026-05-29
fix(stf): reject blocks justifying a target whose source is below finalized

MegaRedHand:
## Problem

leanSpec's `Slot.is_justifiable_after` (pinned `f12000bd`) asserts `self >= finalized_slot`. ethlambda's `slot_is_justifiable_after` instead returned `false` on the `checked_sub` underflow — silently lenient.

As a result ethlambda **silently accepts** a block that justifies a target while its finalization scan would inspect slots *below* the finalized slot. The spec and other clients **reject** such a block (zeam errors with `InvalidJustifiableSlot`). Accepting it is a consensus-split risk: ethlambda could follow a chain its peers reject.

This was observed on a mixed ethlambda/zeam devnet, where it stalled cross-client finality:

- block slot 21 packed an attestation `{source.slot=2, target.slot=9}` (a valid, not-yet-justified, justifiable target);
- it reached supermajority while `finalized=8`, justifying target 9;
- the finalization scan then ran over slots `3..9`, hitting slot 3 `< finalized 8`.

ethlambda swallowed it; zeam (and the spec) rejected the block, so zeam halted at slot ~17 and finality froze.

## Change

- `slot_is_justifiable_after` now returns `Result<bool, Error>` and yields `Error::SlotBeforeFinalized` when `slot < finalized_slot` — the missing assert. It propagates through `is_valid_vote` → `process_attestations` and `try_finalize`, so the offending block is **rejected** (matching the spec and zeam).
- Proposer guard: `entry_passes_filters` now drops candidate attestations whose `source.slot < finalized_slot`, so ethlambda never **builds** a block its own STF would reject.
- Call sites where the slot is provably `>= finalized` (fork-choice target walk-back in `store.rs`, the target check in the builder) use `.expect()` with a documented invariant.

> Note: this is a consensus-safety fix (don't accept/produce spec-invalid blocks), not a liveness fix for the specific stalled chain. A complementary spec/protocol clarification — whether attestations with `source < finalized` should be filtered earlier across clients — is tracked separately.

## Tests

- `block_rejected_when_source_below_finalized` — reproduces the block-21 scenario; asserts `process_attestations` returns `Err(SlotBeforeFinalized)`.
- `slot_is_justifiable_after_errors_below_finalized` — error on `slot < finalized`; boundary/window/square/pronic cases otherwise.
- Existing `state_transition` lib + spec suite (51) and `blockchain` lib (21) pass; clippy `-D warnings` clean. The 24 `forkchoice_spectests` failures are pre-existing fixture drift (`missing field proofData`) on `main`, unrelated to this change.

### DIFF
--- a/crates/blockchain/src/block_builder.rs
+++ b/crates/blockchain/src/block_builder.rs
@@ -302,6 +302,14 @@ fn entry_passes_filters(
     ) {
         return Err("source_not_justified");
     }
+    // The source must not be below the finalized slot. Such a vote, once it
+    // justifies its target, drives the finalization scan over pre-finalization
+    // slots, which the STF now rejects (leanSpec `is_justifiable_after`
+    // assert). Exclude these candidates so we never build a block our own
+    // state transition would reject.
+    if att_data.source.slot < projected_finalized_slot {
+        return Err("source_before_finalized");
+    }
     if !attestation_data_matches_chain(extended_historical_block_hashes, att_data) {
         return Err("chain_mismatch");
     }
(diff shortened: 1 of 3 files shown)


MegaRedHand:
Superseded by #405

PR #407, opened 2026-06-01
Handle req-resp responses with multiple chunks

### LINKED ISSUE #384: Handle req-resp responses with multiple chunks
`decode_payload` reads the stream to the end, when it should instead read chunk by chunk. This makes BlocksByRange responses fail with a snappy error.

dicethedev:
- Fixes req-resp payload decoding so multi-chunk responses are handled correctly.
- `decode_payload` previously read the entire remaining stream, causing BlocksByRange / multi-block responses to fail when the first chunk decoder consumed bytes belonging to following chunks.
- This solves snappy decode failures for responses containing multiple payload chunks.

## What Changed
- Updated `crates/net/p2p/src/req_resp/encoding.rs`.
- Reworked `decode_payload` to read one varint-prefixed snappy payload at a time.
- Added async helpers for reading the varint prefix and one snappy frame.
- Added tests covering adjacent payloads, large snappy-framed payloads, and empty payload boundaries.

- Each response chunk is decoded independently.
- The next response code / payload remains unread for the outer response loop.
- Existing maximum payload and compressed payload size checks are preserved.
- Response size metrics now report the compressed size for the current payload only.

## Tests Added / Run
- Added req-resp encoding tests.
- Ran:
  - `cargo fmt --check`
  - `cargo test -p ethlambda-p2p`

## Related Issues / PRs
- Closes #384
- Related to #351

## ✅ Verification Checklist

- [ x] Ran `make fmt` — clean
- [ x] Ran `make lint` (clippy with `-D warnings`) — clean
- [ x] Ran `cargo test --workspace --release` — all passing

### DIFF
--- a/crates/net/p2p/src/req_resp/encoding.rs
+++ b/crates/net/p2p/src/req_resp/encoding.rs
@@ -11,15 +11,8 @@ pub const MAX_COMPRESSED_PAYLOAD_SIZE: usize = 32 + MAX_PAYLOAD_SIZE + MAX_PAYLO
 /// Decoded payload together with the size of its on-wire snappy-compressed
 /// bytes (excluding the varint length prefix).
 ///
-/// `compressed_size` is accurate for single-chunk streams (Status request /
-/// response, BlocksByRoot request). For multi-chunk streams (BlocksByRoot
-/// response) the value is over-reported on the first chunk because
-/// `decode_payload` slurps the whole stream via `read_to_end` before parsing
-/// the first varint, so `compressed_size` measures everything left after that
-/// varint rather than just this chunk's snappy frame. The metric is still
-/// useful for single-chunk traffic and as an order-of-magnitude signal on
-/// multi-chunk responses; precise per-chunk accounting would require refactoring
-/// `decode_payload` to read one varint + one snappy frame at a time.
+/// `compressed_size` excludes the varint length prefix and covers only the
+/// snappy frame for this payload.
 pub struct DecodedPayload {
     pub uncompressed: Vec<u8>,
     pub compressed_size: usize,
@@ -30,20 +23,7 @@ pub async fn decode_payload<T>(io: &mut T) -> io::Result<DecodedPayload>
 where
     T: AsyncRead + Unpin + Send,
 {
-    let mut buf = vec![];
-    let read = io
-        .take(MAX_COMPRESSED_PAYLOAD_SIZE as u64)
-        .read_to_end(&mut buf)
-        .await?;
-
-    if read < 2 {
-        return Err(io::Error::new(
-            io::ErrorKind::InvalidData,
-            "message too short",
-        ));
-    }
-    let (size, rest) = decode_varint(&buf)?;
-    let compressed_size = rest.len();
+    let size = read_varint(io).await?;
 
     if size as usize > MAX_PAYLOAD_SIZE {
         return Err(io::Error::new(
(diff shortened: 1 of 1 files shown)


crates/net/p2p/src/req_resp/encoding.rs
```diff
@@ -52,22 +32,85 @@ where
...
+
+        frame.extend_from_slice(&header);
+
+        let start = frame.len();
+        frame.resize(start + chunk_len, 0);
+        io.read_exact(&mut frame[start..]).await?;
+
+        let mut decoder = snap::read::FrameDecoder::new(frame.as_slice());
+        let mut uncompressed = Vec::new();
+        io::Read::read_to_end(&mut decoder, &mut uncompressed)?;
```
MegaRedHand: Can we do a `read_exact` instead here? We can use the expected uncompressed size as the number to read, and that would simplify this function by a lot.

MegaRedHand (changes requested):

MegaRedHand:
Can you rebase [f4b3d5e](https://github.com/lambdaclass/ethlambda/pull/407/commits/f4b3d5ebae8e7c81737c2ebc0f4aeeed38521f64) on main? That way we can review and merge that independently of #351

crates/net/p2p/src/req_resp/encoding.rs
```diff
@@ -52,22 +32,78 @@ where
...
+        let start = frame.len();
+        frame.resize(start + chunk_len, 0);
+        io.read_exact(&mut frame[start..]).await?;
+
+        let mut decoder = snap::read::FrameDecoder::new(frame.as_slice());
+        let mut uncompressed = vec![0; expected_uncompressed_size];
+        match io::Read::read_exact(&mut decoder, &mut uncompressed) {
+            Ok(()) => return Ok((uncompressed, frame.len())),
+            Err(err) if err.kind() == io::ErrorKind::UnexpectedEof => continue,
+            Err(err) => return Err(err),
```
MegaRedHand: This is reading frame by frame and decompressing all frames read until then to see if we finished reading the response chunk, right? That's not good.

MegaRedHand: We should instead read the headers and use `decompress_len` until we reach the end of the chunk, before decompressing it fully:

```rust
    let mut frame = Vec::new();
    let mut uncompressed_len = 0_usize;

    while uncompressed_len < expected_uncompressed_size {
        let mut header = [0_u8; 4];
        io.read_exact(&mut header).await?;
        let chunk_type = header[0];
        let chunk_len = u32::from_le_bytes([header[1], header[2], header[3], 0]) as usize;

        if frame.len() + header.len() + chunk_len > MAX_COMPRESSED_PAYLOAD_SIZE {
            return Err(io::Error::new(
                io::ErrorKind::InvalidData,
                "compressed message size exceeds maximum allowed",
            ));
        }

        frame.extend_from_slice(&header);
        let data_start = frame.len();
        frame.resize(data_start + chunk_len, 0);
        io.read_exact(&mut frame[data_start..]).await?;
        let data = &frame[data_start..];

        // Account for the uncompressed bytes this chunk contributes, so we know
        // when the frame is complete without decompressing it repeatedly.
        uncompressed_len += match chunk_type {
            // Compressed data: 4-byte CRC, then a snappy block whose leading
            // varint is the block's uncompressed length (read by `snap`).
            0x00 => {
                let block = data.get(4..).ok_or_else(|| {
                    io::Error::new(io::ErrorKind::InvalidData, "compressed chunk too short")
                })?;
                snap::raw::decompress_len(block)?
            }
            // Uncompressed data: 4-byte CRC, then the raw bytes.
            0x01 => chunk_len.checked_sub(4).ok_or_else(|| {
                io::Error::new(io::ErrorKind::InvalidData, "uncompressed chunk too short")
            })?,
            // Stream identifier (0xff), padding (0xfe), and reserved-skippable
            // chunks (0x80..=0xfd) carry no decoded bytes.
            0x80..=0xff => 0,
            // Reserved-unskippable chunk types are not valid here.
            _ => {
                return Err(io::Error::new(
                    io::ErrorKind::InvalidData,
                    "unsupported snappy chunk type",
                ));
            }
        };

        if uncompressed_len > expected_uncompressed_size {
            return Err(io::Error::new(
                io::ErrorKind::InvalidData,
                "uncompressed size does not match received size",
            ));
        }
    }
    // Now we can decompress in one pass...
```

MegaRedHand (approved):
LGTM

PR #412, opened 2026-06-03
feat: proving-scoped zk-alloc arena allocator (benchmark feature)

MegaRedHand:

Stacked on #408. leanVM 0520822 ships `zk-alloc`, a bump-arena allocator that speeds up proving — but it does nothing unless the binary installs it as the global allocator, and installing it naively **corrupts a live node**: `begin_phase()` resets every thread's slab, so long-lived buffers allocated by tokio/p2p/RocksDB/actor threads during a phase are silently overwritten by the next reset.

This PR adds an **off-by-default `zk-alloc` feature** that scopes the arena to proving only, so a node can run it long enough to benchmark.

## What Changed

- **`ScopedZkAlloc`** (`crates/common/crypto/src/zk_alloc.rs`): a dispatcher `#[global_allocator]` that routes allocations to leanVM's arena only on flagged proving threads; every other thread always uses System. `dealloc` defers to `ZkAllocator`'s address-based check; `realloc` is reimplemented to avoid leaking arena memory onto non-proving threads.
- **Proving threads** = the global rayon pool's workers, flagged at construction (`ThreadPoolBuilder::start_handler` + `build_global`, before leanVM's `setup_prover` builds the pool unflagged). Two failed designs are documented in the module: per-phase `rayon::broadcast` flagging deadlocks the pool, and a dedicated `install()` pool segfaults once its crossbeam internals are first allocated inside a phase.
- **Phase discipline**: all proving serialized behind a global mutex (prevents `begin_phase` nesting between the spawn_blocking aggregation worker and actor-thread block building); proof bytes serialized *after* `end_phase`, while the lock is held, so outputs land in System.
- Bin wiring: `--features zk-alloc` swaps jemalloc for `ScopedZkAlloc` and runs `init_allocator()` + `init_arena_rayon_pool()` at startup. Default build is unchanged.

## Tradeoffs (benchmark build only)

- jemalloc and `/debug/pprof` heap endpoints are unavailable (`ZkAllocator::dealloc` hardcodes System for non-arena frees).
- `zk_alloc::init()` asserts runtime core count == build-host core count → **build on the machine you run on** (binary-mode devnets, not Docker).
- Proving is fully serialized; concurrent aggregation/block-build proofs queue on the mutex.

## Validation

- Crypto aggregate/verify round-trip tests run with `ScopedZkAlloc` installed as the test-binary allocator (real arena phases): 5/5 pass.
- **4-node binary-mode devnet soak, 1019 slots (~68 min)**: all nodes ended on the same head root, `justified=1014 / finalized=1008`, **zero panics / zero ERROR lines**, RSS stable at ~2.7–3.0 GiB/node. 1053 aggregations through the arena: p50 **203 ms**, p90 283 ms, p99 1.1 s (max 3.0 s = cold first proof incl. setup).

A jemalloc-baseline run of the same topology is the natural follow-up to quantify the speedup.

## Side finding (upstream, not addressed here)

`handle_tick` schedules the next tick after `on_tick` returns, relative to its entry interval — any duty exceeding the 800 ms interval (block building proves inline, ~1.3 s) silently **skips the next interval**, so a proposing node produces no attestations that slot. Single-node devnets can never justify; multi-node devnets lose the proposer's attestations every slot. Deserves a separate issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -174,7 +174,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -185,7 +185,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -652,7 +652,7 @@ dependencies = [
 [[package]]
 name = "backend"
 version = "0.1.0"
-source = "git+https://github.com/leanEthereum/leanVM.git?rev=0520822#0520822dc64df67886f4a053b60b4de14b181348"
+source = "git+https://github.com/leanEthereum/leanVM.git?rev=8fcbd779#8fcbd77958a58666e828315de2d6ce7c93297117"
 dependencies = [
  "mt-air",
  "mt-fiat-shamir",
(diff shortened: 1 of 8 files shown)


MegaRedHand:
The latest leanVM version already includes allocator optimizations that make this unnecessary.

PR #414, opened 2026-06-03
Add lean_block_proposal_* metrics for block-proposal attestation selection

pablodeymo:

Ports the five `lean_block_proposal_*` observability metrics from [leanSpec PR #753](https://github.com/leanEthereum/leanSpec/pull/753) into the ethlambda block builder.

These metrics give cross-client visibility into the block-proposal attestation-selection path (`build_block`): how long each phase takes, how many proposal builds run, how many child payloads are greedily consumed, and how many distinct `AttestationData` / aggregated proofs end up in the proposed block. They align with [zeam #914](https://github.com/blockblaz/zeam/pull/914)'s `getProposalAttestations` instrumentation and the leanSpec naming so the [leanMetrics](https://github.com/leanEthereum/leanMetrics) dashboards work across clients.

## What Changed

**`crates/blockchain/src/metrics.rs`** — five new metrics registered with the existing `LazyLock` + `register_*!` pattern, a `BLOCK_PROPOSAL_ATTESTATION_BUILD_PHASES` label constant, registration in `init()`, public API functions, and a unit test:

| Metric | Type | Buckets / Labels |
|--------|------|------------------|
| `lean_block_proposal_attestation_build_phase_seconds` | HistogramVec | `phase` = `select_payloads`, `compact`, `stf_simulate`; buckets `0.001…8` |
| `lean_block_proposal_attestation_builds_total` | Counter | one per proposal attempt |
| `lean_block_proposal_child_payloads_consumed_total` | Counter | greedily-picked proofs before compaction |
| `lean_block_proposal_attestation_data_selected` | Histogram | buckets `0, 1, 2, 4, 8, 16, 32` |
| `lean_block_proposal_aggregates_selected` | Histogram | buckets `0, 1, 2, 4, 8, 16, 32, 64, 128` |

**`crates/blockchain/src/block_builder.rs`** — instruments `build_block`: times the `select_attestations`, `compact_attestations`, and STF (`process_slots` + `process_block`) phases, and emits the counters/histograms after a successful build.

**`docs/metrics.md`** — documents all five in the Block Production Metrics table.

- **No behavior change.** Only metric observations were added around existing logic; block contents, selection order, and the state-transition path are untouched.
- **Architectural divergence from leanSpec, documented.** leanSpec re-runs the STF inside a fixed-point loop and observes `stf_simulate` per round. ethlambda projects justification/finalization incrementally during selection and runs the STF exactly **once** at the end, so its `stf_simulate` is a single observation per build. This is noted on the `BLOCK_PROPOSAL_ATTESTATION_BUILD_PHASES` doc comment, consistent with the upstream PR's own caveat that phase timings are not directly comparable across clients.
- `attestation_data_selected` and `aggregates_selected` are observed from the post-compaction body (one merged proof per distinct `AttestationData`), matching the spec's intent.
- Metrics are only emitted on a **successful** build; a build that errors out in the STF is already counted by the existing `lean_block_building_failures_total`.

## Tests Added / Run

- New unit test `metrics::tests::block_proposal_attestation_build_metrics_are_usable` — verifies the phase metric registers and accepts every label in `BLOCK_PROPOSAL_ATTESTATION_BUILD_PHASES`, and that the companion counters/histograms are callable. Guards against drift between the label constant and the strings passed at the `build_block` call sites.
- Existing `block_builder` tests (`build_block_*`, `compact_attestations_*`, `extend_proofs_greedily_*`) now exercise the new metric paths and pass unchanged.

Commands run:
- `make fmt` — clean
- `cargo test -p ethlambda-blockchain --lib` — 23 passing

## Related Issues / PRs

- Ports [leanEthereum/leanSpec#753](https://github.com/leanEthereum/leanSpec/pull/753)
- Related to [blockblaz/zeam#914](https://github.com/blockblaz/zeam/pull/914)
- Follows the metrics pattern from #406 (per-subnet attestation aggregate coverage)

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
- [ ] Ran `cargo test --workspace --release` — only `ethlambda-blockchain` lib suite run (23 passing); full workspace release run not yet executed

### DIFF
--- a/crates/blockchain/src/block_builder.rs
+++ b/crates/blockchain/src/block_builder.rs
@@ -9,7 +9,10 @@
 //! without re-running the STF. The final STF runs once after selection to
 //! seal `state_root`.
 
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    time::Instant,
+};
 
 use ethlambda_crypto::aggregate_proofs;
 use ethlambda_state_transition::{
@@ -54,6 +57,7 @@ pub(crate) fn build_block(
 ) -> Result<(Block, Vec<AggregatedSignatureProof>, PostBlockCheckpoints), StoreError> {
     info!(slot, proposer_index, "Building block");
 
+    let select_start = Instant::now();
     let selected = select_attestations(
         head_state,
         slot,
@@ -61,10 +65,17 @@ pub(crate) fn build_block(
         known_block_roots,
         aggregated_payloads,
     );
+    metrics::observe_block_proposal_phase("select_payloads", select_start.elapsed());
+
+    // Each entry in `selected` is one greedily-picked child payload, counted
+    // before compaction merges proofs sharing the same AttestationData.
+    let child_payloads_consumed = selected.len();
 
     // Compact: merge proofs sharing the same AttestationData via recursive
     // aggregation so each AttestationData appears at most once (leanSpec #510).
+    let compact_start = Instant::now();
     let compacted = compact_attestations(selected, head_state)?;
+    metrics::observe_block_proposal_phase("compact", compact_start.elapsed());
 
     let (aggregated_attestations, aggregated_signatures): (Vec<_>, Vec<_>) =
         compacted.into_iter().unzip();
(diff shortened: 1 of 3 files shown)


crates/blockchain/src/block_builder.rs
```diff
@@ -80,10 +91,20 @@ pub(crate) fn build_block(
...
+    // ethlambda runs the STF once after selection (it projects justification
+    // incrementally instead of re-running the STF per loop round), so this is
+    // a single `stf_simulate` observation per build.
+    let stf_start = Instant::now();
     process_slots(&mut post_state, slot)?;
     process_block(&mut post_state, &final_block)?;
+    metrics::observe_block_proposal_phase("stf_simulate", stf_start.elapsed());
     final_block.state_root = post_state.hash_tree_root();
 
+    metrics::inc_block_proposal_attestation_builds();
```
MegaRedHand: This should be called for each attestation we select. Ideally, we'd use an `inc_by` instead of incrementing multiple times.

crates/blockchain/src/metrics.rs
```diff
@@ -748,3 +849,42 @@ pub fn set_node_sync_status(status: SyncStatus) {
             .set(i64::from(*label == active));
     }
 }
+
+#[cfg(test)]
+mod tests {
```
MegaRedHand: Let's remove this

crates/blockchain/src/block_builder.rs
pablodeymo: Done in 6bb9f87 — moved `inc_block_proposal_attestation_builds()` into the `select_attestations` fixed-point loop so it increments once per selected `AttestationData`, and updated the help text + docs row to match.

crates/blockchain/src/metrics.rs
pablodeymo: Removed in 6bb9f87. The `build_block_*` unit tests already exercise these metric paths, and the absolute-value assertions were racing against those tests on the shared global statics anyway.

MegaRedHand (approved):
LGTM

PR #418, opened 2026-06-06
fix: skip validator duties while syncing

### LINKED ISSUE #236: Don't propose/attest blocks during sync

dicethedev:

- Prevents validators from proposing blocks or producing attestations while their local view is syncing.
- Signing with a stale local view can place proposals and attestations on a chain the network has already left.
- Builds on the leanEthereum/leanSpec#708 sync heuristic added in the preceding PR (#417 ).

## What Changed

- Gate scheduled block proposals using the node’s sync status.
- Report `has_proposal = false` to the store when a proposal is skipped.
- Gate attestation production at interval 1.
- Log skipped proposal and attestation duties.
- Keep block processing, fork choice, aggregation, metrics, and key advancement active while syncing.
- Add tests covering gating, recovery, and network-wide stalls.

- Proposals and attestations are disabled only while the local node is considered syncing.
- Duties resume once the node catches up past the hysteresis recovery boundary.
- Duties remain enabled during a network-wide stall so validators can help the chain recover.
- Incoming network data and synchronization processing remain unaffected.
- A skipped proposal is not reported to fork choice as locally available.

## Tests Added / Run

- Added tests verifying:
  - Proposals and attestations are gated while syncing.
  - Duties resume after catching up.
  - Duties remain enabled during network-wide stalls.
- Ran `cargo test -p ethlambda-blockchain --lib --offline`.
  - 31 passed; 0 failed.
- Ran `cargo clippy -p ethlambda-blockchain --lib --offline -- -D warnings`.
- Ran `cargo fmt --all -- --check`.
- Ran `git diff --check`.

## Related Issues / PRs

- Closes #236
- Depends on the sync-status heuristic PR.
- Follows leanEthereum/leanSpec#708.

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
- [ ] Ran `cargo test --workspace --release` — all passing

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -9,3 +9,4 @@
 !LICENSE
 !.git/
 !.cargo/
+!shadow/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,96 +47,56 @@ jobs:
     steps:
       - uses: actions/checkout@v6
 
-      # Read the pinned leanSpec commit from the Makefile (single source of truth)
-      - name: Get leanSpec pinned commit
-        id: lean-spec
-        run: echo "commit=$(sed -n 's/^LEAN_SPEC_COMMIT_HASH:= *//p' Makefile)" >> $GITHUB_OUTPUT
+      - name: Get leanSpec fixtures release info
+        id: fixtures-release
+        run: |
+          api_url="https://api.github.com/repos/leanEthereum/leanSpec/releases/latest"
+          json=$(curl -sL "$api_url")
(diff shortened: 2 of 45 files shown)


pablodeymo:
@dicethedev can you fix the conflicts? (we merged the PR 417)

dicethedev:
> @dicethedev can you fix the conflicts? (we merged the PR 417)

okay sure. You can review again @pablodeymo

pablodeymo (approved):

PR #421, opened 2026-06-08
chore: bump leanVM to 2498896

MegaRedHand:
## Motivation

Track upstream leanVM at `2498896` (previously `f66d4a9`).

## Description

No source changes were required: the XMSS aggregation API used by `ethlambda-crypto` (`xmss_aggregate`, `AggregatedXMSS::compress`/`decompress`) is unchanged at this rev. Only `crates/common/crypto/Cargo.toml` and `Cargo.lock` change.

Notes:
- `2498896` is one commit ahead of the previous candidate `a44bca4`: a whir prover perf fix (leanVM#250, avoids a serial 256 MiB copy of combined-statement weights).
- Plonky3 `p3-*` crates are pinned at `bde2e86e`, the commit locked by leanVM's own `Cargo.lock`. leanVM references Plonky3's default branch with no rev, so `cargo update` floats them forward; newer Plonky3 commits use the unstable `maybe_uninit_slice` feature, which doesn't build on Rust 1.92.0.

## Testing

- `cargo check --workspace` and `make lint` are clean.
- `cargo test --workspace --release`: all unit/integration tests pass. The spec fixture suites fail on stale local fixtures (`missing field attestationPubkey` at JSON parse time); verified via `git stash` that the failure is identical without this change.
- `cargo test -p ethlambda-crypto --release -- --ignored` (the slow XMSS aggregation/verification tests) all pass against the new rev.

> Note: previously targeted `8eec56c`, then `a44bca4`; updated to `2498896` (devnet4 HEAD).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -652,7 +652,7 @@ dependencies = [
 [[package]]
 name = "backend"
 version = "0.1.0"
-source = "git+https://github.com/leanEthereum/leanVM.git?rev=f66d4a9#f66d4a974eced803574eb0ea43d812e523c8d7ad"
+source = "git+https://github.com/leanEthereum/leanVM.git?rev=a44bca4#a44bca4d2b991175f3f176c03f9fe681227f5bdc"
 dependencies = [
  "mt-air",
  "mt-fiat-shamir",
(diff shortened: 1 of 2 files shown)


MegaRedHand:
We see lower aggregation performance on this commit. We're pausing merging this while investigating

pablodeymo (approved):

PR #385, opened 2026-05-20
ci: download latest leanSpec fixture release instead of generating fixtures

dicethedev:
## 🗒️ Description / Motivation

This PR updates the CI and test fixture workflow so leanSpec fixtures are downloaded from the published leanSpec release assets instead of being generated in CI from a pinned checkout.

The repository now publishes the latest production fixtures as release assets, so CI should consume those directly rather than regenerating them.

This improves:
- CI performance
- fixture consistency
- workflow simplicity
- alignment with released leanSpec vectors

It also removes the overhead and complexity of generating fixtures during CI runs.

---

###  `ci.yml`
- Removed:
  - leanSpec checkout step
  - fixture generation steps

- Added:
  - latest release SHA fetch step
  - fixture cache restore keyed by release SHA
  - download of `fixtures-prod-scheme.tar.gz`
  - SHA256 verification
  - extraction into `leanSpec/fixtures`
---
### `Makefile`
Updated the `leanSpec/fixtures` target:
- Removed:
  - `uv run fill ...` fixture generation flow
- Added:
  - release download
  - SHA verification
  - extraction workflow
 ---
## `CLAUDE.md`
Updated development documentation to reflect the new fixture workflow:
- fixtures are now downloaded from released leanSpec assets
- local development no longer regenerates fixtures directly

---
### Preserved Invariants
- Existing test paths still use `leanSpec/fixtures`
- `make test` behavior remains unchanged
- Test harness logic remains unchanged

### Behavior Changes
- CI now consumes published leanSpec release fixtures
- Fixture cache invalidation now occurs when the release SHA changes

### Reviewer Notes
- No runtime or product behavior changes
- Changes are limited to:
  - CI fixture acquisition
  - local `make leanSpec/fixtures` workflow

---

## Tests Added / Run
No new unit or integration tests were added.

### Verification Performed
- `make -n leanSpec/fixtures`
  - validated updated Makefile target expansion
- Parsed `ci.yml`
  - validated YAML structure using Python
  
---

## Related Issues / PRs
- Closes #376

## ✅ Verification Checklist
- [x] Ran `make fmt` — clean
- [x] Ran `make lint` (`clippy -D warnings`) — clean
- [x] Ran `cargo test --workspace --release` — all passing

PR #429, opened 2026-06-10
fix: update spec-test runners for the latest leanSpec fixture schema

pablodeymo:
## Motivation

#385 switched CI to download the latest leanSpec fixtures release instead of generating them from a pinned commit. The current release carries fixture-schema changes the test runners don't understand yet, so the Test job on `main` fails.

Note: leanSpec's release is a single rolling `latest` tag — it was republished mid-review (2026-06-10 17:31 UTC), which is what the second commit addresses.

## Description

Runner/harness-only changes — no consensus behavior is touched, and ethlambda already matches the spec in every affected case:

- **`tickToSlot` (fork-choice block steps):** new flag, default `true`. The early-arrival tests set it to `false` to deliver a block ahead of the store clock and assert the clock doesn't move. The forkchoice spectest runner and the Hive test driver now only tick to the block's slot when the flag is set.
- **`rejectionReason`:** leanSpec renamed `expectException`. The `verify_signatures` and `state_transition` fixture types accept both spellings via a serde alias.
- **`ssz_test`:** the SSZ fixture format string was renamed from `ssz`; the runner accepts both.
- **`proofSetting` / mocked proofs:** leanSpec now fills most vectors with placeholder aggregation proofs (`MOCKED-AGGREGATION-PROOF` sentinel) instead of paying for recursive SNARK merges, marking the regime in a top-level `proofSetting` field (0 = mocked, must not be verified; 1 = real and valid; 2 = real and invalid). The forkchoice runner routes mocked vectors through a new `on_gossip_aggregated_attestation_without_verification` (mirroring the `on_block`/`on_block_without_verification` split). Production paths always verify.

Known follow-ups (out of scope):
- The Hive test driver receives fork-choice steps one at a time with no fixture-level context, so it cannot honor `proofSetting` yet; Hive runs against mocked-proof fixtures will need a protocol addition.
- Tracking the rolling `latest` release means any upstream republish can break `main`'s CI at any time; pinning the fixtures tarball SHA256 would restore reproducibility.

## How to Test

```
make test   # downloads latest release fixtures, all suites green
```

Locally verified against the 2026-06-10 17:31 UTC fixtures drop: full `cargo test --workspace --release` passes (108 forkchoice, 118 SSZ, 67 STF, all remaining suites green).

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -393,6 +393,26 @@ pub fn on_gossip_aggregated_attestation(
     store: &mut Store,
     aggregated: SignedAggregatedAttestation,
 ) -> Result<(), StoreError> {
+    on_gossip_aggregated_attestation_core(store, aggregated, true)
+}
+
+/// Process a gossiped aggregated attestation WITHOUT verifying its proof.
+///
+/// Only for spec tests whose fixtures carry mocked (placeholder) proofs
+/// (`proofSetting == 0`); production paths must use
+/// [`on_gossip_aggregated_attestation`].
+pub fn on_gossip_aggregated_attestation_without_verification(
+    store: &mut Store,
+    aggregated: SignedAggregatedAttestation,
+) -> Result<(), StoreError> {
+    on_gossip_aggregated_attestation_core(store, aggregated, false)
+}
+
+fn on_gossip_aggregated_attestation_core(
+    store: &mut Store,
+    aggregated: SignedAggregatedAttestation,
+    verify: bool,
+) -> Result<(), StoreError> {
     validate_attestation_data(store, &aggregated.data)
         .inspect_err(|_| metrics::inc_attestations_invalid())?;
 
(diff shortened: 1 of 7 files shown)


crates/common/types/tests/ssz_spectests.rs
```diff
@@ -5,16 +5,18 @@ use ethlambda_types::primitives::HashTreeRoot;
 mod ssz_types;
 use ssz_types::{SszTestCase, SszTestVector, decode_hex, decode_hex_h256};
 
-const SUPPORTED_FIXTURE_FORMAT: &str = "ssz";
+/// Newer leanSpec releases name the format `ssz_test`; the pinned-commit
+/// fixtures still use `ssz`. Accept both.
+const SUPPORTED_FIXTURE_FORMATS: &[&str] = &["ssz", "ssz_test"];
```
MegaRedHand: If newer leanSpec releases use `ssz_test`, we should only support that

crates/blockchain/src/store.rs
```diff
@@ -392,6 +392,26 @@ pub fn on_gossip_attestation(
...
+///
+/// Only for spec tests whose fixtures carry mocked (placeholder) proofs
+/// (`proofSetting == 0`); production paths must use
+/// [`on_gossip_aggregated_attestation`].
+pub fn on_gossip_aggregated_attestation_without_verification(
+    store: &mut Store,
+    aggregated: SignedAggregatedAttestation,
+) -> Result<(), StoreError> {
+    on_gossip_aggregated_attestation_core(store, aggregated, false)
+}
```
MegaRedHand, line 409: There should be a better way to do this.

MegaRedHand (approved):
LGTM

MegaRedHand (approved):

PR #435, opened 2026-06-14
ci: publish docker image on push to main

### LINKED ISSUE #94: Publish docker image on each commit
We should make each commit to main update the `latest` Docker tag.

mvanhorn:

The `docker_publish` workflow was `workflow_dispatch`-only, so the published `latest` GHCR image only refreshed when someone manually dispatched the workflow and otherwise drifted behind `main`. The issue asks that every commit to `main` update the `latest` Docker tag automatically, which is what this change does.

## What Changed

The single file touched is `.github/workflows/docker_publish.yaml`. A `push` trigger on `main` now runs alongside the existing `workflow_dispatch`. Because `inputs.tags` is empty on a push event, a workflow-level `TAGS_INPUT: ${{ inputs.tags || 'latest' }}` env now provides the effective tag list, and the two `${{ inputs.tags }}` reads (the `Prepare tags` step and the `Create multi-arch manifests` step) consume `$TAGS_INPUT` instead.

Manual dispatch behaves exactly as before, since `TAGS_INPUT` falls through to whatever the user passed in `inputs.tags`, so custom comma-separated tags still expand unchanged. On a push to `main` the list resolves to `latest`, producing the same `latest` + `sha-<short>` manifest that a dispatch with `tags=latest` already produced. The one behavior change worth flagging for reviewers is that `latest` now moves on every `main` commit; nothing inside the repo consumes the image tag, so this only affects external consumers who pin `latest`.

## Tests Added / Run

No automated tests apply to a workflow trigger change. I confirmed the file still parses with `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/docker_publish.yaml'))"` and hand-traced the tag derivation: a dispatch with `tags=latest,v0.1.0` still expands both tags, while a push with empty `inputs.tags` resolves to `latest`.

## Related Issues / PRs

Closes #94

## ✅ Verification Checklist

- [x] Ran `make fmt` — N/A, this PR changes only workflow YAML and no Rust source
- [x] Ran `make lint` (clippy with `-D warnings`) — N/A, no Rust source changed
- [x] Ran `cargo test --workspace --release` — N/A, no Rust source changed; YAML validated via `yaml.safe_load` instead


MegaRedHand (changes requested):
Please address this comment: https://github.com/lambdaclass/ethlambda/pull/435#discussion_r3409318903

mvanhorn:
Addressed — added a concurrency group (`${{ github.workflow }}-${{ github.ref }}`, `cancel-in-progress: false`) so publish runs serialize and can't race into a mixed-arch `latest` manifest. Left `cancel-in-progress` false so an in-flight multi-arch push isn't cut off midway.

.github/workflows/docker_publish.yaml
```diff
@@ -12,9 +14,16 @@ permissions:
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: ${{ github.repository }}
+  # On push to main, inputs.tags is empty, so fall back to "latest" so the
+  # latest tag tracks main. Manual dispatch still honors custom tags.
+  TAGS_INPUT: ${{ inputs.tags || 'latest' }}
```
MegaRedHand: Let's use the "unstable" tag for this

```suggestion
  # On push to main, inputs.tags is empty, so fall back to "unstable" so it
  # tracks main. Manual dispatch still honors custom tags.
  TAGS_INPUT: ${{ inputs.tags || 'unstable' }}
```

MegaRedHand (approved):
Great PR! Thanks for your contribution 🙌

mvanhorn:
Thanks @MegaRedHand! Happy the docker publish workflow is in.

mvanhorn:
Appreciate it, @MegaRedHand! Docker images publishing on every push to main should simplify deployments.

PR #437, opened 2026-06-16
fix: align fork-choice finality and aggregate validation with leanSpec

MegaRedHand:
## Summary

CI (`make test`) is failing against the latest leanSpec fixtures. This fixes the **3** failing fork-choice / state-transition spec-test fixtures.

| Fixture | Root cause | Fix |
|---|---|---|
| `test_losing_fork_higher_finalized_does_not_latch` | `finalized` was latched monotonically (max-seen) in `on_block_core` | Derive `finalized` from the chosen head's post-state in `update_head` and pin it to the head's own ancestry (matches leanSpec `update_head`); keep `justified` monotonic via `advance_to` |
| `test_gossip_aggregated_attestation_empty_participants_rejected` | An aggregate naming no participants slipped through the mocked-proof path | Reject empty participants (`EmptyAggregationBits` / `EMPTY_AGGREGATION_BITS`) before bounds-check and verification, mirroring the spec's `to_validator_indices()` |
| `test_genesis_registry_holds_exact_validator_entries` | The stf runner's `PostState` (`deny_unknown_fields`) rejected the new post-state `validators` registry check | Parse `validators` and compare each entry's index and public keys |

## Why the finality change matters

Finality is **not** a monotonic max over everything ever seen. The spec derives the finalized checkpoint from the chosen head's post-state and pins it to the head's ancestry. When fork choice switches to a fork that finalized *less* than a now-losing fork, finalized must follow the head **downward**; otherwise a losing fork's finalization persists on a chain that never contained it.

## Notes

- Only the first two failures appeared in the most recent CI run (forkchoice spectests). The third (`stf` genesis) comes from an even-newer fixtures drop and is a test-runner schema gap, not a client bug; fixed here so the next CI run stays green.

## Testing

- `cargo test --workspace --release` — all green
- `make fmt`, `make lint` (`-D warnings`) — clean

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -100,29 +100,22 @@ pub fn update_head(store: &mut Store, log_tree: bool) {
 
 /// Resolve the finalized checkpoint that sits on `head`'s chain.
 ///
-/// Mirrors leanSpec `update_head`: take the finalized slot recorded in the
-/// head's post-state, then climb the head's parent chain to the block at that
-/// slot. Returns `None` (caller keeps the existing checkpoint) when no block is
-/// stored at that slot, which happens only for a checkpoint-sync anchor whose
-/// pre-anchor history is absent.
+/// Mirrors leanSpec `update_head`: the finalized checkpoint is the one recorded
+/// in the head's post-state. That checkpoint's root is, by construction, an
+/// ancestor of `head` (the state was produced by a transition over `head`'s
+/// own history), so it always stays on the canonical chain.
+///
+/// Returns `None` (caller keeps the existing checkpoint) when the finalized
+/// block is not in the store. This happens only for a checkpoint-sync anchor
+/// whose pre-anchor history we never downloaded, or before any block has
+/// finalized (genesis post-state names the zero root).
 fn recompute_finalized(store: &Store, head: H256) -> Option<Checkpoint> {
-    let finalized_slot = store.get_state(&head)?.latest_finalized.slot;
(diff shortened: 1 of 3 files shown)


crates/blockchain/src/store.rs
```diff
@@ -52,7 +52,27 @@ pub fn update_head(store: &mut Store, log_tree: bool) {
...
+    // transition over the head's own history), so it always stays on the
+    // canonical chain. When fork choice switches to a fork that finalized less
+    // than a now-losing fork, finalized must follow the head downward rather
+    // than latch the higher value (which would let a losing fork's finalization
+    // persist on a chain that never contained it).
+    //
+    // Adopt it only when the finalized block is actually in the DB. It is absent
+    // for a checkpoint-sync anchor whose pre-anchor history we never downloaded,
+    // or before any block has finalized (genesis post-state names the zero
+    // root); in those cases keep the existing checkpoint.
```
```suggestion
    // Override the store's latest finalized with the head state's
```

crates/blockchain/src/store.rs
```diff
@@ -406,6 +426,12 @@ fn on_gossip_aggregated_attestation_core(
     let num_validators = validators.len() as u64;
 
     let participant_indices: Vec<u64> = aggregated.proof.participant_indices().collect();
+    // An aggregate must name at least one signer. The spec rejects empty
+    // participants (EMPTY_AGGREGATION_BITS) when resolving validator indices,
+    // before any bounds check or signature verification.
```
```suggestion

```

pablodeymo (approved):

PR #442, opened 2026-06-18
fix(fork-choice): reject attestation slot before head block slot

MegaRedHand:
## What

Ports the head-consistency guard from [leanSpec PR #1020](https://github.com/leanEthereum/leanSpec/pull/1020) into `validate_attestation_data`.

A gossip attestation's slot records *when* the head was observed, so it cannot precede the slot of the head block it claims to have seen. We did not enforce this lower bound: the wire `slot` field was untethered from any real block, free to be set arbitrarily relative to the observed head.

## Why

| Property | Risk without the guard |
|----------|------------------------|
| Soundness | An unbounded slot violates 3SF attestation semantics: the slot is meant to anchor when the head was seen. |
| Liveness (spec context) | In leanSpec a crafted near-`2**64` slot overflowed the slot→interval constructor and crashed the node (remote DoS). |

**Note on the overflow half:** ethlambda's time check already uses `saturating_mul` (`data.slot.saturating_mul(INTERVALS_PER_SLOT)`), so a near-`u64::MAX` slot saturates to `u64::MAX` and is rejected cleanly as too-far-in-future rather than panicking. The Rust port was never crashable the way the unchecked Python constructor was, so no change to the time-check arithmetic is needed; the spec's switch to slot-unit comparison is mathematically equivalent to the existing interval comparison. This PR adds the genuinely missing piece: the semantic lower bound.

## Change

- New `StoreError::AttestationSlotBeforeHead { attestation_slot, head_slot }` rejection.
- Check `data.slot < data.head.slot` after the ancestry checks, before the time check (mirrors the spec's ordering).
- Two unit tests: a vote whose slot precedes its head (rejected as `AttestationSlotBeforeHead`), and a `u64::MAX` ceiling slot (rejected as `AttestationTooFarInFuture`, confirming no overflow).

Applies to both gossip paths via the shared `validate_attestation_data` (`on_gossip_attestation` and `on_gossip_aggregated_attestation`).

## Testing

```
cargo test -p ethlambda-blockchain --lib validate_attestation   # 5 passed
cargo fmt --all && cargo clippy -p ethlambda-blockchain --lib    # clean
```

https://claude.ai/code/session_018pvnQJkom3DobK9oSbK42T

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -225,8 +225,6 @@ fn validate_attestation_data(store: &Store, data: &AttestationData) -> Result<()
 
     // Head Consistency Check - A vote cannot have observed its head before that
     // head existed, so the vote's slot must not precede the head block's slot.
-    // This lower bound also keeps the wire slot clear of the overflow edge that a
-    // crafted near-`u64::MAX` slot would otherwise probe in the time check below.
     if data.slot < data.head.slot {
         return Err(StoreError::AttestationSlotBeforeHead {
             attestation_slot: data.slot,
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -163,7 +163,8 @@ fn checkpoint_is_ancestor(
 ///     3. The head must be at least as recent as source and target.
 ///     4. Checkpoint slots must match the actual block slots.
 ///     5. Source, target, and head must lie on one parent chain.
-///     6. The vote's slot must have started locally (a small disparity margin is allowed).
+///     6. The vote's slot cannot precede the slot of the head it claims to have seen.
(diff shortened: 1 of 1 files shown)


crates/blockchain/src/store.rs
```diff
@@ -222,6 +223,17 @@ fn validate_attestation_data(store: &Store, data: &AttestationData) -> Result<()
         return Err(StoreError::TargetNotAncestorOfHead);
     }
 
+    // Head Consistency Check - A vote cannot have observed its head before that
+    // head existed, so the vote's slot must not precede the head block's slot.
+    // This lower bound also keeps the wire slot clear of the overflow edge that a
+    // crafted near-`u64::MAX` slot would otherwise probe in the time check below.
```
```suggestion
```

pablodeymo (approved):

PR #443, opened 2026-06-18
Skip malformed live chain keys

samsamtrum:
LiveChain iterators currently assume every key has the slot/root layout and can panic if a malformed key is present.

This makes key decoding fail safely for invalid lengths and skips those entries in the read paths, while keeping valid entries usable.

### DIFF
--- a/crates/storage/src/store.rs
+++ b/crates/storage/src/store.rs
@@ -468,15 +468,23 @@ fn encode_live_chain_key(slot: u64, root: &H256) -> Vec<u8> {
     result
 }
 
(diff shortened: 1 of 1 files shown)


pablodeymo (changes requested):
I agree with greptile comment, the function should return a Result instead of an Option for better error handling.

MegaRedHand:
Thanks for bringing this up. We're closing the PR due to staleness, and opening #529 to tackle this in the future. Feel free, however, to open a new PR for this.

PR #444, opened 2026-06-18
feat(storage): diff-layer state storage with an in-memory LRU cache

### LINKED ISSUE #238: Implement diff-layers for state calculation

MegaRedHand:
Closes #238

## Summary

Replaces aggressive state pruning with a **diff-layer** storage model plus an in-memory cache, so the full state history stays available cheaply and no state is ever pruned. Builds on the block-signature pruning from #453 (merged to `main`): keeping block headers/bodies forever is what lets historical states be reconstructed.

## State storage

- Every non-genesis state is stored as a **parent-linked `StateDiff`** (`StateDiffs` table, **never pruned**), preserving full state history.
- Full-state **snapshots** (`States` table) are written **only at anchors**: the bootstrap state plus one per `SNAPSHOT_ANCHOR_INTERVAL` (1024-slot) window crossing. Also never pruned. There is no separate anchor table; an anchor is just a state whose slot crosses a 1024 boundary relative to its parent.
- `StateDiff` stores `slot`, both checkpoints, and the justification fields in full, plus the appended `historical_block_hashes` tail. `config`/`validators` come from the nearest ancestor snapshot (they never change); `latest_block_header` is read back from `BlockHeaders` (the stored state caches the real `state_root` there, so it matches byte-for-byte). These assumptions are documented on `StateDiff::from_base`.
- `get_state` resolves in three tiers: **(1)** in-memory LRU cache, **(2)** a full snapshot in `States`, else **(3)** **reconstruct** by walking `base_root` back to the nearest ancestor snapshot and replaying appended tails forward. Results are memoized in the cache (`STATE_CACHE_CAPACITY = 32`) so recent/repeated reads stay hot.
- `DiffBase` captures the parent `(root, hbh_len, slot)` before it is consumed into the post-state; `StateDiff`/`DiffBase` live in the storage crate.

## Why this over snapshot eviction

The earlier approach pruned old snapshots on a hot window and scanned the whole `States` table to do it. Anchors are kept forever and spread across all slots, so a `slot||root`-keyed range scan still touches every anchor: bounding the scan doesn't help. Keeping diffs-only (never pruned) plus an LRU for hot reads removes the periodic scan entirely and keeps full history, at the cost of on-demand reconstruction for cold, non-anchor states.

## Tests

- `StateDiff` build/SSZ round-trip; state reconstruction (single-diff and multi-diff) verified via a cold store with an empty cache so the diff walk is actually exercised; snapshot-only-on-boundary-crossing; bootstrap-anchor snapshot.
- Storage + blockchain suites green; clippy `-D warnings` clean.

## Validation

- Local 4-node devnet, ~52-slot soak: all nodes finalize and stay byte-identical (same head/justified/finalized), aggregated attestations packed, **zero reconstruction errors / panics**.
- A >1024-slot soak is running to also exercise the second-anchor snapshot write and anchor-to-anchor reconstruction (the short soak only crosses the bootstrap anchor).
- Migration: existing DBs keep their full states as snapshots; diffs start accruing going forward (no backfill).


crates/storage/src/state_diff.rs
```diff
@@ -0,0 +1,267 @@
...
+        target.slot = 2;
+        let mut hbh: Vec<H256> = base.historical_block_hashes.to_vec();
+        hbh.push(h256(9));
+        target.historical_block_hashes = hbh.try_into().unwrap();
+
+        let diff = StateDiff::from_base(h256(1), base_len, target);
+        let bytes = diff.to_ssz();
+        let decoded = StateDiff::from_ssz_bytes(&bytes).expect("decodes");
+        assert_eq!(diff, decoded);
+    }
```
MegaRedHand, line 374:
```suggestion
```

crates/storage/src/state_diff.rs
```diff
@@ -0,0 +1,358 @@
...
+            justifications_validators,
+            ..
+        } = target;
+
+        let hbh = historical_block_hashes.into_inner();
+        assert!(
+            hbh.len() >= base_hbh_len,
+            "target historical_block_hashes shorter than base: {} < {base_hbh_len}",
+            hbh.len()
+        );
```
pablodeymo: should we try to avoid this assert?

crates/storage/src/state_diff.rs
```diff
@@ -0,0 +1,358 @@
...
+            parent_root: h256(51),
+            state_root: h256(99),
+            body_root: h256(88),
+        }
+    }
+
+    /// Build a diff against `base_root` that appends `appended` to
+    /// `historical_block_hashes`. Absolute fields default; tests override the
+    /// ones they assert on.
+    fn diff_at(base_root: H256, slot: u64, appended: Vec<H256>) -> StateDiff {
```
pablodeymo: this function is not necessary inside test module

crates/storage/src/store.rs
```diff
@@ -1155,26 +1130,107 @@ impl Store {
...
+        let mut diffs: Vec<StateDiff> = Vec::new();
+        let mut cursor = *root;
+        let snapshot = loop {
+            if let Some(snapshot) = self.get_ssz::<State>(Table::States, &cursor) {
+                break snapshot;
+            }
+            let diff = self.get_ssz::<StateDiff>(Table::StateDiffs, &cursor)?;
+            cursor = diff.base_root;
+            diffs.push(diff);
+        };
```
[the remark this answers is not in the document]
pablodeymo: we should apply this strategy

pablodeymo (approved):

crates/blockchain/src/store.rs
```diff
@@ -586,7 +586,7 @@ fn on_block_core(
             .expect("update_checkpoints should succeed");
     }
 
-    // Store signed block and state
+    // Store signed block and state (as a parent-linked diff + snapshot)
```
```suggestion
    // Store signed block and state
```

PR #445, opened 2026-06-18
feat(blockchain): pre-build proposer block one interval early

### LINKED ISSUE #441: Start block building during the last slot interval
We can start to build a block during the last interval of a slot, when we know we are the proposer of the next slot. This doubles the time budget we have for block production.

MegaRedHand:
## What

Proposers do the heavy leanVM block-building work (attestation compaction + the Type-1 → Type-2 merge, ~1–2s) **synchronously at interval 0** today, squeezed into the slot's first interval. This builds the next slot's block **one interval early** — at the previous slot's interval 4, synchronously on the blockchain actor — and publishes it at interval 0 after a revalidation. On any mismatch or failure it falls back to the existing synchronous build, so there is **no regression**.

## How

- **Interval 4 (of slot S-1):** if one of our validators proposes slot S, `prebuild_block` builds + signs + merges the full block against the current (read-only) head and stashes it as a `PreparedBlock`.
- **Interval 0 (of slot S):** `propose_block` publishes the prepared block iff it is still valid (`prebuilt_block_is_usable`: same slot/proposer, parent still the canonical head, and the built justified slot not behind the store's). Otherwise it rebuilds synchronously.
- Shared `assemble_signed_block` / `process_and_publish_block` helpers back both the pre-build and the fallback path.
- Signing stays on the actor throughout: XMSS keys are stateful (`sign_block_root` takes `&mut self`), so the build cannot be moved off-thread without risking OTS-index reuse.

## Finalization fix (included)

Capturing the pre-build head via `get_proposal_head` — which is **not** read-only, it ticks the store clock — one interval early advanced the clock prematurely and stalled finalization. Fixed by capturing the head read-only via `store.head()`.

## Validation (local devnet: 4 ethlambda nodes, 1 aggregator)

- Finalization healthy (finalized lag ~3 slots).
- Near-100% pre-build hit rate; **zero** stale discards, **zero** skipped ticks, **zero** errors.

## Tradeoff

Building synchronously blocks the actor for the build duration at interval 4. This is acceptable: between interval 4 and the next slot the actor has no other consensus-critical duty, and the devnet run showed zero skipped ticks.

## Tests

`make fmt` + `cargo clippy -p ethlambda-blockchain --all-targets -- -D warnings` clean; lib tests pass, including the new `block_builder::prebuild_tests` covering the usability predicate.


crates/blockchain/src/lib.rs
```diff
@@ -436,43 +530,69 @@ impl BlockChainServer {
...
+        }
 
-        coverage::emit_proposal_coverage(
-            &self.store,
-            self.attestation_committee_count,
-            block.body.attestations.iter(),
-        );
+        if let Some((signed_block, _)) = self.build_signed_block(slot, validator_id, head_root) {
+            self.process_and_publish_block(slot, validator_id, signed_block, "Published block");
+        }
```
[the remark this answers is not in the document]
MegaRedHand: The fallback only triggers if no block was built and no block was signed. There's no key reuse.

crates/blockchain/src/lib.rs
```diff
@@ -441,28 +456,50 @@ impl BlockChainServer {
...
+            };
 
         coverage::emit_proposal_coverage(
             &self.store,
             self.attestation_committee_count,
             block.body.attestations.iter(),
         );
 
-        // Sign the block root with the proposal key
+        // Sign the block root with the proposal key.
```
pablodeymo:
```suggestion
        // Sign the block root with the proposal key
```

crates/blockchain/src/lib.rs
```diff
@@ -565,23 +601,44 @@ impl BlockChainServer {
...
-        drop(type_one_proofs);
         let signed_block = SignedBlock {
             message: block,
             proof,
         };
 
-        // Process the block locally before publishing
+        // Stop timing here: the build is done, and the alignment wait below must
+        // not count toward the block-building metric.
+        drop(_timing);
```
pablodeymo: I would suggest to rename the variable without _

crates/blockchain/src/lib.rs
```diff
@@ -441,20 +456,42 @@ impl BlockChainServer {
...
-        };
+        let (block, type_one_proofs, _post_checkpoints) =
+            match store::produce_block_with_signatures(&mut self.store, slot, validator_id) {
+                Ok(built) => built,
+                Err(err) => {
+                    error!(%slot, %validator_id, %err, "Failed to build block");
+                    metrics::inc_block_building_failures();
+                    return;
+                }
+            };
```
pablodeymo: I prefer to keep the let ... else construction

crates/blockchain/src/lib.rs
pablodeymo: I would prefer to have a match here, to group all the intervals in the same construction.

pablodeymo (changes requested):
I've put some comments

crates/blockchain/src/lib.rs
MegaRedHand: done

crates/blockchain/src/lib.rs
MegaRedHand: done

crates/blockchain/src/lib.rs
MegaRedHand: done

pablodeymo (approved):

PR #448, opened 2026-06-18
fix(storage):make init_store return Result

d4m014:
First step of migrating `Store` functions to return `Result` instead of panicking.
Changes `init_store` to return `Result<Self, Error>`, wrapping the return value in `Ok(...)`. All existing `.expect()` calls inside the function body are left as-is for now, error propagation will come in follow-up PRs.

## Related Issues / PRs
- Part of #306 

## ✅ Verification Checklist

- Ran `make fmt` — clean
- Ran `make lint` (clippy with `-D warnings`) — clean
- Ran `cargo test --workspace --release` — all passing

### DIFF
--- a/crates/storage/src/store.rs
+++ b/crates/storage/src/store.rs
@@ -533,7 +533,7 @@ impl Store {
 
         Ok(
             Self::init_store(backend, anchor_state, Some(anchor_block.body))
-                .expect(" store initialization should succeed in get_forkchoice_store"),
+                .expect("store initialization should succeed in get_forkchoice_store"),
         )
     }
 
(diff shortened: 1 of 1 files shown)


crates/storage/src/store.rs
```diff
@@ -579,7 +579,7 @@ impl Store {
         backend: Arc<dyn StorageBackend>,
         mut anchor_state: State,
         anchor_body: Option<BlockBody>,
-    ) -> Self {
+    ) -> Result<Self, Error> {
```
MegaRedHand, line 582: Just a heads-up, but we probably need some other `crate::error::Error` enum for this, which wraps this `crate::api::Error`.

MegaRedHand (approved):
LGTM. Left a comment regarding future direction

PR #450, opened 2026-06-19
refactor(blockchain): group on_tick conditionals by interval

MegaRedHand:
Split out of #445 (proposer pre-build). **Stacked on #449** (`gate_proposer` removal) — review/merge that first; this PR's base retargets to `main` once #449 lands.

`on_tick` ran its per-interval blocks out of order (interval-4 snapshot, then tick, then 2, 0, 1). This regroups them into ascending interval order behind `==== interval N ====` markers so the slot timeline reads top-to-bottom and lines up with the duty schedule.

**Pure reorder, behavior unchanged.**

### One block intentionally does *not* move
The interval-4 `new_payloads` snapshot stays **ahead** of `store::on_tick`. At interval 4 the tick runs `accept_new_attestations → promote_new_aggregated_payloads()`, which drains `new_payloads`; snapshotting after the tick would capture an already-drained set and silently break the post-block coverage report's "timely" cohort. A comment now pins that ordering constraint.

> Note: commit `7474c15` on #445 currently moves this snapshot *into* the post-tick interval-4 group, which hits exactly that regression (observability-only, so devnet doesn't surface it). #445 should adopt this PR's placement when it rebases.

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -251,11 +251,6 @@ impl BlockChainServer {
         // (interval 4), so the post-block report for this round sees its
         // "timely" cohort just before it is promoted out of `new_payloads`.
         //
-        // This MUST stay ahead of `store::on_tick` below: the interval-4 tick
-        // promotes `new_payloads` out, so snapshotting afterwards would capture
-        // an already-drained set. It is the one interval action that cannot live
-        // in its grouped block downstream.
-        //
         // Only interval 4 — not the proposer's interval-0 promote. By interval 0
         // the round's votes have already been promoted at the previous slot's
         // interval 4; `new_payloads` then holds only stragglers, and snapshotting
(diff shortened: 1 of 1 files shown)


crates/blockchain/src/lib.rs
```diff
@@ -245,10 +245,17 @@ impl BlockChainServer {
...
+
         // Snapshot the pre-merge `new_payloads` set at the end-of-slot promote
         // (interval 4), so the post-block report for this round sees its
         // "timely" cohort just before it is promoted out of `new_payloads`.
         //
+        // This MUST stay ahead of `store::on_tick` below: the interval-4 tick
+        // promotes `new_payloads` out, so snapshotting afterwards would capture
+        // an already-drained set. It is the one interval action that cannot live
+        // in its grouped block downstream.
+        //
```
```suggestion
```

pablodeymo (approved):

PR #459, opened 2026-06-24
feat(rpc): add GET /lean/v0/attestations for validator participation

MegaRedHand:
Adds `GET /lean/v0/attestations`, which returns the currently known (fork-choice-active) attestations. Optionally filterable by slot or validator index.

Enables explorer tooling to surface validator participation rates and diagnose finality stalls. Has unit tests and passed clippy.

Stacked on #454.


MegaRedHand:
Actually, this isn't needed

PR #460, opened 2026-06-24
feat(rpc): add GET /lean/v0/events SSE stream (head/block/finalized_checkpoint)

MegaRedHand:
Adds a Server-Sent Events endpoint (`GET /lean/v0/events`) that pushes real-time chain events to subscribers. All subscribers receive every event; there is no server-side topic filtering.

The three event types are `head`, `block`, and `finalized_checkpoint`.

Also adds the `ChainEvent` broadcast channel in the blockchain actor and wires it through `main.rs` so the RPC server receives events without polling. Enables reactive explorer UIs and monitoring tools. Has unit tests and passed clippy.

Stacked on #454.

## Events not yet included (candidate follow-ups)

This PR ships `head`/`block`/`finalized_checkpoint`. The following are additional `ChainEvent` variants worth adding. Each is state the actor **already computes**; surfacing it is a new enum variant plus one `events.send(...)` at the existing site, so these are cheap follow-ups rather than new machinery.

| Event | Value | Already computed at |
|-------|-------|--------------------|
| `attestation` | Per-attestation seen-time (enables xatu-style propagation timing; the seen-time already exists on the gossip-accept path) | attestation accept pipeline |
| `chain_reorg` | Standard Beacon-API topic; reorgs are already detected with old/new head + depth, but only logged + metriced | `store.rs:50-53` |
| `justified_checkpoint` | 3SF-mini justifies *before* it finalizes; we stream `finalized` but not `justified`, which is the more useful stall-detection signal | `store.rs:580-585` |
| `safe_target` | Lean-specific interval-3 fork-choice output; exposed in the `/fork_choice` REST endpoint but not on the stream | `store.rs:110`, `lib.rs:353` |
| `aggregate` | Lean's interval-2 aggregation step; useful for aggregator monitoring | `lib.rs` aggregation module |

Timing nuance: `head`/`block` fire at **import** time, so there is no gossip-received signal (Beacon-API `block_gossip` analog) to measure block-arrival latency separately from processing.

Intentionally out of scope (not part of lean's domain): `voluntary_exit`, `proposer_slashing`, `attester_slashing`, `bls_to_execution_change`, `blob_sidecar`, `contribution_and_proof`, `light_client_*`, `payload_attributes`.


MegaRedHand:
Superseded by #516 and co.

PR #467, opened 2026-06-24
feat(types): carry proposer signature outside the block proof

### LINKED ISSUE #471: Don't aggregate proposer signature with attestation proofs
We currently aggregate the proposer signature with the rest of the proofs, which is unnecessary: the proposer signature is ~2KB and takes >1ms to verify, so having it travel alongside the block proof adds almost no overhead.

What's more, avoiding the aggregation of it reduces aggregation time by about 50% in the common case of 1 attestation blocks (since we aggregate 1 proof instead of 2, so half the work). It would also make empty blocks trivial to produce and propagate over the network, since they require no aggregation and weigh only as much as the signature plus the block (we assume the block proof is empty when there are no attestations).

MegaRedHand:
Closes #471

## What

Moves the **proposer signature outside the block proof**. `SignedBlock.proof` becomes a two-field `BlockProof`:

```
before:  block-proof = aggregate([prop-sig, att0, att1])
after:   block-proof = (prop-sig, aggregate([att0, att1]))

before (no atts):  aggregate([prop-sig])
after  (no atts):  (prop-sig, empty-proof)
```

```rust
struct SignedBlock { message: Block, proof: BlockProof }

struct BlockProof {
    proposer_signature: XmssSignature,         // raw XMSS sig (reuses the existing fixed-size wire type)
    attestation_proof:  MultiMessageAggregate, // Type-2 over body attestations only (empty if none)
}
```

The proposer signature reuses the **existing `XmssSignature` type** (`SszVector<u8, SIGNATURE_SIZE>`) already carried by `SignedAttestation.signature` — no new wrapper type. `sign_block_root` already returns an `XmssSignature`, so `propose_block` carries it verbatim; genesis anchors use the existing `blank_xmss_signature()` placeholder.

## Why

Previously the proposer signature was wrapped as a singleton Type-1 and merged into a single block Type-2 alongside every attestation, so **even a block with zero attestations required a prover call**, and the merge could only run once the block root was known.

Decoupling the proposer:
- lets the attestation aggregate be built independently of the block root — a prerequisite for **proposer prebuild**;
- removes all prover work from the empty-attestation case;
- the proposer signature is now verified directly with the hash-based XMSS verifier (`ValidatorSignature::is_valid`), never entering the lean-multisig prover/verifier.

## Changes

| Area | Change |
|------|--------|
| `types/block.rs` | New `BlockProof { proposer_signature: XmssSignature, attestation_proof }`; `SignedBlock.proof: BlockProof`. `Default` = blank XMSS placeholder + empty aggregate |
| `blockchain/lib.rs` | `propose_block`: carry the raw `XmssSignature` directly; aggregate attestation Type-1s only (skip the merge entirely when there are no attestations) |
| `blockchain/store.rs` | `verify_block_signatures`: verify raw proposer sig vs `proposal_pubkey`, then verify the attestation Type-2 over attestation components only; reject a stray aggregate on an attestation-less block (`UnexpectedAttestationProof`) |
| `blockchain/reaggregate.rs` | Drop the proposer component from the split layout; split from `attestation_proof` |
| `storage/store.rs` | Persist / reconstruct `BlockProof` in the `BlockSignatures` row |
| `net/p2p`, `net/rpc` | Construction-site updates |

## Test status

- `cargo check --workspace --all-targets`: ✅
- `cargo clippy --workspace --all-targets -- -D warnings`: ✅
- Library unit tests (types/storage/blockchain/rpc/p2p `--lib`): ✅ (incl. `SignedBlock` SSZ round-trip with a populated proposer signature + blank-default)
- **Spec tests fail by design**: this diverges from the leanSpec #799 single-merged-proof wire format, so `signature_spectests` / `ssz_spectests` fail against the current cross-client fixtures until those are regenerated. The fixture crate carries only a minimal type-only compile fix (no adapter logic).

## Open questions

- Wire/cross-client: needs a matching leanSpec change + regenerated fixtures before this can interop.

🤖 Draft — opened for review of the restructure.


MegaRedHand:
The change is included in #604. We may also include this standalone, but we're already past the prototyping stage here

PR #472, opened 2026-06-25
feat(blockchain): cap attestation packing at 2/3 supermajority

MegaRedHand:
## What

Block building stops packing attestation proofs for an `AttestationData` entry as soon as its voters cross the 2/3 supermajority, instead of stitching partial proofs together until ~100% coverage.

## Why

`select_attestations` → `extend_proofs_greedily` greedily covered ~all available voters per entry. Any entry left with multiple partial proofs then paid for a recursive proof merge (`aggregate_proofs`, the leanVM SNARK) in `compact_attestations`. Per profiling, that compaction is the dominant cost of building blocks that carry attestations.

Consensus only needs a 2/3 supermajority: once a target reaches it, the target is justified and `is_valid_vote` rejects further votes for it. So every proof packed beyond 2/3 was added to an expensive merge for zero consensus gain.

## How

One uniform cutoff in `extend_proofs_greedily`, no tier branching:

- Stop packing once `prior_voters + covered` cross 2/3 (`is_supermajority` helper, shared with `score_entry`). `prior_voters` = validators that already voted for this target in the projected state or earlier rounds.
- When the largest single proof already exceeds 2/3, only that proof is selected → `compact_attestations` takes its single-proof fast path and skips `aggregate_proofs` entirely.
- A sub-supermajority (Build-tier) entry can't reach 2/3 by definition, so the cutoff never fires and it keeps full coverage, unchanged.
- `extend_proofs_greedily` now returns the voters it actually packed; `select_attestations` updates its vote projection from that set (not `score_entry`'s full union) so the projection stays consistent with the real block.
- Skipped proofs are traced at `trace` level (`reason = supermajority_reached | no_new_coverage`).

## Safety

| Concern | Why dropping voters above 2/3 is safe |
|---|---|
| Fork choice | LMD GHOST weight is sourced from **gossip** votes (`extract_latest_known_attestations`), not block-included attestations. Head weight is unchanged. |
| Finalization | Justification is binary at 2/3; reaching exactly 2/3 justifies, and post-justification votes are rejected. |
| Rewards | None exist in lean consensus; no penalty for excluding a validator. |

Tradeoff: a non-justifying target accumulates one proposer's selection per block rather than the full gossip union, converging over slots. No safety impact.

## Tests

- New: single proof already >2/3 → 1 proof; two needed to cross → 2, third dropped; sub-2/3 entry → full coverage; prior voters count toward the cutoff.
- Updated `extend_proofs_greedily_*` tests for the new signature (high `validator_count` so they keep testing overlap/exhaustion).
- Existing `build_block_*` integration tests unchanged.
- `make fmt`, `make lint`, and `cargo test -p ethlambda-blockchain --lib` (42 pass) all green.

### DIFF
--- a/crates/blockchain/src/block_builder.rs
+++ b/crates/blockchain/src/block_builder.rs
@@ -163,7 +163,7 @@ fn select_attestations(
     let mut processed_data_roots: HashSet<H256> = HashSet::new();
 
     for _round in 0..MAX_ATTESTATIONS_DATA {
-        let Some((data_root, score, new_voters)) =
+        let Some((data_root, score)) =
             pick_best_candidate(&chain, &processed_data_roots, &projected)
         else {
             trace!(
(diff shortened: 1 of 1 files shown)


MegaRedHand:
Running a devnet with these changes resulted in blocks with compaction dropping down from ~100% to 75%. Outright disabling compaction (#473) resulted in much better results, so we are going for that instead.

PR #473, opened 2026-06-26
feat(blockchain): disable attestation compaction, keep single best proof

MegaRedHand:
## What

An **alternative** to PR #472 (the 2/3 supermajority *cutoff*). This PR disables compaction entirely: block building packs exactly **one proof per `AttestationData`** (the best one) and never merges proofs.

The two PRs are independent (both branch from `main`) so they can be compared/benchmarked; only one should be merged.

| | #472 (cutoff) | This PR (best-proof) |
|---|---|---|
| Proofs per data | minimal to cross 2/3 (may merge a few) | exactly 1, never merge |
| `compact_attestations` | kept, fires less | **removed** |
| Per-block coverage | ≥ 2/3 when reachable | one proof's worth |
| Justification of a fragmented target | this block if proofs cross 2/3 | may take several slots |

## Why

The recursive XMSS merge (`compact_attestations` → `aggregate_proofs`) is the dominant cost of building blocks that carry attestations. Removing it makes the build path much cheaper, at the cost of per-block vote coverage: a target whose 2/3 is split across multiple gossip proofs may take more slots to justify, but each block is built with no recursive proof merge.

## How

- `select_best_proof` replaces `extend_proofs_greedily`: picks the single proof covering the most validators **not already counted** for the target (`prior_voters`), so a proof that is mostly prior voters loses to an all-new one. Ranking iterates the proof slice in order, so ties are deterministic.
- It returns the voters it actually packed; `select_attestations` updates its projection from that set, keeping the projection consistent with the block.
- `build_block` drops the compact step. `score_entry` / `pick_best_candidate` no longer return the per-entry voter set.
- Removed: `compact_attestations`, `union_aggregation_bits`, and the `aggregate_proofs` import. (`StoreError::SignatureAggregationFailed` is retained; it is part of the public error enum.)
- The cross-data Type-2 block proof built in `propose_block` is **unchanged** — that is the block signature, not compaction.

## Safety

Same basis as #472: fork-choice weight is gossip-sourced (not block-sourced), justification is binary at 2/3, and there are no inclusion rewards. Dropping voters from a block changes neither head selection nor finalization.

## Tests

- `select_best_proof` unit tests: packs the single largest proof; ranks by net-new over `prior_voters`; packs nothing when no proof adds a new voter.
- `build_block_packs_single_best_proof_and_justifies`: end-to-end — a data entry with multiple proofs packs only the best one and still justifies the target through the real STF.
- `make fmt`, workspace clippy `-D warnings`, and `cargo test -p ethlambda-blockchain --lib` (38 pass) all green.

### DIFF
--- a/crates/blockchain/src/block_builder.rs
+++ b/crates/blockchain/src/block_builder.rs
@@ -1,4 +1,4 @@
-//! Block building: select attestations, compact, and seal state root.
+//! Block building: select attestations and seal state root.
 //!
 //! The selection algorithm is a tiered greedy modeled on Prysm's
 //! `sortByProfitability`. Each round scores remaining candidates against a
@@ -8,6 +8,10 @@
 //! incrementally so dependent attestations become eligible on the next round
 //! without re-running the STF. The final STF runs once after selection to
 //! seal `state_root`.
+//!
+//! Compaction is disabled: each `AttestationData` is packed with a single
+//! proof (the best one, see `select_best_proof`), so no recursive proof merge
+//! runs during block building.
 
 use std::{
     cmp::Reverse,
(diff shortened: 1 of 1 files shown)


MegaRedHand:
Closing this. We're going to put compaction during block building behind a CLI flag, in a new PR.

PR #480, opened 2026-06-29
feat(blockchain): warn instead of failing on unclosed justified divergence

MegaRedHand:
## What

`produce_block_with_signatures` checks that a produced block's justified
slot does not lag the store's justified checkpoint (leanSpec #595). When it
did, the function returned `StoreError::JustifiedDivergenceNotClosed`, which
aborted block production for that slot.

This PR removes that error variant and replaces the hard failure with a
warn-level log, then publishes the block anyway.

## Why

Aborting production on this condition halts the chain rather than recovering
from it. The proposer cannot re-justify the head because validators source
from the store-justified checkpoint, so every subsequent proposal hits the
same divergence and block production freezes indefinitely (observed as a
"did not converge" stall).

Publishing a block whose justified slot lags by one fork is the lesser evil:
the chain keeps making progress and pool attestations get a chance to close
the divergence on later slots. The divergence is now surfaced as a warning
instead of silently halting the node.

This mirrors the upstream spec change in
leanEthereum/leanSpec#1168.

## Changes

- `crates/blockchain/src/store.rs`: replace the `return Err(JustifiedDivergenceNotClosed)` with a `warn!` log; remove the now-unused `StoreError::JustifiedDivergenceNotClosed` variant.
- `crates/blockchain/state_transition/src/lib.rs`: reword a test doc comment that referenced the removed error type by name.

## Testing

- `cargo build -p ethlambda-blockchain` ✅
- `cargo clippy -p ethlambda-blockchain --all-targets` ✅ (no warnings)

## References

- Upstream spec change: leanEthereum/leanSpec#1168

### DIFF
--- a/crates/blockchain/state_transition/src/lib.rs
+++ b/crates/blockchain/state_transition/src/lib.rs
@@ -710,8 +710,8 @@ mod tests {
     /// unconditional `state.latest_justified = target` assignment caused the
     /// post-state to end at `latest_justified.slot = 27974`. Because the
     /// store had already latched `latest_justified = 27978` from importing a
-    /// fork block, every subsequent proposal produced a block whose justified
-    /// slot lagged the store and the chain froze.
+    /// fork block, every subsequent proposal failed
+    /// `JustifiedDivergenceNotClosed` and the chain froze.
     ///
     /// Compressed setup: finalized=0, source=3 (justified), targets in body
     /// order 4 / 9 / 6 — all justifiable from finalized=0 (Δ=4 ≤ 5, Δ=9=3²,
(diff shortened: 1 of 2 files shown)


crates/blockchain/state_transition/src/lib.rs
```diff
@@ -710,8 +710,8 @@ mod tests {
     /// unconditional `state.latest_justified = target` assignment caused the
     /// post-state to end at `latest_justified.slot = 27974`. Because the
     /// store had already latched `latest_justified = 27978` from importing a
-    /// fork block, every subsequent proposal failed
-    /// `JustifiedDivergenceNotClosed` and the chain froze.
+    /// fork block, every subsequent proposal produced a block whose justified
+    /// slot lagged the store and the chain froze.
```
MegaRedHand:
```suggestion
    /// fork block, every subsequent proposal failed
    /// `JustifiedDivergenceNotClosed` and the chain froze.
```

pablodeymo (approved):

PR #495, opened 2026-07-03
feat: add block root slot index

### LINKED ISSUE #355: Add block number to block root index
For blocks-by-range requests we need to find the blocks for the given slots. We should add an index to our DB so the search is efficient, instead of the current sequential loop over all blocks.

dicethedev:

- Adds a canonical slot-to-block-root index to storage.
- BlocksByRange currently has to walk backward from the head and collect roots by slot.
- This PR makes range lookups efficient by reading canonical block roots directly by slot.

## What Changed

- Added a new `BlockRoots` storage table:
  - key: slot / block number
  - value: canonical block root
- Registers the new table for in-memory and RocksDB backends.
- Initializes the index for anchor/checkpoint stores.
- Maintains the index when fork-choice head changes.
- Handles reorgs by deleting old canonical slot entries and inserting the new canonical branch.
- Rebuilds the index on DB restore if the persisted index is missing.
- Cleans index entries when old block data is pruned.
- Updates BlocksByRange handling to use `Store::get_block_root_by_slot` instead of walking parent headers.

- The index points only to canonical block roots.
- Side-fork blocks do not appear in BlocksByRange responses.
- Skipped slots remain absent from the index and are skipped in responses.
- Reorgs update the index to match the new canonical chain.
- Existing persisted DBs can rebuild the index from the current head chain.
- Pruned blocks have their canonical index entries removed when applicable.

## Tests Added / Run

- Added storage tests for:
  - canonical slot index updates
  - skipped slots
  - side forks
  - reorgs
  - DB restore index rebuild
  - pruning cleanup
- Existing BlocksByRange test now exercises indexed lookup.

## Related Issues / PRs

- Closes #355
- Related to #348
- Related to #351

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


pablodeymo:
can you solve this conflicts?

dicethedev:
> can you solve this conflicts?

You can review again

crates/storage/src/store.rs
```diff
@@ -894,6 +924,99 @@ impl Store {
...
+
+        let mut entries = Vec::new();
+        let mut root = self.head()?;
+        while !root.is_zero() {
+            let Some(header) = self.get_block_header(&root)? else {
+                break;
+            };
+            entries.push((encode_block_root_key(header.slot), root.to_ssz()));
+            root = header.parent_root;
+        }
```
MegaRedHand: This iterates the chain all the way back to the genesis, right? I see we're using it only at startup, but it's still not a good solution. Let's just assume any existing DB has `Table::BlockRoots` properly built, and only initialize the genesis in this case.

crates/storage/src/store.rs
```diff
@@ -894,6 +924,99 @@ impl Store {
...
+                continue;
+            }
+            if new_root.is_zero() {
+                let header = self
+                    .get_block_header(&old_root)?
+                    .expect("old canonical block header exists");
+                deletes.push(encode_block_root_key(header.slot));
+                old_root = header.parent_root;
+                continue;
+            }
```
MegaRedHand: When are roots zero?

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -269,29 +269,10 @@ fn canonical_blocks_by_range(store: &Store, start_slot: u64, count: u64) -> Vec<
...
-    }
-
     (start_slot..=end_slot)
         .filter_map(|slot| {
-            let root = roots_by_slot.get(&slot)?;
-            store.get_signed_block(root).ok().flatten()
+            let root = store.get_block_root_by_slot(slot)?;
+            store.get_signed_block(&root).ok().flatten()
         })
         .collect()
```
MegaRedHand: Let's make this whole part a `get_signed_blocks_by_slot_range`. That way we can further optimize it inside the DB, in the future, when/if it's needed.

MegaRedHand (changes requested):
Left some comments

crates/storage/src/store.rs
dicethedev: They should only become zero if the traversal reaches the anchor/genesis parent (`parent_root == H256::ZERO`). Under normal fork-choice updates both heads should share an ancestor before that, so this is defensive. Tho, I can remove the zero-root branches if we prefer to rely on the invariant that canonical head updates always converge.

crates/storage/src/store.rs
dicethedev: Done. So I remove the startup rebuild path and now assume persisted DBs already have BlockRoots populated. Fresh stores still initialize the anchor/genesis block root through init_store.

crates/net/p2p/src/req_resp/handlers.rs
dicethedev: You can review again

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -269,31 +269,9 @@ fn canonical_blocks_by_range(store: &Store, start_slot: u64, count: u64) -> Vec<
...
-
-    (start_slot..=end_slot)
-        .filter_map(|slot| {
-            let root = roots_by_slot.get(&slot)?;
-            store.get_signed_block(root).ok().flatten()
-        })
-        .collect()
+    store
+        .get_signed_blocks_by_slot_range(start_slot, end_slot)
+        .unwrap_or_default()
```
MegaRedHand, line 282: We should log on error

crates/storage/src/store.rs
MegaRedHand: I think we should remove them if possible. If this is a case that can happen, then let's keep it though

crates/storage/src/store.rs
```diff
@@ -894,6 +920,68 @@ impl Store {
...
+        Ok((deletes, entries))
+    }
+
+    /// Return the canonical block root at `slot`.
+    pub fn get_block_root_by_slot(&self, slot: u64) -> Option<H256> {
+        let view = self.backend.begin_read().expect("read view");
+        view.get(Table::BlockRoots, &encode_block_root_key(slot))
+            .expect("get block root")
+            .map(|bytes| H256::from_ssz_bytes(&bytes).expect("valid block root"))
+    }
```
MegaRedHand: Let's remove this

crates/storage/src/store.rs
```diff
@@ -1182,6 +1270,28 @@ impl Store {
...
+    ) -> Result<Vec<SignedBlock>, Error> {
+        let mut blocks = Vec::new();
+        for slot in start_slot..=end_slot {
+            let Some(root) = self.get_block_root_by_slot(slot) else {
+                continue;
+            };
+            if let Some(block) = self.get_signed_block(&root)? {
+                blocks.push(block);
+            }
+        }
```
MegaRedHand, line 1276: Here we can inline `get_block_root_by_slot` and `get_signed_block` and reuse the `StorageReadView` for the whole loop

crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -269,31 +269,18 @@ fn canonical_blocks_by_range(store: &Store, start_slot: u64, count: u64) -> Vec<
...
+        Ok(blocks) => blocks,
+        Err(err) => {
+            warn!(
+                start_slot,
+                end_slot,
+                ?err,
+                "Failed to get signed blocks by slot range"
+            );
+            Vec::new()
         }
```
MegaRedHand: We can use an `inspect_err` here and keep the `unwrap_or_default`

MegaRedHand (approved):

PR #496, opened 2026-07-03
refactor: represent slot intervals with enum

### LINKED ISSUE #469: Use an enum to represent the different intervals in `BlockChainServer::on_tick`
We currently use an integer, which can only take values from 0..5. We should instead have the interval be represented with an enum, having the integer slot -> enum interval conversion be a constructor for it.

There are multiple places where we already match on the interval, and we should replace those with the new enum. Having the enum let's us type check these so any change to the intervals fails at compile time. It also documents the code better.

dicethedev:
- Replaces raw `0..4` interval dispatch in blockchain tick handling with a typed `SlotInterval` enum.
- This makes interval duties self-documenting and lets Rust enforce exhaustive handling when intervals change.
- Closes the gap where interval-specific behavior was matched using integer literals.

## What Changed
- `crates/blockchain/src/lib.rs`
  - Added `SlotInterval` with `from_slot_index` as the integer slot-to-enum constructor.
  - Updated `BlockChainServer::on_tick` to compare and match on `SlotInterval` variants.
  - Kept interval logging as numeric output through `Display`.

- `crates/blockchain/src/store.rs`
  - Reused `SlotInterval` for the store-level tick interval match.

- No behavior change intended.
- The same interval indices map to the same duties:
  - `0` → block publication
  - `1` → attestation production
  - `2` → aggregation
  - `3` → safe target update
  - `4` → end of slot
- Interval matches are now exhaustive, so adding/removing interval variants requires updating all dispatch sites at compile time.

## Tests Added / Run
- No tests added; this is a refactor with no intended behavior change.
- Ran:
  - `cargo fmt`
  - `cargo check -p ethlambda-blockchain`

## Related Issues / PRs
- Closes #469
- Related to #

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -1,5 +1,4 @@
 use std::collections::{HashMap, HashSet, VecDeque};
-use std::fmt;
 use std::time::{Duration, Instant, SystemTime};
 
 use ethlambda_network_api::{BlockChainToP2PRef, InitP2P};
@@ -59,44 +58,6 @@ pub use ethlambda_types::block::MAX_ATTESTATIONS_DATA;
 ///
 /// See: leanSpec PR #682.
 pub const GOSSIP_DISPARITY_INTERVALS: u64 = 1;
-
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
-pub(crate) enum SlotInterval {
-    BlockPublication,
-    AttestationProduction,
-    Aggregation,
-    SafeTargetUpdate,
-    EndOfSlot,
-}
-
-impl SlotInterval {
-    pub(crate) fn from_slot_index(index: u64) -> Self {
-        match index {
-            0 => Self::BlockPublication,
-            1 => Self::AttestationProduction,
-            2 => Self::Aggregation,
-            3 => Self::SafeTargetUpdate,
-            4 => Self::EndOfSlot,
-            _ => unreachable!("slots only have 5 intervals"),
-        }
-    }
-
-    fn as_slot_index(self) -> u64 {
-        match self {
-            Self::BlockPublication => 0,
-            Self::AttestationProduction => 1,
-            Self::Aggregation => 2,
-            Self::SafeTargetUpdate => 3,
-            Self::EndOfSlot => 4,
-        }
-    }
-}
-
-impl fmt::Display for SlotInterval {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        self.as_slot_index().fmt(f)
-    }
-}
 
 /// Milliseconds until the next interval boundary, measured relative to genesis.
 fn ms_until_next_interval(now_ms: u64, genesis_time_ms: u64) -> u64 {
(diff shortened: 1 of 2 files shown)


crates/blockchain/src/lib.rs
```diff
@@ -59,6 +60,44 @@ pub use ethlambda_types::block::MAX_ATTESTATIONS_DATA;
...
+            Self::BlockPublication => 0,
+            Self::AttestationProduction => 1,
+            Self::Aggregation => 2,
+            Self::SafeTargetUpdate => 3,
+            Self::EndOfSlot => 4,
+        }
+    }
+}
+
+impl fmt::Display for SlotInterval {
```
MegaRedHand: If this is unused, we should just remove it

crates/blockchain/src/lib.rs
```diff
@@ -59,6 +60,44 @@ pub use ethlambda_types::block::MAX_ATTESTATIONS_DATA;
...
+pub(crate) enum SlotInterval {
+    BlockPublication,
+    AttestationProduction,
+    Aggregation,
+    SafeTargetUpdate,
+    EndOfSlot,
+}
+
+impl SlotInterval {
+    pub(crate) fn from_slot_index(index: u64) -> Self {
```
MegaRedHand: If possible, we should give only `current time in ms -> SlotInterval` and `current time in intervals -> SlotInterval` constructors, so that callers don't have to compute the interval themselves. That way we centralize the logic in this enum

crates/blockchain/src/lib.rs
```diff
@@ -192,7 +231,9 @@ impl BlockChainServer {
         // Calculate current slot and interval from milliseconds
         let time_since_genesis_ms = timestamp_ms.saturating_sub(genesis_time_ms);
         let slot = time_since_genesis_ms / MILLISECONDS_PER_SLOT;
-        let interval = (time_since_genesis_ms % MILLISECONDS_PER_SLOT) / MILLISECONDS_PER_INTERVAL;
+        let interval = SlotInterval::from_slot_index(
+            (time_since_genesis_ms % MILLISECONDS_PER_SLOT) / MILLISECONDS_PER_INTERVAL,
+        );
```
MegaRedHand: Here, for example, we could do:

```suggestion
        let interval = SlotInterval::from_ms_since_genesis(time_since_genesis_ms);
```

crates/blockchain/src/store.rs
```diff
@@ -278,7 +278,7 @@ pub fn on_tick(store: &mut Store, timestamp_ms: u64, has_proposal: bool) {
             .expect("set_time should succeed");
 
         let slot = store.time() / INTERVALS_PER_SLOT;
-        let interval = store.time() % INTERVALS_PER_SLOT;
+        let interval = SlotInterval::from_slot_index(store.time() % INTERVALS_PER_SLOT);
```
MegaRedHand: And here:

```suggestion
        let interval = SlotInterval::from_intervals_since_genesis(store.time());
```

MegaRedHand (commented):
Left some comments. The direction is good, but there are some improvements I think are necessary before we merge this

dicethedev:
> Left some comments. The direction is good, but there are some improvements I think are necessary before we merge this

Alright Mega. I will push a fix a few minutes with your suggestions.

MegaRedHand (approved):
Good stuff!

Issue #465, opened 2026-06-24
Block building optimizations


MegaRedHand:
Block building can take anywhere from 0.5s to 3s. We should look for ways to reduce this, to publish blocks in time.

dicethedev:
I will also like to look into this

MegaRedHand:
@dicethedev I'm currently working on these, since they're closely related to problems we see in our internal devnets. If you come up with any cool optimizations to try, however, feel free to open an issue or a PR so we can try them!

PR #497, opened 2026-07-03
feat(cli): add offline block-building benchmark subcommand

pablodeymo:

Adds `ethlambda benchmark synthetic` — an offline harness that measures block building **exactly as executed when the node proposes**, against a reproducible synthetic workload, with no devnet required.

"Optimize block building" (#465) is the top roadmap item, but today the only observability is Prometheus histograms on a live devnet: noisy, not reproducible, and unable to compare an optimization against a baseline. This PR is milestone M1 (mock-crypto mode) of the design in `docs/plans/block-building-benchmark.md`; M2 adds real XMSS/leanVM pools and post-build seal-phase measurement, M3 adds replay-from-datadir.

```
$ make bench
  ...
  compact               10    0.000ms    0.000ms    0.000ms    0.000ms    0.000ms
  select_payloads       10    0.001ms    0.001ms    0.001ms    0.001ms    0.001ms
  stf_simulate          10    0.009ms    0.009ms    0.009ms    0.009ms    0.009ms
  overhead              10    0.058ms    0.060ms    0.059ms    0.063ms    0.064ms
  wall                  10    0.068ms    0.070ms    0.069ms    0.073ms    0.073ms
```

## What Changed

| File | Change |
|------|--------|
| `bin/ethlambda/src/benchmark/{mod,corpus,report}.rs` | New harness: seeded synthetic chain over `InMemoryBackend`, per-slot pool seeding into the pending pool (promoted by the proposal tick, as in production), iteration loop driving `produce_block_with_signatures` + block import; stats + human/JSON report |
| `bin/ethlambda/src/cli.rs` | Optional subcommand via `subcommand_negates_reqs` + `args_conflicts_with_subcommands`; the 7 node-required args become `Option<T>` + `required = true` (clap's missing-arg errors preserved); unit tests pin flat-invocation compat |
| `bin/ethlambda/src/main.rs` | `main` is now synchronous: it parses, sets up logging/metrics and dispatches the benchmark (logs to stderr so JSON on stdout stays pipe-clean) without ever starting the tokio runtime; the node path moved into `run_node`, which carries the `#[tokio::main]` attributes and unwraps the required args once |
| `crates/blockchain/src/block_builder.rs` | **fix (own commit):** `extend_proofs_greedily` broke equal-coverage proof ties by randomized `HashSet` iteration order — block aggregation bits differed run-to-run on a live node; ties now break to the lowest pool index |
| `crates/storage/{lib,store}.rs` | Export `NEW_PAYLOAD_CAP` so the harness rejects `--proofs-per-data` batches the pending pool would evict whole |
| `bin/ethlambda/build.rs` | Embed the resolved leanSig and leanVM revisions from `Cargo.lock` into reports (leansig tracks the moving `devnet4` branch; leanVM does the aggregation, so its pinned rev moves the measured crypto too). The per-`[[package]]` parse collects `name` and `source` before extracting the rev, so it does not depend on TOML field order |
| `Makefile`, `.github/workflows/ci.yml` | `make bench`; seconds-fast mock smoke step in the Test job validating the JSON contract |
| `docs/plans/block-building-benchmark.md` | Design doc and milestone roadmap |

- **Every existing flat node invocation parses unchanged** (lean-quickstart, Dockerfile, devnet skills) — pinned by `cli.rs` unit tests covering parsing, missing-arg errors, and mixed-invocation rejection.
- **Node runtime behavior is unchanged except the tie-break fix**, which only affects cases that were previously *random*.
- **Determinism:** same seed + params → identical per-iteration block roots (recorded in the JSON as a checksum, so a baseline-vs-optimized diff proves an optimization changed only speed, not attestation selection). The harness never reads the wall clock into results.
- **Exact phase attribution:** per-iteration `select_payloads`/`compact`/`stf_simulate` come from sample-sum deltas of the existing phase histogram (sums accumulate raw f64 seconds; each phase observes once per build); a count-delta assertion turns any accounting drift into a hard error. Skipped/unattributed time is reported as `overhead`, never hidden.
- Guarded inputs fail fast instead of producing bogus data: `--proofs-per-data` beyond the pool cap, warmup+iterations overflow, post-seed eviction.

## Tests Added / Run

- New unit tests: CLI compat (6), corpus partitioning/determinism (2), report stats/percentiles (4).
- `make fmt`, `make lint` — clean; `cargo test --workspace --release` — all passing (spec tests included).
- Determinism verified: 3 same-seed runs produce identical block-root sequences; a different seed changes them.
- CI smoke verified locally: `... --format json | jq -e '.schema_version == 1 and (.samples | length == 3)'`.

## Related Issues / PRs

- Related to #465 (Optimize block building) — this provides the measurement harness; later milestones (real crypto, replay) tracked in `docs/plans/block-building-benchmark.md`.

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


bin/ethlambda/build.rs
```diff
@@ -12,5 +14,45 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     Ok(())
 }
+
+/// Embed the resolved leansig git revision from the workspace Cargo.lock.
+///
+/// leansig is pinned to a moving branch, so a `cargo update` changes the
+/// measured crypto with zero ethlambda diff; benchmark reports embed this
+/// revision to keep results interpretable across lock bumps.
+fn emit_leansig_rev() {
+    let rev = leansig_rev_from_lockfile().unwrap_or_else(|| "unknown".to_string());
```
MegaRedHand, line 28: We should add leanVM to this too

bin/ethlambda/src/main.rs
```diff
@@ -117,26 +138,27 @@ async fn main() -> eyre::Result<()> {
...
+    // clap enforces the node-required arguments when no sub-command is given
+    // (`subcommand_negates_reqs` only lifts them for sub-commands, which
+    // returned above), so these unwraps cannot fail on the node path.
+    let config_path = require_arg(options.genesis, "--genesis")?;
+    let validators_path = require_arg(options.validators, "--validators")?;
+    let bootnodes_path = require_arg(options.bootnodes, "--bootnodes")?;
+    let validator_config = require_arg(options.validator_config, "--validator-config")?;
+    let validator_keys_dir = require_arg(options.hash_sig_keys_dir, "--hash-sig-keys-dir")?;
+    let node_key_path = require_arg(options.node_key, "--node-key")?;
+    let node_id = require_arg(options.node_id, "--node-id")?;
```
MegaRedHand, line 153: Why do we need this change?

bin/ethlambda/build.rs
pablodeymo: Done in 9f7a311. Reports now carry both revisions, resolved from `Cargo.lock` — `leansig` for leanSig and `lean-multisig` for leanVM (the direct dependency `ethlambda-crypto` builds against; every other leanVM crate resolves to the same rev).

While in there I also made the parse order-independent per greptile's note: `name` and `source` are collected per `[[package]]` block before the rev is extracted, so a lockfile that emitted `source` first would no longer silently yield `unknown`.

Sample line from a run: `leansig=15cbdd43ec85… leanvm=e2592df4e30f…`.

bin/ethlambda/src/main.rs
pablodeymo: Because the seven node args have to be `Option<T>` for the sub-command to parse at all. `benchmark` passes none of `--genesis/--validators/…`, and `subcommand_negates_reqs` only switches off clap's *requirement check* — the derive still extracts every field, and extraction of a non-`Option<PathBuf>` with no value on the command line fails.

So they are `Option<T>` + `#[arg(required = true)]`: clap keeps emitting its own missing-argument errors for the flat node invocation (pinned by the `cli.rs` tests), and `require_arg` on the node path is just the unwrap of an invariant clap already enforced.

The alternative that keeps them non-optional is moving the benchmark into its own bin target, which means splitting the crate into a lib + two bins so the `benchmark`/`cli` modules can be shared — happy to do that instead if you would rather not have the `Option` churn.

Note this hunk moved in 20fab1b: `main` is now synchronous (parse, logging, metrics, benchmark dispatch) and the node path lives in `run_node`, so the benchmark no longer parks a tokio worker thread.

pablodeymo:
Superseded by #593, which carries the same harness split into reviewable pieces:

- **#590** (merged) — the `extend_proofs_greedily` non-determinism this harness uncovered, as a standalone node fix.
- **#591** — `node`/`benchmark` token dispatch in a new `command.rs`. This drops the `subcommand_negates_reqs` approach reviewed here: `cli.rs` keeps its exact shape, the seven node-required arguments stay non-`Option`, clap keeps emitting its own missing-argument errors, and the `require_arg` helper is gone. That answers the question on `main.rs` about why those fields had to change — they no longer do.
- **#593** — the benchmark itself, stacked on #591.

Closing in favour of that stack; the review comments here are addressed there.

PR #523, opened 2026-07-15
fix(metrics): discard block-processing time for duplicate blocks

MegaRedHand:
## Problem

`lean_fork_choice_block_processing_time_seconds` is sampled by a `TimingGuard` created at the top of `on_block_core`, **before** the idempotent duplicate-block check returns early. A duplicate block does no processing, so it recorded a near-zero sample that skews the histogram low. Duplicates are common during range sync, so the effect is not negligible.

This is the same class of issue as the recently fixed tick-interval metric (#514): the sample boundary didn't line up with the unit of work — there it inflated the tail, here it deflates the percentiles.

## Change

- Add `TimingGuard::discard()` to cancel a measurement before the guard drops (a `disarmed` flag checked in `Drop`).
- Call it on the duplicate-block early return in `on_block_core`.

The block hashing (`hash_tree_root`) stays inside the timed span since it is real work; only the sample for a no-op duplicate is dropped. All other early returns (validation rejections) still record, as they represent real processing time.

## Verification

`make fmt`, `make lint` (clean), full workspace build. Two new unit tests in `timing.rs` cover the guard: `records_a_sample_on_drop` and `discard_suppresses_the_sample` (both pass). `cargo test -p ethlambda-blockchain` 46 unit tests pass. Fork-choice/signature spec tests require downloaded fixtures (absent in this env).

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -576,11 +576,7 @@ fn on_block_core(
     let block_root = block.hash_tree_root();
     let slot = block.slot;
 
-    // Skip duplicate blocks (idempotent operation). The block hashing above is
-    // real work and stays inside the timed span, but a duplicate does no block
-    // processing, so discard the sample instead of recording a near-zero
-    // duration that would skew `lean_fork_choice_block_processing_time_seconds`
-    // low (duplicates are common during range sync).
+    // Skip duplicate blocks (idempotent operation)
     if store
         .has_state(&block_root)
         .expect("DB read should succeed")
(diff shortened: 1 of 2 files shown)


crates/blockchain/src/store.rs
```diff
@@ -569,18 +569,23 @@ fn on_block_core(
...
     let block = &signed_block.message;
     let block_root = block.hash_tree_root();
     let slot = block.slot;
 
-    // Skip duplicate blocks (idempotent operation)
+    // Skip duplicate blocks (idempotent operation). The block hashing above is
+    // real work and stays inside the timed span, but a duplicate does no block
+    // processing, so discard the sample instead of recording a near-zero
+    // duration that would skew `lean_fork_choice_block_processing_time_seconds`
+    // low (duplicates are common during range sync).
```
```suggestion
    // Skip duplicate blocks (idempotent operation)
```

pablodeymo (approved):

Issue #156, opened 2026-02-26
Add guest program and ZK proving of the STF


MegaRedHand:
The Lean chain endgame includes ZK proving of the STF for quick block validity checking. We should start preparing for it, following in Zeam's footsteps.

Sahilgill24:
hey @MegaRedHand, could you please assign the issue :)

MegaRedHand:
Sure 😄

PR #531, opened 2026-07-21
feat(zkvm): state transition function proving using zkvm's  

Sahilgill24:
## Description / Motivation

#156 

## What Changed

The approach is based on this doc [ethlambda's stf proving using zkvm's](https://hackmd.io/@Ayhm2-FHQhSLkoc1OEcr9g/BkaRNBuEzx). 

### Current Progress
- added the StfInput type, which shall be used by the guest-program 
- scaffolded the sp1 guest program inside a new guest-program crate 
- prover core and prover backend for sp1 with a mock prover is complete
- Tests for sp1 prover for the execute path have been completed

## Verification Checklist

- [x] Ran `make fmt` — clean
-  [ ] Remove the TODO's


Sahilgill24:
> [!NOTE]
> I will be updating and adding proper comments where ever necessary and also try to write a more idiomatic approach for the StfInput if possible, the current version is a basic scaffold as mentioned in the 1st step of the implementation plan in the doc. 
> Also will add more steps to the PR heading as well once sp1 integration is complete, currently on kept the progress status

Cargo.toml
```diff
@@ -9,7 +9,7 @@ members = [
     "crates/common/crypto",
     "crates/common/metrics",
     "crates/common/test-fixtures",
-    "crates/common/types",
+    "crates/common/types", "crates/guest-program/sp1",
```
pablodeymo:
```suggestion
    "crates/common/types",
    "crates/guest-program/sp1",
```

crates/guest-program/sp1/src/main.rs
```diff
@@ -0,0 +1,13 @@
...
+sp1_zkvm::entrypoint!(main);
+
+use ethlambda_state_transition::state_transition;
+use ethlambda_types::{primitives::HashTreeRoot, stf::StfInput};
+
+fn main() {
+    let input: StfInput = sp1_zkvm::io::read();
+    let mut state = StfInput::return_state(input.clone());
+    let block = &StfInput::return_block(input.clone());
+    let _ = state_transition(&mut state, block);
```
MegaRedHand: This is ignoring the error, right? This should panic or return an error

crates/common/types/src/stf.rs
```diff
@@ -0,0 +1,28 @@
...
+        }
+    }
+
+    pub fn return_state(input: Self) -> State {
+        State::from_ssz_bytes(&input.state).expect("error decoding State")
+    }
+
+    pub fn return_block(input: Self) -> Block {
+        Block::from_ssz_bytes(&input.block).expect("error decoding block")
+    }
```
MegaRedHand: These can be made references:


```suggestion
    pub fn return_state(&self) -> State {
        State::from_ssz_bytes(&self.state).expect("error decoding State")
    }

    pub fn return_block(&self) -> Block {
        Block::from_ssz_bytes(&self.block).expect("error decoding block")
    }
```

crates/guest-program/sp1/src/main.rs
```diff
@@ -0,0 +1,13 @@
+#![no_main]
+sp1_zkvm::entrypoint!(main);
+
+use ethlambda_state_transition::state_transition;
+use ethlambda_types::{primitives::HashTreeRoot, stf::StfInput};
+
+fn main() {
+    let input: StfInput = sp1_zkvm::io::read();
+    let mut state = StfInput::return_state(input.clone());
+    let block = &StfInput::return_block(input.clone());
```
MegaRedHand: Then here:

```suggestion
    let mut state = input.return_state();
    let block = &input.return_block();
```

Cargo.toml
Sahilgill24: Will do !!

crates/common/types/src/stf.rs
Sahilgill24: I was about to make these changes in the next step :)

Cargo.toml
Sahilgill24: also a suggestion, maybe in the CI a check for `taplo` tool could be added, which checks the toml file formatting using taplo fmt --all.

crates/guest-program/sp1/src/main.rs
Sahilgill24: yup, will return the error

Sahilgill24:
Hey @MegaRedHand, for the [commit#d62d221](https://github.com/lambdaclass/ethlambda/pull/531/changes/d62d2216b0d238007443f415a0067df9ded29dd1), 

the zkVM target is specified for the guest program and it's dependencies, I have currently set the target inside the code in the types and metrics crate, but there might be other ways to do so like creating a different no-op module and setting it to the zkVM target. Could you have a look once and confirm which approach would better set with ethlambda's code design format.

MegaRedHand:
Having a no-op module for metrics sounds better than having `#[cfg(not(target_os = "zkvm"))]` everywhere. But for `crates/common/types/src/signature.rs`, I think we should move the signatures logic to `crypto` instead of `types` (feel free to refactor this in another PR).

Sahilgill24:
Thanks, I will refactor the metrics and make a no-op metrics module as well. 
I will open the refactor PR once this one is complete :)

Sahilgill24:
>  I will open the refactor PR once this one is complete :)

I think I would need to make this refactor first, so while testing out the prover, the prover has dependency conflicts with lean-sig (the prover runs naitvely, where as the guest runs for the zkvm target that's why guest had no such issues) , there are some temporary solutions, but I think as anyways a refactor needs to be done, I will complete that first :) 

The changes were quite trivial for this signature logic port, made them here #541 .

Sahilgill24:
Hey @MegaRedHand , Similiar to zeam, I have added a mock chain generator and a binary for it's proving/verifying and execution.  Could you have a look at it once and mention if any more changes needed to be done to the Mock chain. 

- `prover/core/src/mock_chain.rs` is the Mock chain generator
- `prover/sp1/src/bin/prove.rs` is the binary to run it. 

you can run the binary as follows 

```zsh
  cargo run --bin prove -- --mode execute --blocks 5  # execution only
  cargo run --bin prove -- --mode prove   --blocks 5  # proving + verification
  cargo run --bin prove -- --blocks 5                 # default = prove
```

Cargo.toml
```diff
@@ -13,9 +13,19 @@ members = [
     "crates/net/api",
     "crates/net/p2p",
     "crates/net/rpc",
+    "crates/prover/core",
```
MegaRedHand, line 16: This is missing the SP1 prover

MegaRedHand:
@Sahilgill24 Please add instructions on how to run the mock chain generator to the docs, including any required dependencies.

Regarding the mock chain, I think it's a good first step, but something similar to [ethrex-replay](https://github.com/lambdaclass/ethrex-replay), where we can prove real blocks, would be ideal. Then, we'll be able to prove blocks from a running devnet easily.

Sahilgill24:
Hey @MegaRedHand 
I had missed the instructions but you would need to have the `cargo prove` toolchain on your local from the [sp1 docs](https://docs.succinct.xyz/docs/sp1/getting-started/quickstart)

or directly use this 
```
# sp1 installation
curl -L https://sp1up.succinct.xyz | bash
sp1up
cargo prove --version
```
now it should work on your local directly with the commands in the previous comment. 


> Regarding the mock chain, I think it's a good first step, but something similar to [ethrex-replay](https://github.com/lambdaclass/ethrex-replay), where we can prove real blocks, would be ideal. Then, we'll be able to prove blocks from a running devnet easily.

Oh, I was unaware about ethrex-replay, would have a look at it and make the relevant changes.

Cargo.toml
Sahilgill24: the sp1 prover, has a separate toolchain and is excluded from regular builds `exclude = ["crates/guest-program/sp1", "crates/prover/sp1"]` because otherwise it would force other commands like cargo build to install the sp1 toolchain as well.

Sahilgill24:
Hey @MegaRedHand , I have pushed the replay crate along with some changes related to it, most of these changes are in the guest-program + prover crate, but one change was in the net crate for adding SSZ byte fetching from RPC as the current JSON fetching was not compatible as State/Block are not deserializable to be put as `StfPublicValues`. 

Now for the testing part, could you once run a local devnet and on the finalization of slot 3, try the following commands
This would require the sp1 toolchain and you can install that from the above comment's instructions. 
```
cd tooling/replay
cargo run --features sp1 -- --rpc-url http://127.0.0.1:<PORT> --action execute   block 3
cargo run --features sp1 -- --rpc-url http://127.0.0.1:<PORT> --action prove   block 3
cargo run --features sp1 -- --rpc-url http://127.0.0.1:<PORT> --action prove blocks 2 6
```

MegaRedHand:
@Sahilgill24 Ran it. The tool looks good 👍 
The output is missing the time it took to execute/prove a block, which is one of the best features of ethrex-replay.

Sahilgill24:
> The output is missing the time it took to execute/prove a block, which is one of the best features of ethrex-replay.

my first thought was that adding a timer would not be very useful as the current MockProver generates proofs quite quick, the actual prover could take significant time, but now as you mentioned it, I think it can be added to make it better :) 

also any particular reference to follow for outputting the logs

MegaRedHand:
I think having the same output as ethrex-replay is good

PR #539, opened 2026-07-24
feat(crypto)!: track leanVM main (internalized XMSS + reworked aggregation)

MegaRedHand:
## What

Moves the leanVM dependency off the `devnet4`-era `lean-multisig` + `leansig_wrapper` crates and onto current `main`, pinned at **`a5909d1`**.

leanVM's `main` **removed** the crates ethlambda imported, **internalized XMSS** in its own `xmss` crate (dropping the external leanSig dependency), and **reworked** the recursive-aggregation API. So this is a migration of the whole signature stack, not a version bump: `cargo update` fails outright (`no matching package named leansig_wrapper`).

It also takes the prover off leanVM's bump arena by default, which is what stops a node's RSS from ratcheting upward for its whole lifetime.

## Changes

| Area | Before (`rev e2592df`) | After (`rev a5909d1`) |
|------|------------------------|-----------------------|
| Deps | `lean-multisig`, `leansig_wrapper`, `leansig` (devnet4) | `xmss` + `lean-multisig` (leanVM's facade), `ssz` (ethereum_ssz), `postcard` |
| Key/sig types | leanSig devnet4 | leanVM's own `xmss` crate |
| Proof wire form | `compress_without_pubkeys()` | `to_bytes_without_pubkeys()` — **unchanged shape**: pubkeys stay off the wire |
| Pubkey size | 52 bytes | 32 bytes (`PUB_KEY_SSZ_LEN`) |
| Signature size | 2536 bytes | 1208 bytes (`SIGNATURE_SSZ_LEN`) |
| Prover allocator | leanVM arena | system allocator, arena behind `--prover-arena` |
| Backend init | lazy, on first proof | explicit `init_leanvm(use_arena)` at startup |
| Secret-key window | sliding two-tree preparation window | fixed activation range (`advance_preparation` is now a no-op) |
| Proof verification | pool-dispatched | single-threaded |

- **`types::signature`**: rebuilt `ValidatorPublicKey`/`ValidatorSignature`/`ValidatorSecretKey` on `xmss`. Pubkeys/sigs (de)serialize via SSZ; secret keys via postcard. `SIGNATURE_SIZE` + `PUBLIC_KEY_SIZE` are static-asserted against the scheme constants, so a leanVM bump that changes them breaks the build instead of silently producing wrong-length keys.
- **`crypto`**: rewired aggregation/verification against leanVM's facade. The **wire format and every public signature are kept as on `main`**: proofs serialize without their participant pubkeys, and the caller attaches the resolved signer set at decode (`from_bytes_without_pubkeys`). A set other than the one aggregated attaches fine but fails verification, since the proof binds a hash of the set — covered by a new `test_verify_wrong_pubkey_set_fails`.
- **Validator pubkeys 52 → 32 bytes**: genesis/state layouts, deser messages, and test fixtures updated; pinned genesis state/block roots recomputed.
- `VerificationError::ProofError(#[from] ProofError)` becomes `VerificationFailed(String)`: leanVM's `ProofError` lives in its `backend` crate, which the facade imports but does not re-export, so it is not nameable here.

## Memory: off the arena by default

leanVM's `setup_prover` engages a bump arena that never returns pages to the OS — `free` is a no-op, and a phase reset only rewinds the per-thread bump pointers (`zk-alloc`: no `MADV_DONTNEED`, no `munmap`, and on Linux `enable_arena` also disables heap trimming and large-allocation mmap process-wide). RSS therefore climbs to the process's allocation high-water mark and stays there, which is the unbounded growth seen on long-lived devnet nodes.

`a5909d1` adds `setup_prover_without_arena`, which runs the same prover on the system allocator: slower, bounded, identical proofs. That is now the default, with `--prover-arena` to opt back in on hosts with memory to spare where proving latency is the constraint.

## Prover setup and the proving permit

Setup used to be lazy and reached through the permit helper, which coupled two unrelated things: taking the prove lock also triggered backend init, so the decoding paths had to hold the lock across work that needs no lock, purely to get the aggregation bytecode compiled first.

- `init_leanvm(use_arena)` does both halves in one place; the binary calls it straight after parsing arguments, ahead of the Hive test-driver branch (which verifies signatures).
- `acquire_prover()` is just the permit again, so every entry point takes it immediately before the prove call rather than across argument conversion and proof decoding.
- `ensure_verifier_ready` is gone — nothing needs a lazy hook anymore.

This moves **~2.3s** of bytecode compilation (measured, 3 runs, both allocators) from the first proof or verification into startup. Nodes paid it either way; it is now off the first duty's critical path, at the cost of a slower boot for nodes that never prove.

## ⚠️ Breaking / follow-ups

The wire format of the proof *contents* and the genesis key format change, even though the pubkey-less framing matches `main`:

- Interop-incompatible with clients still on the previous scheme.
- Requires **regenerated genesis keys** (32-byte xmss pubkeys, postcard secret keys).
- Fixture-driven spec tests need **regenerated fixtures**; the currently released ones are the old format and fail while deserializing, before reaching any consensus or crypto code:

| Harness | Result | Error |
|---|---|---|
| `forkchoice_spectests` | 122 failed | `ValidatorPubkey length != 32` |
| `stf_spectests` | 73 failed, 1 passed | same |
| `ssz_spectests` | 7 failed, 112 passed | same, plus `expected 1208, got 2536` |
| `signature_spectests` | 3 failed | same |

Everything else in the workspace passes, including all 11 `ethlambda-crypto` tests (the slow ignored ones plus a new integration test covering the arena path).

Landing this needs ecosystem coordination; it is not a drop-in for the live devnet.


MegaRedHand:
Superseded by the new leanVM version using binary fields: #606

PR #540, opened 2026-07-27
fix(crypto): use leanVM's `setup_prover` and prove one at a time

TomWambsgans:
- `ensure_prover_ready` only called `init_aggregation_bytecode`, one of four steps in `setup_prover`, so leanVM's arena was never engaged and every prover buffer used the system allocator.
- With the arena on, only one proof may run at a time, so a permit guards each proving entry point.
- leanVM pin bumped to 380da82c.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,9 +88,10 @@ dependencies = [
 [[package]]
 name = "air"
 version = "0.1.0"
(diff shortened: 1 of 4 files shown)


MegaRedHand (approved):

MegaRedHand:
The repo config doesn't allow merging unsigned commits, so I manually cherry-picked it into the base PR: [db7bc13](https://github.com/lambdaclass/ethlambda/pull/539/commits/db7bc133f5e66a45bcd7ce065152fdbfe43b136b)

Thanks again for the help!

PR #541, opened 2026-07-28
refactor: moving signature logic from ethlambda-types to ethlambda-crypto

Sahilgill24:
## Description / Motivation

porting the lean-sig based signature logic from types crate to crypto crate as mentioned [here](https://github.com/lambdaclass/ethlambda/pull/531#issuecomment-5093478330)

It is first of all a cleaner and more idiomatic approach and secondly ethlambda-types crate builds/compiles for the prover crate(zkVM prover), which also leads to dependency conflicts between leansig dependencies and the zkVM sdk dependencies being used. This port removes all leansig dependencies from the ethlambda-types

## What Changed

- added a new `signature.rs` file to the crypto crate along with a extension trait `ValidatorPublicKeys` with the same methods as before, it helps in avoiding cyclic dependencies and the call sites remain the same. 
- removed the older `signature.rs` from the types crate and moved the `SIGNATURE_SIZE` directly to the `attestation.rs` 

## Verification Checklist

- [x] Ran `make fmt`
- [x] Ran `make lint` (clippy with `-D warnings`) 


crates/common/crypto/src/signature.rs
```diff
@@ -125,6 +122,26 @@ impl ValidatorSecretKey {
...
+
+impl ValidatorPubkeys for Validator {
+    fn get_attestation_pubkey(&self) -> Result<ValidatorPublicKey, SignatureParseError> {
+        ValidatorPublicKey::from_bytes(&self.attestation_pubkey)
+    }
+
+    fn get_proposal_pubkey(&self) -> Result<ValidatorPublicKey, SignatureParseError> {
+        ValidatorPublicKey::from_bytes(&self.proposal_pubkey)
+    }
+}
```
MegaRedHand: We should inline these instead. Having a new trait for this doesn't seem like a good trade

MegaRedHand (commented):
Looks good. Left a suggestion

crates/common/crypto/src/signature.rs
Sahilgill24: I added it just so the call sites remain unchanged, if we inline them we would have to replace them with `ValidatorPublicKey::from_bytes()` instead of `get_attestation_key` respectively. 
Will inline them instead :)

MegaRedHand (approved):
Great work! Thanks for your contribution!

PR #547, opened 2026-07-30
test(spec): assert the rejection reason on expected-failure fixtures

MegaRedHand:
## Summary

Negative leanSpec fixtures name **why** their input must be rejected, in a
`rejectionReason` field. The spec-test runners parsed that field only so
`deny_unknown_fields` would accept it and asserted nothing more than "some error
came back", so a fixture could pass on a failure unrelated to the rule it
exercises.

This wires the reason into the assertion for all three offline runners
(fork choice, state transition, verify signatures).

## What changed

| Piece | Where |
|---|---|
| `RejectionReason`: typed mirror of leanSpec's `RejectionReason` StrEnum (36 reasons) | `crates/common/test-fixtures/src/rejection.rs` |
| `StoreError` → reason classifier, exhaustive so new variants must be classified | `spec_test_runner::rejection_reason` |
| `state_transition::Error` → reason, as a total `From` impl | `test-fixtures/src/rejection.rs` |
| `apply_fork_choice_step` now returns a typed `StepError` (store rejection vs harness failure) instead of `String` | `spec_test_runner` |
| Shared comparison + failure wording used by all three runners | `rejection::check_rejection_reason` |

Two failure modes are treated as test failures, not passes, since accepting
either restores "any error will do":

- the client error carries no spec counterpart (unclassified), and
- the fixture names a reason this build does not know.

Unknown reasons still **deserialize** (as `Unknown(_)`) so the Hive test driver
keeps answering a step over HTTP rather than 422-ing the request when leanSpec
adds a reason name; only the offline runners fail on them.

`StoreError::AggregateVerificationFailed` was raised both for an attestation
aggregate (`INVALID_SIGNATURE`) and for a block's merged multi-message proof
(`INVALID_BLOCK_PROOF`). Since the spec separates those, block-proof
verification now returns its own `BlockProofVerificationFailed` variant, which
also makes production logs say which proof failed.

## Results

- **fork choice: 122/122 pass** — all 35 negative steps already failed for
  exactly the reason their fixture names.
- **verify signatures: 3/3 pass.**
- **state transition: 72 pass, 2 fail, 2 skipped.**

### The 2 remaining failures are real divergences

They passed before *for the wrong reason*; the new assertion is what surfaces
them. Left failing rather than allowlisted, so the divergence stays visible:

| fixture | expects | ethlambda raises | diagnosis |
|---|---|---|---|
| `test_block_exceeding_distinct_attestation_data_cap_rejects_block` | `TOO_MANY_ATTESTATION_DATA` | `STATE_ROOT_MISMATCH` | the `MAX_ATTESTATIONS_DATA` cap is enforced on block import (`store.rs`), not in the STF, so replaying the STF alone misses it |
| `test_source_slot_beyond_tracked_range_rejects_block` | `JUSTIFIED_SLOT_OUT_OF_RANGE` | `STATE_ROOT_MISMATCH` | `is_valid_vote` treats a source past the tracked justification window as "not justified" and skips the vote, where the spec rejects the block |

The first is a layering gap rather than live exposure: a node still rejects the
capped block at import. Both want follow-up PRs.

### The 2 skips are unreplayable fixtures, not client bugs

`test_block_with_wrong_slot` and
`test_block_at_parent_slot_rejected_when_slot_processing_skipped` are authored
with leanSpec's `BlockSpec.skip_slot_processing` (`block_spec.py:71`), which
makes the filler call `process_block` alone. That flag never reaches the emitted
fixture — `StateTransitionFixture` carries only `pre`, `blocks`, `post`,
`postStateRoot`, `rejectionReason` — and the failing block is written with a
placeholder zero `stateRoot`. Replaying `state_transition()`, which is what the
fixture format's own description prescribes, therefore advances the slot first
and then dies on the state root instead of reaching the rule under test.

Nothing in the JSON marks the entry point, so **no client can reproduce these
two**. Verified against fresh clones: ream, zeam, gean, lantern and grandine all
"pass" them on the same state-root mismatch because they assert only that *some*
error occurred (grandine's negative path counts a state-root mismatch as the
expected failure outright). qlean's 2-month-old vendored copy has the identical
shape under the older `expectException: AssertionError` spelling, so this is not
a regression: sharpening the fixture's claim to a specific reason is what made
it checkable, and therefore visibly unsatisfiable.

Upstream fix would be to emit the entry point per block (`skipSlotProcessing`,
or `entryPoint: state_transition | process_block`), or to drop these two vectors
since they pin a Python-level API contract rather than cross-client behaviour.

## Testing

```
make fmt && make lint                       # clean
cargo test --workspace --profile release-fast
# stf_spectests: 72 passed / 2 failed (table above); all other suites pass
```

Three unit tests cover reason parsing, the unknown-reason fallback and JSON
deserialization.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2187,6 +2187,7 @@ dependencies = [
 name = "ethlambda-test-fixtures"
 version = "0.1.0"
 dependencies = [
+ "ethlambda-state-transition",
  "ethlambda-types",
  "hex",
  "libssz",
--- a/crates/blockchain/src/block_builder.rs
+++ b/crates/blockchain/src/block_builder.rs
@@ -668,10 +668,13 @@ fn compact_attestations(
                 let pubkeys = proof
                     .participant_indices()
                     .map(|vid| {
+                        let not_in_state = StoreError::ValidatorNotInState {
+                            validator_index: vid,
+                        };
                         let validator = head_state
                             .validators
                             .get(vid as usize)
-                            .ok_or(StoreError::InvalidValidatorIndex)?;
+                            .ok_or(not_in_state)?;
                         ValidatorPublicKey::from_bytes(&validator.attestation_pubkey)
                             .map_err(|_| StoreError::PubkeyDecodingFailed(vid))
                     })
--- a/crates/blockchain/src/spec_test_runner.rs
+++ b/crates/blockchain/src/spec_test_runner.rs
@@ -54,19 +54,14 @@ impl StepError {
 /// report as an unclassified rejection rather than accepting silently. The match
 /// is exhaustive so a new [`StoreError`] variant forces that decision here.
 ///
-/// Two variants are context-dependent and classified for the gossip path they
-/// are reached through in fixtures:
-///
-/// * `InvalidValidatorIndex` is raised both for a gossip attestation from an
-///   unregistered validator (`VALIDATOR_NOT_IN_STATE`) and for a block-level
-///   participant bounds check (`VALIDATOR_INDEX_OUT_OF_RANGE`, spec
-///   `signatures.py`). Only the former has fixtures today.
-/// * `StateTransitionFailed` defers to the state-transition classification,
-///   which the STF runner asserts directly.
+/// `StateTransitionFailed` is the one context-dependent variant: it defers to
+/// the state-transition classification, which the STF runner asserts directly.
 pub fn rejection_reason(err: &StoreError) -> Option<RejectionReason> {
     let reason = match err {
         StoreError::MissingParentState { .. } => RejectionReason::UnknownParentBlock,
-        StoreError::InvalidValidatorIndex => RejectionReason::ValidatorNotInState,
+        StoreError::ValidatorNotInState { .. } => RejectionReason::ValidatorNotInState,
+        StoreError::AttesterIndexOutOfRange { .. } => RejectionReason::ValidatorIndexOutOfRange,
+        StoreError::ProposerIndexOutOfRange { .. } => RejectionReason::ProposerIndexOutOfRange,
         StoreError::SignatureDecodingFailed | StoreError::SignatureVerificationFailed => {
             RejectionReason::InvalidSignature
         }
--- a/crates/blockchain/src/spec_test_runner.rs
+++ b/crates/blockchain/src/spec_test_runner.rs
@@ -4,7 +4,7 @@
 //! functions so fixture replay cannot drift between the two entry points.
 
 use ethlambda_storage::Store;
-use ethlambda_test_fixtures::fork_choice::ForkChoiceStep;
+use ethlambda_test_fixtures::{RejectionReason, fork_choice::ForkChoiceStep};
 use ethlambda_types::{
     attestation::{
         AggregationBits, HashedAttestationData, SignedAggregatedAttestation, SignedAttestation,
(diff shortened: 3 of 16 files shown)


pablodeymo (commented):
Reviewed the branch at `cb9bd7d` and reproduced the results locally in a worktree off the PR head.

## Verification

| Suite | Claimed | Measured |
|---|---|---|
| `forkchoice_spectests` | 122/122 | **122 passed** ✅ |
| `signature_spectests` | 3/3 | **3 passed** ✅ |
| `stf_spectests` | 72 pass / 2 fail / 2 skip | **72 passed, 2 failed**, both skips confirmed firing ✅ |
| `test-fixtures` unit tests | 3 | **3 passed** ✅ |

CI on the PR: `Lint` pass, `Verify License` pass, **`Test` fail** (the two STF fixtures).

The core design is good. Both classifiers (`spec_test_runner::rejection_reason` over `StoreError`, `From<&Error>` over the STF error) are wildcard-free, so a new error variant is a compile error until someone classifies it; `Unknown(_)` still deserializes so the Hive driver answers a step instead of 422-ing the request; and splitting `BlockProofVerificationFailed` out of `AggregateVerificationFailed` is a genuine fix that also makes production logs say which proof failed. Because negative fixtures ship a zero placeholder `stateRoot`, any unenforced rule surfaces as `STATE_ROOT_MISMATCH` — that is what makes the new assertion sharp.

## Findings

**1. CI is red by design, and the two known-bad classes are handled inconsistently.** Two STF fixtures fail; two others are silently skipped. Both groups are "known non-conformance", handled two different ways, and the red one makes `make test` unusable as a gate. Suggest a single mechanism: an expected-divergence table (fixture → reason the fixture names, reason we actually produce) that fails if the mismatch *changes or disappears*. That keeps the divergence visible, stops it rotting, and self-clears when it's fixed. Plus tracking issues for both.

**2. `InvalidValidatorIndex` is knowingly misclassified for the block path** (`crates/blockchain/src/spec_test_runner.rs:69`). It maps unconditionally to `VALIDATOR_NOT_IN_STATE`, but `verify_block_signatures` raises the same variant for out-of-range attestation participants (`store.rs:1122`, `store.rs:1147` → spec `VALIDATOR_INDEX_OUT_OF_RANGE`) and for an out-of-range `proposer_index` (`store.rs:1127`, `store.rs:1160` → `PROPOSER_INDEX_OUT_OF_RANGE`). Consequence: both of those reasons are unreachable from the store classifier, so the moment leanSpec adds a `verify_signatures` fixture for either, it fails for the wrong reason. The doc comment acknowledges the ambiguity, but the fix is the same shape this PR already applies to the block proof — split the variant. Worth doing here, since "one error ⇒ one reason" is the PR's thesis.

**3. A harness error can satisfy an expected rejection.** In `crates/blockchain/tests/forkchoice_spectests.rs`, the arm `(Err(_), false) if expected_reason.is_none() => Ok(())` accepts a `StepError::Harness` — e.g. `"unknown step type: X"` on a `valid: false` step passes. That contradicts `StepError`'s documented invariant that a harness failure "never satisfies an expected `rejectionReason`". I checked the pinned fixtures: all 34 invalid fork-choice steps carry a reason, so the arm is unreachable today — either drop it (strictly stronger) or match `Harness` first and always fail.

### Nits

- `crates/common/test-fixtures/src/verify_signatures.rs:50` keeps `alias = "expectException"` but now parses it as a `RejectionReason`. A legacy value like `"AssertionError"` becomes `Unknown(_)` and fails with *"add it to `RejectionReason` and classify the error that must produce it"* — misleading advice for a field that never carried a spec reason. Either drop the alias or keep the old field as presence-only.
- Commit 3 inlined the macro, so `as_str` and `From<&str>` are now two parallel 37-arm tables. Drift fails loudly (`Unknown`) rather than silently, but a `const ALL: &[RejectionReason]` round-trip test pins it cheaply; the current tests cover one variant per direction.
- `crates/common/test-fixtures/src/state_transition.rs:22` still spells the field `expectException` as `Option<String>`. That's fine — the Hive driver only checks `is_some()` — but it leaves three spellings in-tree; one line of doc would settle why.
- `SKIP_TESTS` matches by substring with no upstream link. Linking the leanSpec issue would give the skip an expiry path. Alternative worth considering instead of skipping: when a negative fixture's `state_transition()` yields `STATE_ROOT_MISMATCH` against a different expected reason, retry with `process_block` alone. `test_block_at_parent_slot_rejected_when_slot_processing_skipped` carries two blocks at slot 1 with pre-state slot 1, so that path would actually reach `BLOCK_OLDER_THAN_LATEST_HEADER` — recovering the coverage rather than dropping it.

The `StoreError` classifier living in `blockchain/src/` while the STF one is a `From` impl in `test-fixtures` is forced by the dependency direction (`blockchain → test-fixtures`), not sloppiness — no action needed.

pablodeymo (approved):

PR #552, opened 2026-07-31
fix(storage): store fork-choice votes independently

### LINKED ISSUE #504: Store fork-choice votes independently from proofs and signatures
Currently, we store proofs and signatures in memory and use them for computing fork-choice weights on the fly. We should instead store the proof-less attestations of each validator in an independent field, to avoid the FIFO eviction of proofs to affect the latest vote of each validator.

We want to have, in addition to signatures and payloads:

```rust
/// Map from validator index to its latest vote
votes: Vec<AttestationData>
```

dicethedev:

This PR fixes fork-choice vote tracking so validator votes are stored independently from aggregated proofs and gossip signatures.

Previously, fork-choice votes were derived from bounded in-memory proof/signature buffers. When those buffers evicted old entries through FIFO cleanup, a validator’s latest vote could disappear even though the vote should still count for fork choice. This PR adds a separate proof-less vote store so fork-choice weights are not affected by proof/signature retention.

## What Changed

- Updated `crates/storage/src/store.rs`
  - Added an independent latest-vote store keyed by validator index.
  - Records known fork-choice votes when aggregated payloads become known.
  - Records block-included attestation votes separately from proof storage.
  - Keeps proof buffers for aggregation/proposal behavior only.
  - Prunes stored votes once their target is finalized.

- Updated `crates/blockchain/src/store.rs`
  - Inserts block-included attestation votes into the independent vote store during block processing.

- Fork-choice votes no longer disappear when proof or signature buffers evict entries.
- Existing proof/signature buffer behavior is preserved for aggregation and block proposal.
- Latest-vote selection keeps the existing rule: higher slot wins, and same-slot ties use the canonical attestation-data root.
- Finalized-target votes are pruned once they can no longer affect fork choice.

## Tests Added / Run

- Added regression coverage for votes surviving aggregated proof FIFO eviction.
- Added coverage for pruning finalized-target votes from the independent vote store.

## Related Issues / PRs
- Closes #504 

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


crates/storage/src/store.rs
```diff
@@ -1447,6 +1513,16 @@ impl Store {
             .extract_latest_attestations()
     }
 
+    /// Insert proof-less fork-choice votes from block-included attestations.
+    pub fn insert_known_attestation_votes(&mut self, attestations: &[AggregatedAttestation]) {
```
MegaRedHand: Can we embed this inside `insert_signed_block`?

crates/storage/src/store.rs
```diff
@@ -339,6 +342,7 @@ pub type GossipSignatureSnapshot = Vec<(HashedAttestationData, Vec<(u64, Validat
 type StorageKey = Vec<u8>;
 type StorageEntry = (StorageKey, Vec<u8>);
 type BlockRootIndexChanges = (Vec<StorageKey>, Vec<StorageEntry>);
+type VoteStore = Vec<Option<AttestationData>>;
```
MegaRedHand: Let's make this a `HashMap<usize, AttestationData>` instead. That'll make this easier to reason about, and avoid performance edge-cases.

crates/storage/src/store.rs
```diff
@@ -1063,6 +1080,22 @@ impl Store {
         pruned_new + pruned_known
     }
 
+    /// Prune latest fork-choice votes whose target is at or below `finalized_slot`.
+    pub fn prune_known_votes(&mut self, finalized_slot: u64) -> usize {
```
MegaRedHand: Pruning isn't needed for votes, since we always keep the last vote of each validator only

crates/storage/src/store.rs
```diff
@@ -553,6 +557,8 @@ pub struct Store {
     backend: Arc<dyn StorageBackend>,
     new_payloads: Arc<Mutex<PayloadBuffer>>,
     known_payloads: Arc<Mutex<PayloadBuffer>>,
+    /// Latest fork-choice vote per validator, independent from bounded proof buffers.
+    votes: Arc<Mutex<VoteStore>>,
```
MegaRedHand: Let's wrap the `VoteStore` with a `ForkChoiceState` struct, so we can later extend it as needed without having a mutex for each field.

Also, another thing I forgot about is that we have to store both known and new votes (each in a separate map). Here we are storing known votes only.

crates/storage/src/store.rs
MegaRedHand: This function isn't used anywhere, right? Let's remove it.

crates/storage/src/store.rs
```diff
@@ -566,6 +572,10 @@ fn new_state_cache() -> Arc<Mutex<LruCache<H256, State>>> {
     Arc::new(Mutex::new(LruCache::new(capacity)))
 }
 
+fn new_vote_store() -> Arc<Mutex<VoteStore>> {
+    Arc::new(Mutex::new(Vec::new()))
+}
```
MegaRedHand: This is just `Default::default()`, right?

MegaRedHand (commented):
Looks good

Left some comments about pruning. The issue didn't mention it, but the main issue with how we stored votes is that proof are pruned regularly, which shouldn't affect votes: LMD-GHOST uses the latest message from each validator, so otherwise no pruning should happen.

Also, an additional PR is needed to finish addressing the issue, since we still need to replace the payloads and proofs in fork-choice with the new votes.

crates/storage/src/store.rs
dicethedev: Yep, this is `Default::default()` now. I removed the helper and wrapped the vote maps in `ForkChoiceState` as requested.

dicethedev: @MegaRedHand You can review again

crates/storage/src/store.rs
```diff
@@ -1418,20 +1434,65 @@ impl Store {
...
+        Self::record_votes(&mut fork_choice.known_votes, data, validator_ids);
+    }
+
+    fn record_new_votes<I>(&self, data: &AttestationData, validator_ids: I)
+    where
+        I: IntoIterator<Item = u64>,
+    {
+        let mut fork_choice = self.fork_choice.lock().unwrap();
+        Self::record_votes(&mut fork_choice.new_votes, data, validator_ids);
+    }
```
MegaRedHand: We should inline these. No reason to use generics for 2 lines of code

crates/storage/src/store.rs
```diff
@@ -304,6 +307,7 @@ impl PayloadBuffer {
     /// broken toward the larger canonical attestation-data root — the same rule the
     /// block-level fork-choice tiebreak applies to block roots (leanSpec #1181). The
     /// pool key is already `hash_tree_root(data)`, so the tie needs no extra hashing.
+    #[cfg(test)]
```
MegaRedHand: Why only tests? Can we remove this instead?

MegaRedHand:
There are some conflicts

dicethedev:
> There are some conflicts

You can review again

crates/storage/src/store.rs
```diff
@@ -1594,6 +1614,13 @@ impl Store {
     /// Drains the new buffer and pushes all entries into the known buffer.
     pub fn promote_new_aggregated_payloads(&mut self) {
         let drained = self.new_payloads.lock().unwrap().drain();
+        {
+            let mut fork_choice = self.fork_choice.lock().unwrap();
+            let new_votes = std::mem::take(&mut fork_choice.new_votes);
```
MegaRedHand: We should keep the underlying buffer here. I think using `.drain()` does that

crates/storage/src/store.rs
```diff
@@ -341,6 +319,13 @@ pub type GossipSignatureSnapshot = Vec<(HashedAttestationData, Vec<(u64, Validat
 type StorageKey = Vec<u8>;
 type StorageEntry = (StorageKey, Vec<u8>);
 type BlockRootIndexChanges = (Vec<StorageKey>, Vec<StorageEntry>);
+type VoteStore = HashMap<u64, AttestationData>;
+
+#[derive(Clone, Default)]
+struct ForkChoiceState {
+    known_votes: VoteStore,
+    new_votes: VoteStore,
+}
```
MegaRedHand: No need for the alias now

```suggestion

#[derive(Clone, Default)]
struct ForkChoiceState {
    known_votes: HashMap<u64, AttestationData>,
    new_votes: HashMap<u64, AttestationData>,
}
```

crates/storage/src/store.rs
```diff
@@ -1458,20 +1447,42 @@ impl Store {
...
     // ============ Attestation Extraction ============
 
-    /// Extract per-validator latest attestations from known (fork-choice-active) payloads.
+    fn should_replace_vote(existing: &AttestationData, candidate: &AttestationData) -> bool {
+        candidate.slot > existing.slot
+            || (candidate.slot == existing.slot
+                && candidate.hash_tree_root() > existing.hash_tree_root())
+    }
+
+    fn record_vote(votes: &mut VoteStore, validator_id: u64, data: &AttestationData) {
```
```suggestion
    fn record_vote(votes: &mut HashMap<u64, AttestationData>, validator_id: u64, data: &AttestationData) {
```

MegaRedHand (approved):
Good work

PR #561, opened 2026-08-03
feat(consensus): two-tier heartbeat fork choice on a 4-interval slot

MegaRedHand:
> **Draft.** `forkchoice_spectests` is red against the current fixtures, by
> construction: this branch changes the interval grid and the fork-choice shape,
> and the fixtures still encode the old ones. Details in
> [Test status](#test-status). Opening early for design review on the fork-choice
> split and on the heartbeat wire choices, before fixture regeneration.

Introduce the heartbeat committee and the two-tier (Goldfish-style) fork choice,
and re-cut the slot into 4 intervals of 1000 ms to make room for it.

The slot stays **4000 ms**. That is the load-bearing constraint: the XMSS epoch
*is* the slot, so key lifetime in wall-clock terms is untouched, every existing
`GENESIS_TIME` config stays valid, and slot numbers still line up with other
clients even while the interval grid inside the slot diverges.

### 1. Interval grid: 5×800 ms → 4×1000 ms

| interval | duty |
|---|---|
| 0 | block published at the slot boundary (built during the previous slot's interval 3); import merges the block's slot-(T-1) committee bits into the heartbeat vote store |
| 1 | attestation production; committee members republish the same signature to the global heartbeat topic; early-aggregation window check scheduled |
| 2 | `safe_target` recomputed from this slot's heartbeat votes; aggregation session starts (or already started early) |
| 3 | `lagging_head` then `fast_head`; accept accumulated attestations; build **and** publish the next slot's block, aligned to its interval 0 |

### 2. Two-tier fork choice

| value | interval | base | vote window | pool | `min_score` |
|---|---|---|---|---|---|
| `safe_target` | 2 | `latest_justified` | `slot == S` only | heartbeat (gossip) | `ceil(3K'/4)` |
| `lagging_head` | 3 | `latest_justified` | `[S-N, S)` | heartbeat ∪ `known_payloads` | `ceil(2n/3)` |
| `fast_head` | 3 | `lagging_head` | `slot == S-1`, expanding | heartbeat | `0` |

`K` is `HEARTBEAT_COMMITTEE_SIZE`, `K' = min(K, n)`, `N` is `RLMD_LOOKBACK_LIMIT`.
Every threshold is denominated in `K'`, never the raw `K` — at the default `K = 16`
a `n ≤ 16` devnet would otherwise demand more votes than there are validators and
pin the safe target to `latest_justified` forever.

### 3. Heartbeat votes on the wire — `BlockBody` unchanged

- New global gossip topic `/leanconsensus/{fork_digest}/heartbeat/ssz_snappy`,
  carrying an ordinary `SignedAttestation`. Committee membership plus an in-slot
  recency check gate admission; both the vote and its raw signature are stored by
  **every** node, not just aggregators (new `P2PToBlockChain::new_heartbeat_attestation`).
- In blocks, committee votes ride in `body.attestations`. `store::extract_heartbeat_votes`
  recovers them on import with a `data.slot == block.slot - 1` gate that exactly
  mirrors the packer's `Tier::Heartbeat`. No new `BlockBody` field, so the SSZ wire
  format is untouched.
- View-merge on import is a **union**, not a replacement: a lazy or byzantine
  proposer cannot blank out a vote the node already had from gossip.

### 4. Heartbeat fold (`crates/blockchain/src/heartbeat_fold.rs`, new)

The next slot's proposer folds the buffered committee signatures into a single
type-1 aggregate. `heartbeat_aggregation_snapshot` picks the `AttestationData`
with the most buffered signers and reduces the raw set to `B \ A` — signers not
already covered by an existing type-1 — before `aggregate_mixed`, so no validator
reaches the aggregator both raw and inside a child (`MultipleMessages`). It rides
the ordinary `AggregateProduced` → `new_payloads` → promotion path and wins at
packing time on tier alone.

### 5. Aggregation ordering (second commit)

A committee aggregator that does **not** propose the next slot now ranks subnet
candidates by tier alone (`SlotOrdering::TierOnly`), dropping the current-slot
recency bucket. Those groups are the committee's view-merge payload: the next
proposer folds them itself off the global topic, and everyone else already has
them raw. Spending an aggregator's scarce `MAX_AGGREGATION_JOBS` leanVM jobs on
recency bought nothing while starving `Tier::Finalize` / `Tier::Justify`. The
proposer's subnet fallback keeps `CurrentSlotFirst`.

### 6. Config, persistence, RPC, metrics

- `HEARTBEAT_COMMITTEE_SIZE` is a **genesis config** key, not a CLI flag: two nodes
  disagreeing about `K` disagree about which bits of a block are heartbeat votes,
  which is a silent fork-choice divergence. Optional (defaults to
  `DEFAULT_HEARTBEAT_COMMITTEE_SIZE`), rejected at load if `0` or `> MAX_HEARTBEAT_COMMITTEE_SIZE`.
- Persisted in `Metadata` on first boot and authoritative thereafter; a later
  config edit is warned about, not applied. `lagging_head` is persisted next to
  `safe_target` so a restart resumes from a real tree base. New
  `get_optional_metadata` lets an existing database upgrade in place instead of
  forcing a resync.
- `GET /lean/v0/config/spec` now serves `HEARTBEAT_COMMITTEE_SIZE`, read from the
  store rather than the config file.
- `lean_lagging_head_slot` alongside `lean_head_slot`, plus eight metrics under a
  new *Heartbeat / two-tier fork choice* section in `docs/metrics.md`. The two
  worth alerting on:
  `lean_heartbeat_committee_participation` against `ceil(3K'/4)` (safe-target stall
  predictor) and `lean_fast_head_window_slots` (`> 1` means slots were missed).
  `lean_tick_interval_duration_seconds` buckets re-centred on 1.0 s.

**Interop-breaking, network-wide.** A node on this branch does not interoperate
with a `main` node: different interval grid and an extra gossip topic. This needs
a coordinated devnet, not a rolling upgrade.

**Preserved:** slot duration and therefore XMSS epoch semantics; `GENESIS_TIME`
configs; the SSZ wire format of every container; databases written by `main`
(the new metadata keys default rather than panic).

**Unchanged: 3SF-mini justifiability.** An earlier revision of this branch had
dropped `slot_is_justifiable_after` and collapsed finalization to
`source.slot + 1 == target.slot`. That is reverted — the two-tier fork choice
changes which head each interval computes, not which slots may be justified, and
unjustifiable slots still do their job of funnelling votes under latency. The
rule is back at all four sites that had diverged (`is_valid_vote`,
`try_finalize`, `get_attestation_target_with_checkpoints`, and the block
builder's `finalizes` / `entry_passes_filters`).

## Test status

`cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`
are clean. `cargo test --workspace --profile release-fast --no-fail-fast`:

| target | result |
|---|---|
| all unit tests (blockchain, storage, rpc, p2p, types, state_transition, …) | ✅ green |
| `signature_spectests`, `ssz_spectests`, `test_driver_e2e` | ✅ green |
| `stf_spectests` | ✅ 74 / 74 |
| `forkchoice_spectests` | ❌ 97 / 122 |

`forkchoice_spectests` is **122/122 on `main`**, so the 25 failures are this
branch's, not fixture drift.

Keeping 3SF-mini justifiability accounts for the difference against the earlier
revision of this branch: `stf_spectests` went 70/74 → 74/74 (all four were
state-root mismatches in tests asserting that rule, e.g.
`test_non_adjacent_justification_finalizes_across_non_justifiable_gap`), and
`forkchoice_spectests` went 91 → 97, clearing all of
`test_attestation_target_selection`.

The remaining 25 are the two genuine shape changes, and no longer touch
justifiability at all:

| group | count |
|---|---|
| tick grid (`test_tick_system`, `test_tick_acceptance_branches`) | 4 |
| safe target threshold (`test_safe_target`, `test_safe_target_supermajority`) | 5 |
| two-tier head (`test_equivocation`, `test_head_movement`, `test_fork_choice_head`, `test_fork_choice_reorgs`, `test_lmd_latest_message`, `test_finalized_safety`) | 9 |
| other (`test_prune_finalized_orphaned_branch`, `test_store_pruning`, `test_gossip_*_validation`, `test_checkpoint_sync`, `test_block_production`) | 7 |

They have not been walked individually yet. That audit plus fixture regeneration
against the updated spec is the remaining work before this leaves draft.

## Remaining before merge

- [ ] Regenerate `leanSpec/fixtures` against the new grid and two-tier head, and
      confirm every one of the 25 failures is a fixture expectation rather than a defect
- [ ] Multi-client devnet run to size `K` against real interval-1 timing
      (`lean_heartbeat_fold_time_seconds` is the budget check)

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
- [ ] Ran `make test` — `forkchoice_spectests` red, see [Test status](#test-status)

## Review note: incidental churn removed

`10db765` reverts changes that were not required by the heartbeat work, so the
diff is 20 files / +3183 −199 rather than 22 / +3342 −245. Reverted:
`build_block`'s `BlockTarget` parameter object (the new field moved to the
existing `ProposerConfig`), the `SlotInterval::HeadUpdate` rename (which also
collided with `store::HeadUpdate`), the `lean_fast_head_slot` gauge (a duplicate
of `lean_head_slot`), the `lagging_head` RPC / Hive-driver fields and
`RLMD_LOOKBACK_LIMIT` in `/config/spec` (no readers), an unused re-export, a dead
parameter, and two comment-only edits.

Deliberately kept: `JobSource`, which looks like the same kind of noise but is
the only thing routing prover time into `lean_heartbeat_fold_time_seconds`; and
the `Tier` / `EntryEffect` split, which is load-bearing because `Tier::Heartbeat`
sorts below `Justify` and would otherwise let a thin committee vote mark its
target justified in the projection.


MegaRedHand:
We'll update the design before reopening

PR #562, opened 2026-08-03
docs: move the storage reference into docs/data_storage.md

MegaRedHand:
## Motivation

`CLAUDE.md` carried ~60 lines of storage documentation that duplicated `docs/data_storage.md`, and the copy had drifted: it claimed 7 tables and omitted `BlockRoots` entirely, so a reader never learns that the canonical `slot -> root` index exists or that head updates rewrite it.

Rather than fix the duplicate, this collapses it into a pointer, matching how the `HTTP Servers` section already defers to `docs/rpc.md`. `docs/data_storage.md` becomes the single place storage is described.

## Changes

**`docs/data_storage.md`** gains everything that was missing:

- A `BlockRoots` section: `slot -> H256`, rewritten on every head update inside `update_checkpoints` via `block_root_index_changes`, which walks both branches to their common ancestor so a reorg touches only the affected slot range. Added to the table list, the key-encoding section (it has its own slot-only layout), the write-path diagram, the never-pruned list, and the architecture diagram.
- What never changes at runtime: `Metadata["config"]` is written once by `init_store`, has no setter, and doubles as the DB fingerprint since `from_db_state` refuses to resume a data directory whose `genesis_time` disagrees with the node's config file. The genesis validator registry is constant too but rides inside `States` snapshots rather than having a table.
- The `StateDiff` bet spelled out: a diff carries no copy of `config` or `validators`, so a state transition that mutated either would silently corrupt every reconstructed state. Contrast with `historical_block_hashes`, which is also omitted but is checked by `validate_history_append` instead of trusted.
- Corrections: seven -> eight tables throughout; `ChainConfig` described as "genesis time, validator count" when it only has `genesis_time`; `init_store`'s `BlockRoots` write was missing from the startup prose; remaining hardcoded `1024`s replaced with `SNAPSHOT_ANCHOR_INTERVAL`.

**`CLAUDE.md`** drops from ~60 lines to a short pointer paragraph plus three bullets kept inline because they are easy to get wrong from memory: `BlockSignatures` is the only pruned block table, `StateDiff` trusts `config`/`validators` never to mutate, and `Metadata["config"]` is write-once and acts as the resume fingerprint.

## Note on an error in the first commit

The first commit claimed `BlockRoots` backs `get_block_by_slot`. There is no such function. `BlockRoots`' only production reader is `get_signed_blocks_by_slot_range`, which serves BlocksByRange over req/resp. The RPC `GET /lean/v0/blocks/:slot` resolves through the head state's `historical_block_hashes` (`resolve_slot` in `crates/net/rpc/src/blocks.rs`), which is why a block on a side fork is reachable there by root but never by slot. The second commit corrects this and documents the distinction.

## Testing

Docs only, no code touched. `mdbook build` with `mdbook-linkcheck2` passes.


docs/data_storage.md
```diff
@@ -233,6 +269,16 @@ small under healthy finality). The rest is deliberately omitted:
 | `latest_block_header`     | The `BlockHeaders` table                                                                                                                                      |
 | `historical_block_hashes` | Regenerated from `base_root` + the slot gap (the state transition appends the parent root plus one zero per skipped slot, so the append is fully predictable) |
 
+Omitting `config` and `validators` is a bet, not a fallback: a diff carries no
```
pablodeymo: we can improve this text

pablodeymo (approved):

PR #579, opened 2026-08-12
feat(p2p): add opt-in discv5 peer discovery

MegaRedHand:
## What

Adds opt-in discv5 peer discovery, so a lean node can find peers instead of
being handed them. Off by default; `--discovery.enable` turns it on and needs
nothing else: discv5 keeps port `9000` and libp2p QUIC now defaults to `9001`,
so the two UDP sockets no longer collide. Overriding either onto the other is
still rejected at startup rather than failing later with an opaque
`EADDRINUSE`. Static bootnode dialing is untouched.

Built on ethrex's discovery stack: `DiscoveryServer` runs discv5-only and
writes what it finds into a `PeerTable`, which `P2PServer` polls, filters and
dials over libp2p QUIC. We build the local ENR ourselves and hand it to
`spawn`, so the record ethrex answers queries with is the one this node
reports.

## How peers are judged

Admission follows the beacon phase0 p2p spec, mirroring lighthouse's
`eth2_fork_predicate`:

| Check | Rule |
| --- | --- |
| `eth2` entry | must be present and decode |
| `fork_digest` | must equal ours |
| `next_fork_version` / `next_fork_epoch` | may differ (the spec's MAY) |
| `quic` port | required, and non-zero |
| `secp256k1`, `ip`/`ip6` | required to derive a dialable target |

These live in a `LeanFilter` handed to the peer table as its `PeerFilter`, so
each record is judged the moment it arrives rather than at dial time. No
rejection is final: the peer table re-runs the filter as soon as the peer
publishes a higher-`seq` ENR, so a node that adds a `quic` entry or gains an
address through discv5's IP voting is reconsidered without a restart.

Admitted peers are ranked by how many attestation subnets they advertise that
no connected peer covers, so discovery fills coverage gaps first. `attnets` is
self-reported and unauthenticated, so subnet ids at or beyond the local
committee count are dropped before ranking sees them: otherwise an ENR padding
its bitfield with a few hundred bytes of `0xFF` would outrank every honest peer
forever.

Dialing stops once `--discovery.target-peers` peers are connected and resumes
when the count drops back below it. That is the whole of what the flag does: it
is the dial loop's cutoff, and it is deliberately not handed to ethrex's peer
table (see the limitation on lookup pacing below). A target of `0` means
"discover and serve, never dial".

## Also here

- `--discovery.advertise-ip` separates the bound address from the advertised
  one, for a node behind NAT or on a host whose public IP is not what it binds.
- A bootnode entry no longer needs a `quic` port: one with only a `udp` entry
  is kept as a discv5 seed even though it cannot be dialed over libp2p.
- `lean_discovered_peers_dialed_total` counts dials discovery initiated, as
  distinct from the static bootnode dials every node makes. Connection outcomes
  are not duplicated: a discovery dial that succeeds or fails already shows up
  in `lean_peer_connection_events_total`.
- `docs/discovery.md` covers the ENR layout, the admission rules, the operator
  flags and the known limitations.

## On the dependency

The heavy ethrex crates were already in the graph on `main`: `ethrex-levm`,
`ethrex-vm`, `ethrex-blockchain`, `ethrex-storage` and `ethrex-trie` all
predate this PR, which already depended on `ethrex-p2p` by `rev`. What this
changes is the pin, from that `rev` to the `v25.0.0` release tag. The delta is
`ethrex-guest-program` and `ethrex-l2-common` in, `ethrex-threadpool` out, and
the workspace's total crate count goes **down**: 722 to 712 unique names in
`Cargo.lock`.

## Known limitations

**One lean devnet is not separated from another.** Lean's `fork_digest` is the
hardcoded cross-client dummy `0x12345678`, so the `eth2` check separates lean
from non-lean but **not one lean devnet from another**. Two devnets running
this code will peer with each other. Closing that needs lean to adopt a
genesis-derived fork digest, which is a cross-client change to gossip topic
names.

**discv5 lookups run at the startup rate.** ethrex paces its iterative lookups
by how full its own peer table is, easing from one every 500ms to one every 10s
as the table fills. That table only counts peers registered through
`NewConnectedPeer`, which carries an RLPx connection: ethlambda connects over
libp2p and registers nothing, so the count is pinned at zero and the pacing
never eases. Lookups keep firing every 500ms for the life of the process,
roughly 20x the intended steady-state `FindNode` traffic. This is also why
`--discovery.target-peers` is not passed to that table: a target of `0` would
make the pacing divide by zero, and `NaN as u64` saturates to a zero-delay
re-fire. Closing it properly means ethrex learning about non-RLPx connections,
which is an upstream change.

**`attnets` is not a fixed-width SSZ `Bitvector`.** The spec's is
`Bitvector[ATTESTATION_SUBNET_COUNT]`, a constant every conformant client
shares, which is what makes an undelimited bitfield decodable. ethlambda
derives the width from `attestation_committee_count`, which is runtime
configuration, so two nodes can legitimately exchange bitfields of different
lengths. The bit-packing convention is the spec's; only the width is
negotiable. Readers tolerate a foreign length by treating bits past the end as
unset, and a peer's advertised subnets are clamped to the local committee count
before they influence anything.

## Changes since the first revision

- **The ENR we reported and the ENR we served are now the same record.**
  `DiscoveryServer::spawn` used to take an ethrex `Store` and derive its own
  record, so what it answered queries with carried `ip`/`udp`/`secp256k1` but
  none of `eth2`, `attnets` or `quic`: a lean peer applying these same rules to
  it would have refused us. `spawn` now takes a prepared `NodeRecord` and we
  pass what `build_local_enr` produces. IP voting `edit`s and re-signs that
  record rather than rebuilding it, so the consensus entries survive a sequence
  bump. The empty in-memory `Store` and the `ethrex-storage` dependency are
  gone.
- **The ENR is no longer exposed on `GET /lean/v0/node/identity`.** It is
  logged once at startup instead; the RPC crate is untouched by this PR.
- **A `udp: 0` bootnode entry is now read as no entry at all**, matching what
  `quic: 0` already did, so discv5 is never seeded with a contact on a port
  nothing listens on.
- **A bootnode list that yields no dial target now says so at `warn`.** Both
  the every-entry-unusable case and the every-entry-`quic`-less case (a
  beacon-chain bootstrap list) previously left the node isolated with nothing
  above `debug`.
- **Discovery bookkeeping is no longer dropped for a live peer**: a failed
  second dial to an already-connected peer used to clear its `attnets`, making
  subnet-coverage ranking under-count.
- **Discovery starts inside `P2P::spawn`**, which takes an
  `Option<DiscoverySpawnConfig>` and owns the resulting handle, rather than
  `main` spawning the server and threading a handle in.
- **`--discovery.target-peers` replaces the hardcoded dial cutoff**, and is
  deliberately not handed to ethrex's peer table: see the lookup-pacing
  limitation above for what that would have done at `0`.
- **The ethrex dependency is pinned to the `v25.0.0` release tag.** The first
  revision tracked the unmerged `feat/discovery-peer-requirements` by `branch`,
  which let a force-push change what this built against. That work has since
  landed and shipped, so the pin names a release rather than a moving branch or
  a bare commit.
- **`--discovery.enable` works on its own.** `--gossipsub-port` now defaults to
  `9001`, one above `--discovery.port`. Both are UDP and both defaulted to
  `9000`, so enabling discovery without also moving a port was a guaranteed
  startup failure: a default that is never valid. The collision check stays for
  the case where an operator sets them explicitly. Every devnet script passes
  `--gossipsub-port` already, so nothing in tree changes behaviour.
- **Discovery bookkeeping is only recorded once the swarm takes the dial.**
  `peer_attnets` and `lean_discovered_peers_dialed_total` were written before
  `swarm.dial`, but libp2p rejects some dials *synchronously* — `LocalPeerId`,
  `NoAddresses`, `Denied`, and `DialPeerConditionFalse` (already connected, or
  already dialing) — with no `OutgoingConnectionError` to follow, so nothing
  ever cleared the entry or corrected the counter. The realistic trigger is a
  discovery dial racing an in-flight bootnode redial to the same peer. Ranking
  was never wrong (`covered_subnets` filters on `connected_peers`), but the map
  grew unbounded over the process lifetime and the counter overcounted.
  `SwarmHandle::dial_accepted` now reports back whether the swarm queued the
  dial, and the insert and the metric happen only then.

## Testing

- `make lint` clean.
- `make test`: 604 passed, 0 failed, 7 ignored, including the forkchoice,
  signature, STF and SSZ spec tests.
- New: the default ports let `--discovery.enable` stand alone, and the
  collision check still fires when both are pointed at one port.
- Unit coverage for the ENR round trip, every admission rejection reason, the
  oversized-`attnets` ranking attack, subnet ranking, subnet coverage counting
  only connected peers, both zero-port bootnode cases (`udp: 0` read as absent,
  and a record dropped when neither port is dialable), and `spawn_discovery`
  binding a real socket (including `--discovery.advertise-ip` and a busy port).


crates/net/p2p/src/lib.rs
```diff
@@ -374,11 +381,34 @@ pub struct P2P {
...
+    ///
+    /// `discovery` is `Some` when discv5 discovery is enabled: the discv5
+    /// server is started here, and its handle seeds the dial loop's state and
+    /// schedules its first tick. `None` leaves the dial loop permanently
+    /// dormant, so peering relies solely on the static bootnode list dialed by
+    /// `build_swarm`.
+    ///
+    /// Discovery is started before the swarm adapter so a fatal discovery
+    /// failure (a busy UDP port, say) surfaces before any actor is running.
+    pub async fn spawn(
```
MegaRedHand, line 408: The `async` here is only needed in the discv4 path of ethrex's discovery server, and it can probably be refactored to not need it. I leave it like this for now since it's not really a problem for the integration, but it's something to keep in mind for the future.

crates/net/p2p/src/lib.rs
```diff
@@ -374,11 +381,34 @@ pub struct P2P {
...
+        built: BuiltSwarm,
+        store: Store,
+        node_names: HashMap<PeerId, String>,
+        discovery: Option<DiscoverySpawnConfig>,
+    ) -> Result<P2P, DiscoveryError> {
```
MegaRedHand, line 413: This `Result` can probably be removed too.

crates/net/p2p/src/discovery/mod.rs
```diff
@@ -0,0 +1,280 @@
...
+    // `spawn` rather than `spawn_with_filter` would install ethrex's own filter,
+    // which wants an EIP-2124 `eth` entry compatible with an execution chain lean
+    // does not have and rejects any record without one, so every lean contact
+    // would be stamped rejected and never dialed.
+    //
+    // The peer table owns the filter it runs, so the dial loop keeps a clone
+    // rather than sharing one: the two carry the same fork id and committee
+    // count, which is what makes their judgments agree.
+    let filter = LeanFilter::new(EnrForkId::local(), config.attestation_committee_count);
+    let peer_table = PeerTableServer::spawn_with_filter(
```
MegaRedHand, line 159: ethrex paces its lookups by `peers.len() / target_peers`, but peers is only populated by `NewConnectedPeer`, which arrives from ethrex's RLPx layer. We never send it, so the ratio is pinned at 0 and lookups already run at the fast 500ms end.

`peer_table.new_connected_peer` takes a `PeerConnection`, so fixing this would require changes from ethrex's side. Let's revisit this after this PR is merged

pablodeymo:
Read the full diff, then pulled ethrex at the pinned commit `bf401280` to verify the load-bearing claims about `DiscoveryServer::spawn`, `PeerFilter`, `extra_int`, and IP voting. Three of the four check out exactly as documented. One integration assumption does not.

## 1. `target_peers` is fed to a peer table this node never populates

This is the one I'd want fixed before this runs anywhere real.

`spawn_discovery` passes `config.target_peers` to `PeerTableServer::spawn_with_filter`, but ethlambda only ever calls `get_contact_to_initiate()` on that table (`dial.rs:118`) — it never registers a connected peer. ethrex's `peers` map is populated solely by `handle_new_peer`, which takes an RLPx connection + capabilities (`peer_table.rs:594-597`); we dial over libp2p and never go through it. So `self.peers.len()` is **permanently 0**.

That number drives discv5's lookup pacing:

```rust
// disc_server.rs:404
let peer_completion = self.peer_table.target_peers_completion().await...;  // peers.len() / target_peers
lookup_interval_function(peer_completion, ITERATIVE_LOOKUP_INITIAL_MS /*500*/, ITERATIVE_LOOKUP_INTERVAL_MS /*10_000*/)
```

Reproducing `lookup_interval_function` standalone against the real constants:

| `--discovery.target-peers` | completion | lookup interval |
|---|---|---|
| 200 (default) | `0/200` = 0.0 | **500ms, forever** (never eases to the 10s steady state) |
| 0 (documented mode) | `0/0` = **NaN** | **0ns** |

- **Default:** discv5 iterative lookups stay pinned at the *startup* rate for the life of the process — roughly 20x the intended steady-state FindNode traffic, permanently.
- **`--discovery.target-peers 0`:** `NaN` survives the easing curve, and `NaN as u64` saturates to 0, so `send_after(Duration::ZERO, ..., LookupV5)` re-fires immediately. That's an unthrottled lookup loop spinning the actor and flooding FindNode packets. The CLI help advertises this exact value as supported: *"0 means discover and serve, never dial."*

Two things to fix: guard `target_peers == 0` (in `validate_discovery`, or clamp in `spawn_discovery`), and decide what the flag actually means for the peer table — either feed connection state back into it, or stop passing `target_peers` there and document the flag as dial-loop-only. Right now `docs/discovery.md`'s *"The same number sizes the discv5 peer table"* is true but functionally inert, which is the more misleading of the two.

## 2. The PR description promises an RPC change that isn't in the diff

> `GET /lean/v0/node/identity` reports the local ENR alongside the peer id, grouped into a `NodeIdentity` struct.

`crates/net/rpc/` isn't among the changed files, and `main`'s `IdentityResponse` carries only `version` and `peer_id`. There is no `NodeIdentity` struct anywhere. Correspondingly, `DiscoveryHandle::local_enr` is built, logged once, and then dropped — `DiscoveryState::new` doesn't keep it, so in production it's write-only (its own doc comment says "a *future* RPC identity endpoint", which is the honest version). Either drop the bullet or land the endpoint.

## 3. The dependency bump is much larger than "add a discovery dep"

`ethrex-p2p` goes **8.0.0 -> 24.0.0**, dragging 65 ethrex lock entries and 891 deletions of unrelated lockfile churn. It also lands a **duplicate `libssz` (0.2.2 alongside 0.3.0)** in the graph. The PR body notes the libssz split and that nothing SSZ-typed crosses the boundary — that's correct, `EnrForkId` is lean's own type — but the review surface here is a whole ethrex upgrade riding along with the feature, not just the discovery seam.

## 4. Minor

- **`read_quic_port`'s doc comment is wrong about why it works.** It claims "an absent entry RLP-decodes to `0u16` via left-padding." It doesn't — ethrex's `extra_int` returns `None` on the `.find(...)?` before any decode happens (`types.rs:452-455`). The `.filter(|port| *port != 0)` is still needed for an explicit `quic: 0`; only the stated reason is off.
- **`forget_discovered_peer` can fire for a still-connected peer.** It's called unconditionally on `OutgoingConnectionError` (`lib.rs:726`). If a *second* dial to an already-connected discovery peer fails — e.g. via the bootnode redial path — its `attnets` are dropped while the peer is live, so `covered_subnets` under-counts. Only affects ranking eagerness, never correctness. Gating on `!server.connected_peers.contains(&pid)` would close it.

## Already flagged above, worth keeping

- **`udp: 0` is accepted verbatim** as a discv5 seed — real, and it's the exact asymmetry the code deliberately avoids for `quic`. Worth fixing for consistency alone.
- **Bootnode parsing degrading to warnings** means a fully-malformed bootnode file boots an isolated node silently. Also note `build_swarm` skips quic-less bootnodes at `debug!` level, so feeding a beacon-chain list yields zero static dials with nothing at `warn!`.
- **`branch = "..."` instead of `rev = ...`, while the PR is out of draft.** The body still says "Draft because the ethrex dependency is still an unmerged branch," but the PR is no longer a draft. This is the merge blocker.

## What's solid

Checked the claims that mattered and they hold. `update_local_ip` really does `edit`+re-sign rather than rebuild, preserving `eth2`/`attnets`/`quic` across an IP-voting seq bump — and ethrex has tests for exactly that (`update_local_ip_preserves_entries_it_does_not_touch`). The serve/report ENR unification is genuinely closed: one `NodeRecord` feeds both `enr_url()` and `spawn`. `subnets_from_attnets` iterating the local committee rather than the peer's bitfield is the right defense and is tested with a 290-byte hostile pad. The `set_extra` vs. bare-`Vec<u8>` footgun the comments warn about is real (`types.rs:465`) and correctly avoided. Test coverage on admission, ENR round-trip, and bootnode parsing is genuinely thorough, and `docs/discovery.md` is unusually good — including honest limitations.

Finding 1 is the only one I'd call blocking on its own merits, alongside the dependency pin.

pablodeymo (commented):
CI is green (Lint, Test 602 passed, Build, License) and the branch is `MERGEABLE`. This is genuinely strong work — the docs and the comments carry real reasoning, and the test coverage on the adversarial paths is better than most PRs get. One thing I'd want fixed before we rely on it, one description fix, and a handful of nits.

I checked the two headline claims against the actual ethrex v25.0.0 source rather than taking the description's word:

- **ENR extras survive IP voting** ✅ — `NodeRecord::edit` (`types.rs:598`) clones pairs, applies the closure, bumps `seq`, re-signs; `update_local_ip` touches only `ip`/`ip6`. The "we serve the record we report" fix is real.
- **ENRs are signature-verified before reaching `LeanFilter`** ✅ — `peer_table.rs:1525` and `discv5_handlers.rs:202`. So `attnets`/`eth2` can only ever be *self*-attested, never forged under another node's key, which is exactly the threat the oversized-`attnets` clamp is scoped to.
- Also ✅: libp2p identity and the discv5 node id derive from the same secp256k1 key (`lib.rs:277` vs `discovery/mod.rs:126`), so the self-dial guard at `dial.rs:90` actually works. And `draw_candidates` is right that `passes_filter == None` contacts are still offered (`peer_table.rs:1332`), and correctly drops them via `contact.record.as_ref()`.

---

## 1. `admit()` does no sanity check on the advertised IP — and our own default publishes `0.0.0.0`

`discovery/admission.rs:152` takes whatever `read_ip` returns and feeds it straight into `quic_multiaddr`. There is no filter for unspecified, loopback, link-local, multicast or broadcast addresses. I checked, and **ethrex does not filter either**: `is_private_ip` appears only in the IP-vote accept path (`discv5/server.rs:148`) and in WHOAREYOU rate limiting (`discv5_handlers.rs:766`), never against discovered records. So `LeanFilter` is the only gate, and it does not gate this.

This is not hypothetical, because of how the defaults compose:

- `main.rs:131` hardcodes the bind IP to `0.0.0.0`
- `discovery/mod.rs:137`: `advertise_ip.unwrap_or(config.bind_ip)`
- so a node with `--discovery.enable` and no `--discovery.advertise-ip` publishes a **signed, admissible** ENR advertising `0.0.0.0`

Every peer that receives it builds `/ip4/0.0.0.0/udp/<quic>/quic-v1/p2p/<id>` and dials it. On Linux that connects to localhost, so each reader dials *its own* QUIC listener, fails the peer-id check, and repeats. The `warn!` at `discovery/mod.rs:198` correctly tells the publisher, but the cost lands on every reader, who gets no warning at all.

Two fixes, and I'd take both:

- **Cheap, closes the common case:** make `--discovery.advertise-ip` required when `--discovery.enable` is set, in `validate_discovery` alongside the existing port check. Five lines, and it makes the footgun unreachable.
- **The actual hardening:** a `RejectReason::UnroutableAddress` in `admit()`. Careful with a blanket `is_global()` check — `docs/discovery.md:36` *recommends* `127.0.0.1` for local devnets, so that would break the documented setup. Suggestion: reject unspecified/multicast/broadcast unconditionally, and allow loopback/private only when our own advertised IP is in the same class.

The feature is off by default, so this is "fix before anyone turns it on" rather than a hard blocker on the merge itself. Fine by me either here or as an immediate follow-up.

## 2. The description is stale on the dependency

The whole **Dependency** section argues for a decision that was then reversed. It says `rev` `275da92` (24.0.0), "still pinned by `rev`", "ethrex publishes no release to name", "Still a `rev` and not a moving pin". The branch head is `7751fb5 chore(deps): pin ethrex to the v25.0.0 release tag` → `tag = "v25.0.0"` (`f3b90bc`). Same for the last bullet under "Changes since the first revision".

The rest of the numbers hold up, so it is just that section: the lockfile delta is 486/783 against the body's "~485 / ~780" ✅, and the crate delta is exactly as described ✅.

**Worth adding to the description, because it pre-empts the obvious objection:** the heavy ethrex crates were *already* in the graph on `main` — `ethrex-levm`, `ethrex-vm`, `ethrex-blockchain`, `ethrex-storage` and `ethrex-trie` all predate this PR. This adds `ethrex-guest-program` + `ethrex-l2-common` and drops `ethrex-threadpool`. The net dependency count actually goes **down**, 722 → 712. Saying that stops anyone relitigating "why is an EVM in our consensus client".

(The Test job at 20m40s vs ~10min on main is almost certainly the cold cache from the changed pin rather than real cost, but a warm-cache re-run would settle it.)

## 3. `peer_attnets` and the dial counter drift when `swarm.dial` fails synchronously

`dial.rs:100-104` inserts into `peer_attnets` and increments `lean_discovered_peers_dialed_total` *before* the dial. But `swarm_adapter.rs:150-154` swallows a synchronous `swarm.dial` error at `debug!`, and libp2p returns `Err` synchronously — with **no `OutgoingConnectionError` event** — for `DialError::LocalPeerId`, `NoAddresses`, `DialPeerConditionFalse` (already connected *or already dialing*) and `Denied`. In those cases `forget_discovered_peer` never runs.

Not a correctness bug — `covered_subnets` filters on `connected_peers`, so a stale entry contributes nothing to ranking — but the map is unbounded over the process lifetime and the counter overcounts. `DialPeerConditionFalse` is the realistic trigger: a discovery dial racing an in-flight bootnode redial to the same peer.

Cheapest fix is moving the insert and the metric into the `ConnectionEstablished` handler, where you know the dial took.

## 4. `--discovery.port` defaults to 9000, same as `--gossipsub-port`, so `--discovery.enable` alone always fails

`cli.rs:140`, with `validate_discovery` at `cli.rs:166`. The error message is good, but a default that is *never* valid is a strange default: `--discovery.enable` on its own is a guaranteed startup failure. `9001`, or `gossipsub_port + 1`, would make the flag work standalone. Pre-existing from the earlier CLI PR, but this is the PR that gives the flag meaning, so it seems like the right place to fix it.

## 5-8. Nits

- **Ranking goes stale mid-batch.** `dial_tick` ranks once at refill (`dial.rs:84-86`), then drains one per tick over up to 8 ticks / 40s. By candidate #8 the coverage snapshot is 40s old, which undercuts the stated "fill coverage gaps first" goal. Recomputing `covered` and picking the max on each pop is O(8).
- **Skipped candidates are dropped for good.** `dial.rs:89-94` `continue`s an already-connected candidate out of the queue, and ethrex has already put it in `already_tried_peers` (`peer_table.rs:1338`), which only clears when a full pool scan finds nothing. A peer that disconnects later is not re-offered promptly. Self-heals in a small devnet, slower in a big pool — probably just wants a comment.
- **`DiscoveryHandle::local_enr` is dead outside tests.** `spawn_discovery` logs it at `discovery/mod.rs:197`, then `DiscoveryState::new` drops it. The doc calls it "what a future RPC identity endpoint would read", but that endpoint was explicitly reverted in `7a02f094`. Either drop the field, or drop the speculative sentence.
- **`fork_digest()` parses a `&str` at runtime and panics** (`enr.rs:65-69`). Cosmetic — `FORK_DIGEST` also builds every gossip topic, so a malformed one fails elsewhere first — but it could be `const`.

One observation rather than a finding: `--discovery.target-peers` defaults to 200 at one dial per 5s, which is ~17 minutes to saturate. Fine for realistic pool sizes, but the default reads more ambitious than the dial loop can deliver.

pablodeymo (approved):
Great work! 🚀

PR #587, opened 2026-08-25
fix(p2p): summarize the undelivered response instead of Debug-dumping it

MegaRedHand:

libp2p's `request_response::Behaviour::send_response` has signature
`Result<(), TCodec::Response>`: on failure it hands the **whole undelivered
response back as the error value**. The swarm adapter logged that with `?err`,
so what looked like an error description was actually a complete
`BlocksByRoot`/`BlocksByRange` answer, Debug-expanded: every block header,
every `H256` as a 32-element decimal array, and every attestation bitlist
printed byte-by-byte.

On devnet-5 this produced single log lines of 200-278 KB. Loki rejected them
outright with `400 max entry size '262144' bytes exceeded`, so those lines were
dropped rather than stored, on every ethlambda node in the fleet (1-16
rejections per node per day; the underlying failure fires more often than that,
only the larger dumps breach the cap). No other client showed this.

`SignedBlock` already hand-writes `Debug` to truncate its opaque proof bytes
for exactly this reason (`proof: <218441 bytes>`), so the largest single field
was already handled. The surrounding envelope never was.

Note that #586 lowered this call site from `warn!` to `debug!`, which hides the
symptom at the default log level but leaves the unbounded dump in place for
anyone running with debug logging. This makes the log line correct rather than
just quiet.

## What Changed

- `crates/net/p2p/src/req_resp/messages.rs` — new `Display` impl for `Response`
  producing a bounded summary:
  - `Success(Status head=9/cdcdcdcd finalized=7/abababab)` (roots via
    `ShortRoot`, per the logging convention in `CLAUDE.md`)
  - `Success(Blocks count=1024 first_slot=0 last_slot=1023)`
  - `Error(RESOURCE_UNAVAILABLE(3): block not found)`
- `crates/net/p2p/src/swarm_adapter.rs` — log via `%response` instead of
  `?err`, and rename the binding from `err` to `response` since it is the
  payload, not a diagnostic. Comment at the call site records why.

Blocks are reported as `first_slot`/`last_slot` rather than as a range: a
`BlocksByRoot` response follows the requested-root order, so its slots are not
necessarily contiguous or ascending.

- Log output only. No protocol, wire-format, or control-flow change; the
  `send_response` result is still discarded exactly as before.
- The derived `Debug` on `Response` is intentionally left in place for when the
  full detail is genuinely wanted. Only the log call site changes.
- The rendering is O(1) in payload size, so a maximum-size response
  (`MAX_REQUEST_BLOCKS`) still fits comfortably inside any log backend's line
  limit.

## Tests Added / Run

No tests added.

```
make fmt
make lint
make test
cargo test -p ethlambda-types -p ethlambda-p2p --profile release-fast
```

## Related Issues / PRs

- Related to #586 (lowered this call site to `debug!`, masking the symptom)

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
      (564 passed, 0 failed, 7 ignored, across 19 suites)

### DIFF
--- a/crates/net/p2p/src/swarm_adapter.rs
+++ b/crates/net/p2p/src/swarm_adapter.rs
@@ -167,11 +167,14 @@ fn execute_command(swarm: &mut libp2p::Swarm<Behaviour>, cmd: SwarmCommand) {
             }
         }
         SwarmCommand::SendResponse { channel, response } => {
+            // On failure libp2p hands the whole undelivered response back as the
+            // error value, so this is the response itself, not a diagnostic.
+            // Log it via `Display` (a bounded summary), never `Debug`.
             let _ = swarm
                 .behaviour_mut()
                 .req_resp
                 .send_response(channel, response)
-                .inspect_err(|err| debug!(?err, "Swarm adapter: send_response failed"));
+                .inspect_err(|response| debug!(%response, "Swarm adapter: send_response failed"));
         }
     }
 }
--- a/crates/net/p2p/src/swarm_adapter.rs
+++ b/crates/net/p2p/src/swarm_adapter.rs
@@ -167,9 +167,6 @@ fn execute_command(swarm: &mut libp2p::Swarm<Behaviour>, cmd: SwarmCommand) {
             }
         }
         SwarmCommand::SendResponse { channel, response } => {
-            // On failure libp2p hands the whole undelivered response back as the
-            // error value, so this is the response itself, not a diagnostic.
-            // Log it via `Display` (a bounded summary), never `Debug`.
             let _ = swarm
                 .behaviour_mut()
                 .req_resp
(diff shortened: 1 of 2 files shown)


crates/net/p2p/src/swarm_adapter.rs
```diff
@@ -167,11 +167,14 @@ fn execute_command(swarm: &mut libp2p::Swarm<Behaviour>, cmd: SwarmCommand) {
             }
         }
         SwarmCommand::SendResponse { channel, response } => {
+            // On failure libp2p hands the whole undelivered response back as the
+            // error value, so this is the response itself, not a diagnostic.
+            // Log it via `Display` (a bounded summary), never `Debug`.
```
MegaRedHand:
```suggestion
```

pablodeymo (approved):

PR #589, opened 2026-08-26
feat(p2p): enable gossipsub peer scoring and react to slow peers

MegaRedHand:
Supersedes #588, which is closed: the `flood_publish(false)` change is included here.

## Problem

A peer whose gossipsub send queue fills is never removed from our mesh, so every later publish and forward is attempted and dropped forever. On devnet-5 one node lost roughly **33 messages/second to one peer, at a flat rate, for over 12 hours**, while request/response to that same peer kept working throughout. 642k `Send Queue full` lines in one hour network-wide, and 64% of the dropped RPCs were real data:

| dropped RPC | 30 min | share |
|---|---:|---:|
| Forward | 80,949 | 37.5% |
| IDontWant | 73,477 | 34.0% |
| Publish | 58,218 | 27.0% |
| IHave | 2,857 | 1.3% |

## Why nothing reacted

Both reaction paths existed and both were dead, and both trace to never calling `with_peer_score`:

- gossipsub's own `failed_message_slow_peer` penalty sits behind a `PeerScoreState::Active` match arm that `Disabled` never takes.
- The heartbeat's negative-score mesh prune reads scores from an empty map, so `unwrap_or_default()` hands it `0.0` and `0.0 < 0.0` is never true.
- Separately, gossipsub *does* raise `Event::SlowPeer`, but `handle_swarm_event` only matched `Event::Message`, so it fell into the catch-all.

## What this does

| | |
|---|---|
| Peer scoring | `with_peer_score` with lighthouse's parameters. Everything topology-independent is copied unchanged, including the slow-peer weights it deliberately sets 50x harsher than the libp2p defaults (`-10.0` against `-0.2`). Rates and decay horizons are re-derived for lean, which has no epochs and where every validator attests every slot. |
| `Event::SlowPeer` | Handled, logged, metered per queue. |
| No disconnect | Nothing closes a connection. gossipsub owns no connection-close path, and lighthouse scales its contribution so a fully graylisted peer lands one point short of its own disconnect threshold. |
| Flood publish | Off (was #588). |
| Metrics | `lean_gossip_peer_score` per peer, `lean_gossip_slow_peer_reports_total`, `lean_gossip_failed_messages_total{queue}`, `lean_gossip_publish_failed_total{reason}`. |

## A/B on a 6-node devnet

One binary in both arms, scoring toggled by env var, so the comparison isolates scoring alone. One node `SIGSTOP`ped and **held for 360s** so it never resumed, with `max_idle_timeout` raised to 900s so QUIC never tore the connection down. `PEERS` stayed at 5 in both arms, confirming no disconnect confounded either result.

| | scoring off | scoring on | |
|---|---:|---:|---|
| dropped `Forward` | 2441 | **20** | **-99.2%** |
| dropped `Publish` | 610 | 648 | flat |
| dropped `IHave` | 0 | 3613 | new |
| mesh peers | 5 | **4** | pruned |
| connected peers | 5 | 5 | never disconnected |
| WARN slope | linear ~9/s, 270s, no plateau | linear, no plateau | |

**The mesh prune is the mechanism and it works.** `Forward` targets only mesh members, so pruning removed 99.2% of the dropped payload — and `Forward` is the single largest category on devnet-5.

**Total WARN count goes up, not down.** A pruned peer becomes a lazy-gossip target and collects `IHave` every heartbeat. So raw `Send Queue full` count is the wrong success metric; the RPC-kind split is the right one. Reviewers watching only the log volume will think this made things worse.

**Dropped `Publish` stayed flat** because `filter_publish_candidates` tops the mesh up to `mesh_n` with random extras that are not re-checked against mesh membership. With 5 peers and `mesh_n` of 8 the mesh can never be full, so the pruned peer is re-selected on every publish. Networks with more peers than `mesh_n` should not see this — but that is an inference from the code, not something this devnet can show.

## Two claims I had wrong

**The publish gate does not end this flood.** `slow_peer_penalty` decays every `decay_interval`, so it settles rather than accumulating:

```
penalty_steady = rate * decay_interval / (1 - slow_peer_decay)
score_steady   = slow_peer_weight * penalty_steady   =  -44.4 * rate
```

Crossing `publish_threshold` needs a sustained **~180 failed sends/second**. devnet-5's ~33/s settles near -1500 and never approaches it. The unit test reaches 800 only because it publishes in a tight synchronous loop with no decay tick in between; that is an artifact of the test and is now documented as one.

**An earlier single-arm run was confounded.** With `max_idle_timeout` at 120s, the connection died at exactly t+120s, which is also when the flood stopped and when mesh went 5->4. I had read that as the score prune. The rerun above raises the timeout and samples `PEERS` every interval so a disconnect can never be misread as a prune again.

## Deviation worth reviewing

The `mesh_message_deliveries` terms (P3/P3b) are **explicitly zeroed**, and the second commit exists because they weren't. `TopicScoreParams::default()` ships `mesh_message_deliveries_weight: -1.0` with a threshold of 20 messages per **10ms** window. Not configuring those fields does not disable them, it enables them with parameters no node can satisfy. Lighthouse's `get_topic_params` has an `else` branch zeroing all eight for exactly this reason.

The healthy 6-node run caught it, and no unit test would have:

| | before fix | after fix |
|---|---|---|
| mesh peers | **0/5 on every node** | 5/5 |
| peer scores | **-20 to -67** | +29 to +37 |
| finality | advancing | advancing |

Finality kept advancing either way, because an empty mesh still gets topped up with random peers. The failure was invisible in head, justified and finalized, which is why there is now a test pinning those eight fields.

## What is still unverified

Reproducing the trigger needed three temporary overrides, none on this branch: `max_stream_data` 10MB -> 64KiB, `connection_handler_queue_len` 5000 -> 100, `max_idle_timeout` 10s -> 900s. Plain stalls produce nothing, because **QUIC's default 10MB per-stream window absorbs ~10,000 undelivered messages before back-pressure reaches the gossipsub queue at all.**

A 6-node/6-validator devnet offers well under 1 message/sec per peer, roughly three orders of magnitude below devnet-5. So the `Publish` behaviour at real scale, and whether a full 8-member mesh avoids the top-up, are both inferences.

Falsifiable prediction for a devnet-5 canary: dropped `Forward` should collapse on the affected node, `lean_gossip_mesh_peers` should drop by one, `lean_gossip_slow_peer_reports_total` should become non-zero, `IHave` drops should rise, and peer counts and finality should not move.

`make fmt`, `make lint` and the full workspace test suite pass.

### DIFF
--- a/bin/ethlambda/src/main.rs
+++ b/bin/ethlambda/src/main.rs
@@ -266,6 +266,7 @@ async fn main() -> eyre::Result<()> {
         bootnodes,
         listening_socket: p2p_socket,
         validator_ids,
+        network_validator_count: genesis_config.genesis_validators.len() as u64,
         attestation_committee_count,
         subscription_subnets: subscribed_subnets,
     })
--- a/crates/common/metrics/src/lib.rs
+++ b/crates/common/metrics/src/lib.rs
@@ -5,10 +5,10 @@ pub mod timing;
 
 // Re-export prometheus types and macros we use
 pub use prometheus::{
-    Encoder, Error as PrometheusError, Histogram, HistogramVec, IntCounter, IntCounterVec,
-    IntGauge, IntGaugeVec, TextEncoder, core::Collector, gather, register_histogram,
-    register_histogram_vec, register_int_counter, register_int_counter_vec, register_int_gauge,
-    register_int_gauge_vec,
+    Encoder, Error as PrometheusError, GaugeVec, Histogram, HistogramVec, IntCounter,
+    IntCounterVec, IntGauge, IntGaugeVec, TextEncoder, core::Collector, gather, register_gauge_vec,
+    register_histogram, register_histogram_vec, register_int_counter, register_int_counter_vec,
+    register_int_gauge, register_int_gauge_vec,
 };
 
 // Re-export commonly used items
--- a/crates/net/p2p/src/gossipsub/mod.rs
+++ b/crates/net/p2p/src/gossipsub/mod.rs
@@ -1,6 +1,7 @@
 mod encoding;
 mod handler;
 mod messages;
+pub mod scoring;
 
 pub use encoding::decompress_message;
 pub use handler::{
--- a/crates/net/p2p/src/gossipsub/scoring.rs
+++ b/crates/net/p2p/src/gossipsub/scoring.rs
@@ -1,0 +1,457 @@
+//! Gossipsub peer scoring parameters, ported from lighthouse.
+//!
+//! Without a call to `Behaviour::with_peer_score`, gossipsub keeps
+//! `PeerScoreState::Disabled`, and three of its defences are dead code:
+//! `below_threshold` returns `(false, 0.0)` for every peer so the publish and
+//! gossip gates pass everyone; the heartbeat's "prune peer with negative
+//! score" pass reads a score of 0.0 and never fires; and the
+//! `failed_message_slow_peer` penalty at the `Send Queue full` site takes the
+//! `PeerScoreState::Active` arm never. Enabling scoring is what turns a peer
+//! that cannot drain its send queue into a peer we stop feeding.
+//!
+//! Everything topology-independent is taken from lighthouse's
+//! `gossipsub_scoring_parameters.rs` unchanged, including the deliberately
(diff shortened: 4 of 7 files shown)


MegaRedHand:
Doesn't seem to improve things

PR #591, opened 2026-08-26
feat(cli): accept a `node` sub-command for running the node

pablodeymo:

The binary has only ever run the node, so an invocation is a bare list of node flags.
The offline block-building benchmark adds a second entry point, which means the node
first needs a name of its own.

`node` is an ordinary clap sub-command on a top-level parser that owns the binary's name,
version and about. `NodeOptions` (renamed from `CliOptions`) becomes a plain `clap::Args`
group and keeps every field
exactly as it is — no `Option<T>`, no `required = true`, no unwrap helper on the node
path, which is what the review of #497 objected to.

The flat `ethlambda --genesis ...` form keeps working, because that is what the
Dockerfile, lean-quickstart, the hive shim and the devnet skills all pass. clap has no
`default_subcommand`, so exactly one thing sits in front of the parser: a command line
that names no sub-command gets `node` inserted.

## What Changed

| File | Change |
|------|--------|
| `bin/ethlambda/src/command.rs` | New. Top-level `Cli` parser + `Command` sub-command enum, and `default_subcommand`, which inserts `node` unless the first token is a sub-command, `-h/--help/-V/--version`, or clap's generated `help` |
| `bin/ethlambda/src/cli.rs` | `clap::Parser` → `clap::Args`, and `CliOptions` renamed to `NodeOptions`; the `#[command(...)]` attribute moves to the top-level parser. No field changes |
| `bin/ethlambda/src/main.rs` | Parses through `command::parse()` and matches on `Command` |

`command.rs` also carries a test-only `parse_node_options` helper. Merging `main` brought
#579's `cli.rs` tests, which called `CliOptions::parse_from` — a `clap::Parser` method the
group lost when it became `clap::Args`. Git merged both sides cleanly, so nothing flagged
it; the test build was broken until `a3d7e52`, and both test modules now parse a node
command line through the real dispatch.

- **Every existing invocation keeps working**, and clap owns everything a reader should
  not have to trust us for: `--help` lists the sub-commands itself, usage lines name the
  sub-command, and an unknown sub-command produces clap's error rather than a
  stray-positional one.
- `NodeOptions` declares no positional arguments, so the first token after the program
  name is either a flag or a sub-command — a flag *value* never lands there and is never
  mistaken for one. A leading flag therefore means the flat node form.
- **`--version` after node flags still works and still prints the same string.** It used
  to live on the node options, so it was accepted anywhere; `propagate_version` keeps that,
  and `display_name = "ethlambda"` keeps the output byte-identical rather than
  `ethlambda-node`. All three forms are asserted equal.
- **One deliberate change:** a bare `ethlambda` now prints clap's top-level help, listing
  the sub-commands, instead of a missing-argument list. It still exits non-zero, and the
  test asserts both.

## Tests Added / Run

Unit tests in `command.rs` pin: the flat parse; the two forms agreeing field for field; a
`--node-id` value that is literally `node`; a trailing `node` token still rejected; a
second `node` token rejected; missing required flags in both forms; the bare invocation's
error kind and non-zero exit; `--help`/`--version` staying top-level; `--version` printing
one identical string across all three forms; and `--help` listing the sub-commands.

`make fmt`, `make lint`, `make test` (574 tests, 30 suites) — all clean.

## Related Issues / PRs

- Replaces the CLI approach reviewed in the now-closed #497
- Design doc in #594; the benchmark stacks on this in #595 → #596
- Related to #465

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -217,7 +217,6 @@ pub(crate) struct ShadowOptions {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use clap::Parser as _;
 
     /// The required flags, so a test can vary only what it cares about.
     fn parse(extra: &[&str]) -> CliOptions {
@@ -239,7 +238,7 @@ mod tests {
             "ethlambda_0",
         ];
         argv.extend_from_slice(extra);
-        CliOptions::parse_from(argv)
+        crate::command::parse_node_options(argv)
     }
 
     /// `--discovery.enable` on its own has to work: a default that is never
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -217,8 +217,12 @@ pub(crate) struct ShadowOptions {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::command::{Command, try_parse_from};
 
     /// The required flags, so a test can vary only what it cares about.
+    ///
+    /// `NodeOptions` is a `clap::Args` group rather than a parser of its own,
+    /// so this parses through the real dispatch, as the binary does.
     fn parse(extra: &[&str]) -> NodeOptions {
         let mut argv = vec![
             "ethlambda",
@@ -238,7 +242,8 @@ mod tests {
             "ethlambda_0",
         ];
         argv.extend_from_slice(extra);
-        crate::command::parse_node_options(argv)
+        let Command::Node(options) = try_parse_from(argv).expect("node options parse");
+        options
     }
 
     /// `--discovery.enable` on its own has to work: a default that is never
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -6,7 +6,13 @@ use std::path::PathBuf;
 use crate::version;
 
 #[derive(Debug, clap::Parser)]
-#[command(name = "ethlambda", author = "LambdaClass", version = version::CLIENT_VERSION, about = "ethlambda consensus client")]
+#[command(
+    name = "ethlambda",
+    author = "LambdaClass",
+    version = version::CLIENT_VERSION,
+    about = "ethlambda consensus client",
+    after_help = crate::command::HELP_NOTE
+)]
 pub(crate) struct CliOptions {
     /// Path to the chain genesis config (e.g., config.yaml).
     #[arg(long)]
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -3,16 +3,7 @@
 use std::net::IpAddr;
 use std::path::PathBuf;
 
-use crate::version;
-
-#[derive(Debug, clap::Parser)]
-#[command(
-    name = "ethlambda",
-    author = "LambdaClass",
-    version = version::CLIENT_VERSION,
-    about = "ethlambda consensus client",
-    after_help = crate::command::HELP_NOTE
-)]
+#[derive(Debug, clap::Args)]
 pub(crate) struct CliOptions {
     /// Path to the chain genesis config (e.g., config.yaml).
     #[arg(long)]
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -5,7 +5,7 @@ use std::net::IpAddr;
 use std::path::PathBuf;
 
 #[derive(Debug, clap::Args)]
-pub(crate) struct CliOptions {
+pub(crate) struct NodeOptions {
     /// Path to the chain genesis config (e.g., config.yaml).
     #[arg(long)]
     pub(crate) genesis: PathBuf,
@@ -159,7 +159,7 @@ pub(crate) struct DiscoveryConfig {
     pub(crate) target_peers: usize,
 }
 
-impl CliOptions {
+impl NodeOptions {
     /// Reject a discovery port that collides with the QUIC port.
     ///
     /// Both are UDP. Without this the collision surfaces at bind time as an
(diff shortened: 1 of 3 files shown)


bin/ethlambda/src/cli.rs
```diff
@@ -6,7 +6,13 @@ use std::path::PathBuf;
 use crate::version;
 
 #[derive(Debug, clap::Parser)]
-#[command(name = "ethlambda", author = "LambdaClass", version = version::CLIENT_VERSION, about = "ethlambda consensus client")]
+#[command(
+    name = "ethlambda",
+    author = "LambdaClass",
+    version = version::CLIENT_VERSION,
+    about = "ethlambda consensus client",
+    after_help = crate::command::HELP_NOTE
```
MegaRedHand: I don't like how we're essentially hacking our way around clap here. We should find a way to use clap's subcommands for this: https://docs.rs/clap/latest/clap/trait.Subcommand.html

MegaRedHand (approved):

MegaRedHand (changes requested):
.

bin/ethlambda/src/cli.rs
pablodeymo: You're right, and the argv surgery is gone. `node` is a real `Subcommand` now.

`CliOptions` became a plain `clap::Args` group and keeps every field exactly as it was — no `Option<T>`, no `required = true`, no unwrap on the node path — with the binary's name/version/about moved to a top-level parser holding `#[command(subcommand)]`. So clap owns the parts we were hand-rolling: `--help` lists the sub-commands itself (the hand-written `after_help` note is deleted), usage lines name the sub-command, and an unknown sub-command gets clap's own error instead of a stray-positional one.

The one thing I could not get from clap is a **default** sub-command. There is no `default_subcommand`, and the flat `ethlambda --genesis …` form has to keep working — the Dockerfile, lean-quickstart, the hive shim and the devnet skills all pass exactly that. So a single check remains in front of the parser: if the first token names no sub-command and is not `-h/--help/-V/--version` or clap's generated `help`, `node` is inserted. That is the whole of it, four lines in `default_subcommand`.

Two behaviour changes fell out of moving the parser identity up a level, both pinned by tests:

- `--version` used to sit on `CliOptions`, so it was accepted after node flags. `propagate_version` keeps that working, and `display_name = "ethlambda"` keeps the printed string byte-identical instead of `ethlambda-node`; a test asserts all three forms produce the same output.
- A bare `ethlambda` now prints clap's top-level help listing the sub-commands, rather than a missing-argument list for the node. Still exits non-zero.

Worth saying plainly: this is *longer* than the token stripping it replaces — a sub-command enum plus a top-level parser costs more lines than a `Vec::remove` — but the parts a reader has to trust are clap's now instead of ours.

If you would rather have no argv handling whatsoever, the alternative is dropping the flat form and requiring `ethlambda node …` everywhere. That is genuinely clean, but it breaks every running devnet and needs a lean-quickstart change, so I did not take it on unilaterally — say the word and I will.

Unrelated, so it does not mislead anyone reading down the thread: the Claude review bot above reports `after_help`/`HELP_NOTE` as missing from the diff. That was true of the commit it ran on, not of the branch head — its run predates a force-push. Moot either way now, since clap lists the sub-commands on its own and `HELP_NOTE` is deleted.

bin/ethlambda/src/command.rs
```diff
@@ -0,0 +1,260 @@
...
+)]
+struct Cli {
+    #[command(subcommand)]
+    command: Command,
+}
+
+/// What the command line asked the binary to do.
+#[derive(Debug, clap::Subcommand)]
+pub(crate) enum Command {
+    /// Run the consensus node (assumed when no sub-command is given).
```
MegaRedHand:
```suggestion
    /// Run the consensus node (default when no sub-command is given).
```

bin/ethlambda/src/cli.rs
```diff
@@ -4,10 +4,7 @@ use ethlambda_p2p::discovery::DEFAULT_DISCOVERY_TARGET_PEERS;
 use std::net::IpAddr;
 use std::path::PathBuf;
 
-use crate::version;
-
-#[derive(Debug, clap::Parser)]
-#[command(name = "ethlambda", author = "LambdaClass", version = version::CLIENT_VERSION, about = "ethlambda consensus client")]
+#[derive(Debug, clap::Args)]
 pub(crate) struct CliOptions {
```
MegaRedHand: We should rename this to something else, like `NodeOptions`

MegaRedHand (approved):

bin/ethlambda/src/command.rs
```diff
@@ -0,0 +1,272 @@
...
+/// tests that exercise it go through the real dispatch like everything else.
+#[cfg(test)]
+pub(crate) fn parse_node_options<I>(args: I) -> NodeOptions
+where
+    I: IntoIterator,
+    I::Item: Into<OsString>,
+{
+    let Command::Node(options) = try_parse_from(args).expect("node options parse");
+    options
+}
```
MegaRedHand: Can we inline this instead?

bin/ethlambda/src/command.rs
pablodeymo: Taken verbatim in 8784780.

bin/ethlambda/src/cli.rs
pablodeymo: Renamed to `NodeOptions` in 8784780 — it is the `node` sub-command's option group now, not the whole CLI. Rename only, no field or attribute changes. The PR description is updated to match.

bin/ethlambda/src/command.rs
pablodeymo: Inlined in 9385f6d. `try_parse_from` is now `pub(crate)` — it is already the testable core that `parse()` wraps — and both test modules destructure the `node` command inline, so no test-only function sits in the module surface. Flagging the one tradeoff: this widens `try_parse_from` from private to `pub(crate)`, since the `cli.rs` tests are in a different module. Happy to go back to a helper if you would rather keep it private.

MegaRedHand (approved):

PR #593, opened 2026-08-26
feat(cli): add offline block-building benchmark sub-command

pablodeymo:

Adds `ethlambda benchmark synthetic` — an offline harness that measures block building
**exactly as executed when the node proposes**, against a reproducible synthetic
workload, with no devnet required.

"Optimize block building" (#465) is the top roadmap item, but the only observability
today is Prometheus histograms on a live devnet: noisy, not reproducible, and unable to
compare an optimization against a baseline. This is milestone M1 (mock-crypto mode) of
the design in `docs/plans/block-building-benchmark.md`; M2 adds real XMSS/leanVM pools
and post-build seal-phase measurement, M3 adds replay-from-datadir.

> **Supersedes #497.** Same harness, but the CLI plumbing it needed is now split out:
> the `node`/`benchmark` token dispatch is #591, which this PR stacks on, and the
> non-determinism this harness uncovered in `extend_proofs_greedily` merged separately
> as #590. What is left here is the benchmark itself. **Review #591 first**; the base
> moves to `main` once it lands.

```
$ make bench
  ...
  compact               10    0.000ms    0.000ms    0.000ms    0.000ms    0.000ms
  select_payloads       10    0.001ms    0.001ms    0.001ms    0.001ms    0.001ms
  stf_simulate          10    0.011ms    0.011ms    0.011ms    0.011ms    0.011ms
  overhead              10    0.047ms    0.048ms    0.048ms    0.049ms    0.049ms
  wall                  10    0.060ms    0.061ms    0.061ms    0.062ms    0.062ms
```

## What Changed

| File | Change |
|------|--------|
| `bin/ethlambda/src/benchmark/{mod,corpus,report}.rs` | New harness: seeded synthetic chain over `InMemoryBackend`, per-slot pool seeding into the pending pool (promoted by the proposal tick, as in production), iteration loop driving `produce_block_with_signatures` + block import; stats + human/JSON report |
| `bin/ethlambda/src/command.rs` | `benchmark` joins `node` as a dispatched token, parsed by a `clap::Parser` of its own so harness arguments never enter `CliOptions`; argv[0] is rewritten to `ethlambda benchmark` so usage lines name the sub-command that owns them |
| `bin/ethlambda/src/main.rs` | `main` is synchronous: it dispatches, and only the node path enters the tokio runtime (`run_node` carries the `#[tokio::main]` attributes). Benchmark logs go to stderr at WARN so the report owns stdout |
| `crates/storage/{lib,store}.rs` | Export `NEW_PAYLOAD_CAP` so the harness rejects `--proofs-per-data` batches the pending pool would evict whole |
| `bin/ethlambda/build.rs` | Embed the resolved leanSig and leanVM revisions from `Cargo.lock` into reports (leansig tracks the moving `devnet4` branch; leanVM does the aggregation, so its pinned rev moves the measured crypto too). The per-`[[package]]` parse collects `name` and `source` before extracting the rev, so it does not depend on TOML field order |
| `Makefile`, `.github/workflows/ci.yml` | `make bench`; seconds-fast mock smoke step in the Test job validating the JSON contract |
| `docs/plans/block-building-benchmark.md` | Design doc and milestone roadmap |

- **`cli.rs` is not touched, and the node runtime is unchanged.** The seven node-required
  arguments stay plain `PathBuf`/`String`, so clap keeps emitting its own
  missing-argument errors and there is no `Option<T>` to unwrap on the node path. That
  replaces #497's `subcommand_negates_reqs` approach, which needed both.
- **Determinism:** same seed + params → identical per-iteration block roots (recorded in
  the JSON, so a baseline-vs-optimized diff proves an optimization changed only speed,
  not attestation selection). Verified across repeated runs. The harness never reads the
  wall clock into results.
- **Exact phase attribution with zero hot-path changes:** per-iteration
  `select_payloads`/`compact`/`stf_simulate` come from the sample sums of the existing
  `lean_block_proposal_attestation_build_phase_seconds` histogram, deltaed between
  iterations; `overhead` is the clamped remainder of wall minus the phases.
- The benchmark never starts the tokio runtime, so it cannot park a worker thread for the
  duration of a CPU-bound run.

## Tests Added / Run

- `command.rs` unit tests cover the benchmark token: it parses with no node argument, it
  rejects node flags, and its usage line names the sub-command.
- Verified by hand on this branch: `make bench`; the CI smoke assertion
  (`jq -e '.schema_version == 1 and (.samples | length == 3)'`); identical block-root
  sequences across two runs at the same seed; and `ethlambda --genesis config.yaml` still
  failing with clap's own missing-argument list.
- `make fmt`, `make lint`, `make test` (582 tests, 30 suites) — all clean.

## Related Issues / PRs

- Supersedes #497
- Stacked on #591
- Related to #465, #590

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean


pablodeymo:
Split into three smaller PRs, each with a goal of its own:

- **#594** — the design plan, on its own so the measurement boundary, the phase-attribution approach and the milestones can be argued before there is code to argue around. Targets `main` directly, so it does not deepen the stack.
- **#595** — the harness: the real proposer path driven deterministically, reporting one row per measured iteration. Stacked on #591.
- **#596** — comparable reports: aggregate statistics, the leansig/leanVM provenance that says whether two reports may be compared at all, and JSON + the CI contract. Additive on top of #595.

#595 + #596 reconstruct what was reviewed here, plus a per-iteration table (#595 would otherwise collect the phase breakdown without ever showing it). Each of the three was verified independently: `make fmt`, `make lint`, `make test`, and the benchmark run end-to-end.

Closing in favour of that stack.

PR #594, opened 2026-08-26
docs: document the block-building benchmark

pablodeymo:

Documents the block-building benchmark: what it measures, how to run it, how to read a
report, and what it cannot do yet.

This replaces the design plan this PR originally carried. Per review, a `docs/plans/`
file written as "this is how the benchmark was originally designed" goes stale the moment
the benchmark changes and then actively misleads, so the page now describes the tool
rather than a schedule for building it. The plan served its purpose — it was the shared
reference while #595 and #596 were reviewed — and is not something the tree should keep.

> **Restacked.** Now based on #596, so the page documents behaviour that actually exists
> rather than behaviour that is still in review. Base moves to `main` once the two land.

## What Changed

| File | Change |
|------|--------|
| `docs/benchmarking.md` | New. Running it (flag table with defaults), what the measured span includes and deliberately excludes, how phase times come from the existing histogram, reading the per-iteration and summary tables, comparing two runs, current limitations, the CI smoke step |
| `docs/SUMMARY.md` | Listed under Development |
| `docs/plans/block-building-benchmark.md` | Removed |
| `bin/ethlambda/src/benchmark/mod.rs` | Module doc points at the new page |

Documentation only — the one code change is a doc-comment path.

Two things the plan file never stated, both of which a reader needs:

- **When two reports may not be compared at all.** The header carries the resolved
  leanSig and leanVM revisions plus the machine fields; leanSig tracks a moving branch
  and leanVM performs the aggregation, so either one moving changes the measured crypto.
- **What is not supported yet** — real crypto, the seal phase, replay from a datadir —
  written as current limitations rather than as milestones, so the page does not promise
  a schedule it cannot keep.

## Tests Added / Run

- `make docs` builds the site with the new page in place; no dangling references to the
  removed plan file anywhere in the tree.
- `make fmt`, `make lint`, `make test` (622 tests) — all clean.

## Related Issues / PRs

- Stacked on #596, which stacks on #595
- Related to #465

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/docs/benchmarking.md
+++ b/docs/benchmarking.md
@@ -6,8 +6,7 @@ it proposes, against a reproducible synthetic workload, with no devnet running.
 Block building is otherwise only observable through the Prometheus histograms a
 live node exports. Those are noisy, depend on whatever the network happened to
 be doing, and cannot be diffed against a baseline — which makes them a poor
-instrument for the work tracked in
-[#465](https://github.com/lambdaclass/ethlambda/issues/465). The benchmark
+instrument for tracking performance. The benchmark
 trades network realism for repeatability: the same parameters produce the same
 blocks every run, so two reports differ only where the code differs.
 
--- a/docs/benchmarking.md
+++ b/docs/benchmarking.md
@@ -1,0 +1,138 @@
+# Benchmarking block building
+
+`ethlambda benchmark` measures block building the way the node performs it when
+it proposes, against a reproducible synthetic workload, with no devnet running.
+
+Block building is otherwise only observable through the Prometheus histograms a
+live node exports. Those are noisy, depend on whatever the network happened to
+be doing, and cannot be diffed against a baseline — which makes them a poor
+instrument for the work tracked in
+[#465](https://github.com/lambdaclass/ethlambda/issues/465). The benchmark
+trades network realism for repeatability: the same parameters produce the same
+blocks every run, so two reports differ only where the code differs.
+
+## Running it
+
+```bash
+make bench                                  # defaults, mock crypto
+BENCH_ARGS="synthetic --iterations 50" make bench
+```
+
+`make bench` is a thin wrapper. The binary takes the same arguments directly:
+
+```bash
+ethlambda benchmark synthetic --mock-crypto --num-validators 8 --iterations 10
+```
+
+A default mock run finishes in well under a second, which is why CI can afford
+to run one on every pull request.
+
+| Flag | Default | Meaning |
+| --- | --- | --- |
+| `--num-validators` | `8` | Validators in the synthetic genesis |
+| `--warmup-slots` | `8` | Unmeasured slots built first, so measured builds run on a state with realistic historical roots and justifications |
+| `--iterations` | `10` | Measured builds, one block each |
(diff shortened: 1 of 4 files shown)


docs/plans/block-building-benchmark.md
```diff
@@ -0,0 +1,159 @@
+# Plan: `ethlambda benchmark` — offline block-building benchmark sub-command
+
+## Context
+
+The README roadmap lists **"Optimize block building" (issue #465)** as the top near-term
+priority, but block building is only observable today through Prometheus histograms on a
+live devnet — there is no reproducible, offline way to measure it or to compare an
+optimization against a baseline. This adds an `ethlambda benchmark` sub-command that
```
MegaRedHand: This will get stale soon

docs/plans/block-building-benchmark.md
MegaRedHand, line 1: We should rewrite this as benchmark documentation instead of "this is how benchmarks were originally designed". This is useful for reviewing the other PRs, but will end up being misleading when we start modifying the benchmarks.

pablodeymo: Rewritten. `docs/plans/block-building-benchmark.md` is deleted; `docs/benchmarking.md` documents the benchmark in the present tense and is listed in SUMMARY.md under Development, so it sits with the other operational pages instead of as a design record.

I also restacked this PR onto #596, so the page describes behaviour that exists rather than behaviour still in review, and squashed it to one commit — otherwise the diff would have read "add plan doc, delete plan doc".

Two things the plan file never said and a reader needs: when two reports may not be compared at all (the resolved leanSig and leanVM revisions, plus the machine fields), and what is not supported yet — real crypto, the seal phase, replay from a datadir — written as current limitations rather than as milestones, so the page does not promise a schedule.

docs/plans/block-building-benchmark.md
pablodeymo: That was the right call, and it is why the plan file is deleted rather than edited: nothing in the tree now records how the benchmark was originally designed, only what it currently does. The milestone table went with it — the parts that were still true are in `docs/benchmarking.md` as current limitations.

docs/benchmarking.md
```diff
@@ -0,0 +1,138 @@
+# Benchmarking block building
+
+`ethlambda benchmark` measures block building the way the node performs it when
+it proposes, against a reproducible synthetic workload, with no devnet running.
+
+Block building is otherwise only observable through the Prometheus histograms a
+live node exports. Those are noisy, depend on whatever the network happened to
+be doing, and cannot be diffed against a baseline — which makes them a poor
+instrument for the work tracked in
+[#465](https://github.com/lambdaclass/ethlambda/issues/465). The benchmark
```
MegaRedHand:
```suggestion
Block building is otherwise only observable through the Prometheus histograms a
live node exports. Those are noisy, depend on whatever the network happened to
be doing, and cannot be diffed against a baseline — which makes them a poor
instrument for tracking performance. The benchmark
```

MegaRedHand (approved):

PR #595, opened 2026-08-26
feat(cli): add offline block-building benchmark sub-command

pablodeymo:

Adds `ethlambda benchmark synthetic` — an offline harness that measures block building
**exactly as executed when the node proposes**, against a reproducible synthetic
workload, with no devnet required.

Second of three (design doc → **this** → comparable reports). What lands here is the
smallest thing that runs: the real proposer path, driven deterministically, reporting one
row per measured iteration. Aggregate statistics, build provenance and machine-readable
output follow in the next PR, so this one can be reviewed for *what it measures* rather
than how it formats.

> **Stacked on #591**, which adds the `node`/`benchmark` token dispatch this uses. Review
> that first; the base moves to `main` once it lands.

```
$ make bench
  enable_proposer_aggregation=false max_attestations_per_block=3
  ethlambda/v0.1.0/aarch64-apple-darwin/rustc-v1.97.1 os=macos arch=aarch64 threads=14

  iter           compact  select_payloads     stf_simulate   overhead       wall         root
  1              0.000ms          0.002ms          0.015ms    0.068ms    0.085ms   0x7282cc99
  2              0.000ms          0.002ms          0.015ms    0.066ms    0.083ms   0xb9065af0
  3              0.000ms          0.002ms          0.015ms    0.064ms    0.081ms   0x303f6b0f
```

## How to read this

In this order — each piece is understandable without the next:

1. **`corpus.rs`** — the workload. Deterministic validators, a genesis store over
   `InMemoryBackend`, and `seed_pool`, which fills the pending pool for one slot and
   reports how many entries the next build will see.
2. **`build_one_slot`** (`mod.rs`) — one slot end to end: seed, time the build, import the
   block. Warmup and measured slots run this same path; only whether the sample is kept
   differs, so there is no "am I warming up?" branching inside.
3. **`PhaseTimer`** (`mod.rs`) — `start()` before the build, `finish()` after. Two
   readings of the existing phase histogram; the difference between their sample *sums* is
   the build's phase time, so nothing is added to the hot path.
4. **`run_synthetic`** (`mod.rs`) — validate, set up, loop over slots, report. 46 lines.
5. **`report.rs`** — the types and the per-iteration table.

## What Changed

| File | Change |
|------|--------|
| `bin/ethlambda/src/benchmark/mod.rs` | Harness driver: clap options + validation, `run_synthetic`'s slot loop, `build_one_slot` for one slot's work, and `PhaseTimer` for per-phase attribution |
| `bin/ethlambda/src/benchmark/corpus.rs` | Seeded synthetic corpus: genesis store over `InMemoryBackend`, deterministic pubkeys via splitmix64, and per-slot pool seeding in fixed insertion order, which also rejects a batch the pool would evict whole |
| `bin/ethlambda/src/benchmark/report.rs` | Params/Environment/Sample types and the human-readable per-iteration table |
| `bin/ethlambda/src/command.rs` | `benchmark` joins `node` as a second clap sub-command, so clap lists it in `--help` and names it in its own usage lines. The node payload becomes `Box<CliOptions>` now that a much smaller variant sits beside it |
| `bin/ethlambda/src/main.rs` | `main` becomes synchronous and dispatches; only the node path enters the tokio runtime (`run_node` carries the `#[tokio::main]` attributes). Benchmark logs go to stderr at WARN so the report owns stdout |
| `crates/storage/{lib,store}.rs` | Export `NEW_PAYLOAD_CAP` so the harness rejects a `--proofs-per-data` batch the pending pool would evict whole |
| `Makefile` | `make bench` (override `BENCH_ARGS` to customize) |

- **`cli.rs` is not touched by this PR and the node runtime is unchanged.** The harness
  arguments live in their own `Args` group; the node's stay plain `PathBuf`/`String`, so
  clap keeps emitting its own missing-argument errors.
- **It measures the production path**, not a copy: the harness enters through
  `produce_block_with_signatures`, the same function `BlockChainServer::propose_block`
  calls, and seeds the *pending* pool so the proposal tick promotes it exactly as on a
  live node.
- **Determinism:** same seed + params → identical per-iteration block roots. Verified
  across repeated runs; the roots are printed so a baseline-vs-optimized diff proves an
  optimization changed only speed, not attestation selection. The harness never reads the
  wall clock into results.
- **Exact phase attribution with zero hot-path changes:** per-iteration
  `select_payloads`/`compact`/`stf_simulate` come from the sample sums of the existing
  `lean_block_proposal_attestation_build_phase_seconds` histogram, deltaed between
  iterations, with a per-phase assertion that the count advanced by exactly one.
  `overhead` is the clamped remainder of wall minus the phases.
- The benchmark never starts the tokio runtime, so it cannot park a worker thread for the
  duration of a CPU-bound run.

## Tests Added / Run

- `corpus.rs`: participant groups partition every validator; synthetic pubkeys are
  deterministic for a seed.
- `command.rs`: the benchmark token parses with no node argument, rejects node flags, and
  its usage line names the sub-command.
- Verified by hand: `make bench`; identical block-root sequences across two runs at the
  same seed; `ethlambda --genesis config.yaml` still failing with clap's own
  missing-argument list.
- `make fmt`, `make lint`, `make test` (576 tests, 30 suites) — all clean.

## Related Issues / PRs

- Stacked on #591; design doc in the accompanying docs PR
- Followed by the comparable-reports PR
- Splits the now-closed #497 / #593
- Related to #465

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/bin/ethlambda/src/benchmark/mod.rs
+++ b/bin/ethlambda/src/benchmark/mod.rs
@@ -1,0 +1,313 @@
+//! Offline block-building benchmark (`ethlambda benchmark`).
+//!
+//! Drives the exact production proposer path — `produce_block_with_signatures`,
+//! the same entry `BlockChainServer::propose_block` uses — against a synthetic
+//! in-memory chain, and reports per-phase timing distributions. Gossip publish
+//! and the slot-alignment sleep are outside the measured span, matching the
+//! node's own `lean_block_building_time_seconds` boundary.
+//!
+//! See docs/plans/block-building-benchmark.md for the design and roadmap
+//! (real-crypto pools and replay-from-datadir land in later milestones).
+
+mod corpus;
+mod report;
+
+use std::collections::{BTreeMap, HashMap};
+use std::time::Instant;
+
+use ethlambda_blockchain::block_builder::ProposerConfig;
+use ethlambda_blockchain::metrics::BLOCK_PROPOSAL_ATTESTATION_BUILD_PHASES;
+use ethlambda_blockchain::store::{on_block_without_verification, produce_block_with_signatures};
+use ethlambda_storage::{NEW_PAYLOAD_CAP, Store};
+use ethlambda_types::block::{MultiMessageAggregate, SignedBlock};
+use ethlambda_types::primitives::HashTreeRoot as _;
+use eyre::WrapErr as _;
+
+use report::{Environment, Params, Report, Sample};
+
+#[derive(Debug, clap::Args)]
+pub(crate) struct BenchmarkOptions {
+    #[command(subcommand)]
+    workload: Workload,
+}
+
+#[derive(Debug, clap::Subcommand)]
+enum Workload {
+    /// Benchmark block building on a synthetic in-memory chain.
+    Synthetic(SyntheticOptions),
+}
+
+#[derive(Debug, clap::Args)]
+struct SyntheticOptions {
+    /// Number of validators in the synthetic genesis.
+    #[arg(long, default_value = "8", value_parser = clap::value_parser!(u64).range(1..=4096))]
+    num_validators: u64,
+    /// Unmeasured chain-advancement slots before measuring. Builds and imports
+    /// one block per slot so the measured builds run on a state with
+    /// representative historical roots and justifications, and warms the state
+    /// cache.
+    #[arg(long, default_value = "8")]
+    warmup_slots: u64,
+    /// Aggregate proofs seeded per AttestationData, mimicking committee
+    /// aggregators covering disjoint validator subsets. The default of 1 (one
+    /// full-coverage proof per data) keeps justification/finalization
+    /// advancing every slot. Higher values exercise multi-proof selection and
+    /// same-data collapse, but without --enable-proposer-aggregation the block
+    /// then carries only the best partial proof (< 2/3 coverage), so
+    /// justification stalls — the real coverage cost of disabling proposer
+    /// aggregation.
+    #[arg(long, default_value = "1", value_parser = clap::value_parser!(u64).range(1..))]
+    proofs_per_data: u64,
+    /// Deterministic seed for the synthetic validator set. Two runs with the
+    /// same seed and parameters produce identical per-iteration block roots.
+    #[arg(long, default_value = "42")]
+    seed: u64,
+    #[command(flatten)]
+    common: CommonOptions,
+}
+
+impl SyntheticOptions {
+    fn validate(&self) -> eyre::Result<()> {
+        eyre::ensure!(
+            self.common.mock_crypto,
+            "real-crypto benchmarking is not implemented yet; rerun with --mock-crypto"
+        );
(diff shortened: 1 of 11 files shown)


bin/ethlambda/src/benchmark/mod.rs
```diff
@@ -0,0 +1,313 @@
...
+        wall_seconds,
+        phases,
+        overhead_seconds,
+        attestations_packed,
+        aggregates: aggregates_count,
+        pool_entries,
+    })
+}
+
+const PHASE_HISTOGRAM: &str = "lean_block_proposal_attestation_build_phase_seconds";
```
MegaRedHand, line 251: This is a really hacky way of recording phase times. We should record the duration of each phase inside the relevant functions and return a report, which the metrics consume. That way we don't have to do this here.

We can do this in another PR

bin/ethlambda/src/command.rs
```diff
@@ -50,8 +52,18 @@ pub(crate) enum Command {
...
+    // (~312 bytes against ~80), which `clippy::large_enum_variant` rightly
+    // flags: one allocation per process is cheaper than carrying that size in
+    // every value of this enum.
     #[command(display_name = "ethlambda")]
-    Node(NodeOptions),
+    // Boxed because the node options dwarf the other variant's payload (~312
+    // bytes against ~80), which `clippy::large_enum_variant` rightly flags: one
+    // allocation per process is cheaper than carrying that size in every value
+    // of this enum.
+    Node(Box<NodeOptions>),
```
MegaRedHand: The comment is duplicate.
Also, I think we should drop the `Box` and tag this enum with `#[allow(clippy::large_enum_variant)]`. Adding the Box on each variant just adds noise, and having the enum be unbalanced isn't a performance problem here really.

MegaRedHand (approved):
LGTM

bin/ethlambda/src/command.rs
pablodeymo: Done in 39aa610: the `Box` is gone and the enum carries `#[allow(clippy::large_enum_variant)]` with a line on why the imbalance is fine here — exactly one variant is built per process and `main` consumes it immediately. Same reasoning `req_resp::messages` already uses, so it is at least consistent with the rest of the tree.

You were right about the duplicate too: the previous rebase left that comment in the file twice. Both copies are gone with the `Box`.

bin/ethlambda/src/benchmark/mod.rs
pablodeymo: Agreed, and filed as #599 so it does not get lost.

Two reasons beyond removing the indirection, which I wrote up there: having the phases return their own timings makes them available to anything that builds a block, rather than only to a process that can read the global registry — which is what a replay-from-datadir mode would want — and it removes the reason the harness is limited to one configuration per process invocation. It also lets the "observed exactly once per build" assertion go away, since that check only exists because a registry diff cannot tell a mis-attribution from a real measurement.

PR #596, opened 2026-08-26
feat(cli): make benchmark reports comparable across runs and machines

pablodeymo:

Per-iteration rows show what one build cost. Comparing an optimization against a baseline
needs three more things, and this adds them: aggregate statistics, build provenance, and
machine-readable output.

Third of three (design doc → harness → **this**).

> **Stacked on the harness PR.** Its diff here is additive — the per-iteration rows stay,
> the summary is appended below them.

```
  iter           compact  select_payloads     stf_simulate   overhead       wall         root
  1              0.000ms          0.002ms          0.015ms    0.068ms    0.085ms   0x7282cc99
  2              0.000ms          0.002ms          0.015ms    0.066ms    0.083ms   0xb9065af0
  3              0.000ms          0.002ms          0.015ms    0.064ms    0.081ms   0x303f6b0f

  compact                3    0.000ms    0.000ms    0.000ms    0.000ms    0.000ms
  select_payloads        3    0.002ms    0.002ms    0.002ms    0.002ms    0.002ms
  stf_simulate           3    0.015ms    0.015ms    0.015ms    0.015ms    0.015ms
  overhead               3    0.064ms    0.066ms    0.066ms    0.068ms    0.068ms
  wall                   3    0.081ms    0.083ms    0.083ms    0.085ms    0.085ms
```

## What Changed

| File | Change |
|------|--------|
| `bin/ethlambda/src/benchmark/report.rs` | `Stats`/`Summary` plus `stats()`, `percentile()` and the aggregate table: count, min, mean, p50, p90, max per phase, and a CV flagged above 10%. `schema_version` + `to_json()`. `Environment` gains the two resolved crypto revisions |
| `bin/ethlambda/build.rs` | Resolve the leansig and leanVM revisions from `Cargo.lock` into `rustc-env` vars. The per-`[[package]]` parse collects `name` and `source` before extracting the rev, so it does not depend on TOML field order |
| `bin/ethlambda/src/benchmark/mod.rs` | `--format human\|json` and `--output <path>` |
| `bin/ethlambda/Cargo.toml`, `Cargo.lock` | `serde_json` |
| `.github/workflows/ci.yml` | Seconds-fast mock smoke step in the Test job asserting the JSON contract |

- **Nearest-rank percentiles, no interpolation.** Sample counts are small, so an exact
  observed value beats a blend of two neighbours.
- **Outliers are never discarded** and the raw per-iteration rows stay above the summary,
  so a heavy tail stays visible instead of being averaged away. A CV above 10% is flagged
  so a noisy run is not read as a result.
- **Provenance is a comparability guard, not decoration.** leansig is pinned to a moving
  branch and leanVM does the signature aggregation, so either revision moving moves the
  measured crypto. Two reports that disagree on them are not comparable, and without this
  the report cannot say so.
- **The JSON shape is pinned by CI**, so a change to the report contract cannot land
  unnoticed. Logs already go to stderr, so the JSON pipes straight into `jq`.
- Node behavior is untouched; `build.rs` only adds env vars consumed by the report.

## Tests Added / Run

- `report.rs`: percentile on a single sample and on odd/even lengths; `stats()` against a
  known set whose population stddev gives CV = 0.4; `stats()` on empty input is zeroed.
- Verified by hand: the CI assertion
  `jq -e '.schema_version == 1 and (.samples | length == 3)'` passes, and reports carry
  both resolved revisions.
- `make fmt`, `make lint`, `make test` (580 tests, 30 suites) — all clean.

## Related Issues / PRs

- Stacked on the harness PR; design doc in the accompanying docs PR
- Splits the now-closed #497 / #593
- Related to #465

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,6 +80,6 @@ jobs:
       # harness end-to-end and its JSON output contract in a few seconds.
       - name: Benchmark smoke (mock crypto)
         run: |
-          cargo run --release --bin ethlambda -- benchmark synthetic --mock-crypto \
+          cargo run --profile release-fast --bin ethlambda -- benchmark synthetic --mock-crypto \
             --num-validators 4 --warmup-slots 4 --iterations 3 --format json \
             | jq -e '.schema_version == 1 and (.samples | length == 3)'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,3 +75,11 @@ jobs:
 
       - name: Run fixture-based tests
         uses: ./.github/actions/run-fixture-tests
+
+      # Reuses the release build from the test step; validates the benchmark
+      # harness end-to-end and its JSON output contract in a few seconds.
+      - name: Benchmark smoke (mock crypto)
+        run: |
+          cargo run --release --bin ethlambda -- benchmark synthetic --mock-crypto \
+            --num-validators 4 --warmup-slots 4 --iterations 3 --format json \
+            | jq -e '.schema_version == 1 and (.samples | length == 3)'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1932,6 +1932,7 @@ dependencies = [
  "libssz-types",
  "reqwest",
  "serde",
+ "serde_json",
  "serde_yaml_ng",
  "thiserror 2.0.18",
  "tikv-jemallocator",
--- a/bin/ethlambda/Cargo.toml
+++ b/bin/ethlambda/Cargo.toml
@@ -38,6 +38,7 @@ tracing.workspace = true
 tracing-subscriber = "0.3"
 
 serde.workspace = true
+serde_json.workspace = true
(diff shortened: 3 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -75,3 +75,11 @@ jobs:
 
       - name: Run fixture-based tests
         uses: ./.github/actions/run-fixture-tests
+
+      # Reuses the release build from the test step; validates the benchmark
+      # harness end-to-end and its JSON output contract in a few seconds.
+      - name: Benchmark smoke (mock crypto)
+        run: |
+          cargo run --release --bin ethlambda -- benchmark synthetic --mock-crypto \
```
MegaRedHand: We use the `release-fast` profile for tests

```suggestion
          cargo run --profile release-fast --bin ethlambda -- benchmark synthetic --mock-crypto \
```

MegaRedHand (approved):

PR #597, opened 2026-08-28
fix: align Lean justifiability overflow semantics

ss1738:
## Context

While extending the existing justifiability formalization, I found a
fixed-width arithmetic boundary outside the current proof domain.

For `delta = 2^62 + 2`, the previous Lean implementation returned `true`
because `UInt64` wrapping reduced `4 * delta + 1` to `9`, satisfying the
odd-square check. The corresponding mathematical delta is neither square nor
pronic.

## Change

The Lean implementation now evaluates the pronic branch only when
`4 * delta + 1` fits in `UInt64`, matching the checked-arithmetic structure in
the native Rust source.

Regression proofs cover values around the overflow boundary.

## Proof contract

`justifiable_equiv` remains explicitly bounded by `d.toNat < 2^62`.
This PR does not strengthen that theorem to a total equivalence result.

The Rust-side documentation has also been updated to state the actual proof
domain rather than claiming correctness for all natural numbers.

## Testing

- `cd formal && lake build EthLambdaProofs`
- regression cases for `2^62 - 1`, `2^62`, `2^62 + 2`, and `UInt64::MAX`

### DIFF
--- a/formal/EthLambda/Justifiability.lean
+++ b/formal/EthLambda/Justifiability.lean
@@ -33,12 +33,16 @@ def isqrt (n : UInt64) : UInt64 :=
     -- r+1 ≤ n/(r+1)  ↔  (r+1)*(r+1) ≤ n  (when r+1 > 0)
     if r + 1 <= n / (r + 1) then r + 1 else r
 
-/-- Computable justifiability check mirroring the Rust implementation. -/
+/-- Computable justifiability check mirroring Rust's checked arithmetic.
+    The pronic discriminant is evaluated only when `4 * delta + 1` fits in
+    `UInt64`; otherwise that branch returns `false`. -/
 def justifiable (delta : UInt64) : Bool :=
   delta <= 5
     || isqrt delta ^ 2 == delta
-    || (let val := 4 * delta + 1
-        isqrt val ^ 2 == val && val % 2 == 1)
+    || if delta <= 4611686018427387903 then
+         let val := 4 * delta + 1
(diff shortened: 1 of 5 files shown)


formal/EthLambda/Justifiability.lean
```diff
@@ -33,12 +33,16 @@ def isqrt (n : UInt64) : UInt64 :=
...
-/-- Computable justifiability check mirroring the Rust implementation. -/
+/-- Computable justifiability check mirroring Rust's checked arithmetic.
+    The pronic discriminant is evaluated only when `4 * delta + 1` fits in
+    `UInt64`; otherwise that branch returns `false`. -/
 def justifiable (delta : UInt64) : Bool :=
   delta <= 5
     || isqrt delta ^ 2 == delta
-    || (let val := 4 * delta + 1
-        isqrt val ^ 2 == val && val % 2 == 1)
+    || if delta <= 4611686018427387903 then
```
MegaRedHand, line 42: We should make this work correctly with all values if possible, otherwise just leaving it at < 2**62 like before is enough. Making this return a fixed wrong result, true or false, isn't really a fix here.

MegaRedHand:
We're not currently accepting PRs to the `lean-formalization` branch. As I mentioned in Telegram, the Lean4 formalization is expected to land on the leanSpec repo, so the PR should go there. Feel free to use the code in `lean-formalization` as the base, however, or ask us for feedback.

PR #598, opened 2026-08-31
feat(config): read the slot duration from the network config file

MegaRedHand:
## What

`config.yaml` gains an optional `MILLISECONDS_PER_SLOT`. It defaults to `4000`, the value the whole network already ran on, so an existing config file behaves exactly as before.

```yaml
GENESIS_TIME: 1770407233
MILLISECONDS_PER_SLOT: 8000   # optional, defaults to 4000
GENESIS_VALIDATORS:
  - attestation_pubkey: "cd32..."
    proposal_pubkey: "b7b0..."
```

The value must be a positive multiple of `INTERVALS_PER_SLOT`, rejected at parse time otherwise: the five intervals are cut from it and every duty is scheduled off those boundaries, so a duration that does not divide evenly would leave the last interval short and drift the grid against the wall clock.

`INTERVALS_PER_SLOT` stays a compile-time constant. Each interval carries a distinct validator duty, so the count is part of the protocol rather than a tuning knob.

## Why not put it in the state's `Config`

`State.config` is SSZ-merkleized into the state root, so its layout is fixed by the spec and cannot gain a field without changing every state root and breaking the SSZ fixtures.

So there are now two config types:

| Type | Where it lives | Holds |
| --- | --- | --- |
| `StateConfig` | `State.config`, merkleized | `genesis_time`, unchanged spec layout (this is the old `ChainConfig`, renamed) |
| `ChainConfig` | `Metadata["config"]` in the DB | `genesis_time` + `milliseconds_per_slot` |

Consumers read the second one through `Store::config`, which already carried the genesis time, so most call sites changed from a constant to a field on something they already had in hand.

## What now follows the configured cadence

- Interval length, the tick grid, and every slot/interval conversion (`SlotInterval`, `ms_until_next_interval`, `store::on_tick`).
- The aggregation deadline, which was a hardcoded 800 ms whose doc already described it as "one full interval".
- The early-aggregation window, now clamped to one interval rather than guarded by a `const` assert. It is subtracted from an interval offset in two places, and a short enough cadence would make the nominal 600 ms wider than an interval and underflow both.
- Gossipsub's duplicate cache, which leanSpec defines as `SECONDS_PER_SLOT * JUSTIFICATION_LOOKBACK_SLOTS * 2` and which was pinned at `4 * 3 * 2` seconds.
- `GET /lean/v0/config/spec`, which now reports what the node is actually running rather than what it was built with.

Slot-count constants (`SNAPSHOT_ANCHOR_INTERVAL`, `MAX_RESUMABLE_DB_STATE_AGE`, the payload buffer caps) are deliberately left alone: their cost is per-slot, not per-second. Their doc comments now say which wall-clock figure assumes the default cadence.

Prometheus histogram buckets are also left alone, since Prometheus fixes them at registration. A network on a different cadence reads the arrival and tick histograms against the default grid; noted in `docs/metrics.md` and in the bucket definitions.

## Compatibility

- **Existing data directories resume.** A DB written before this change holds a bare SSZ `StateConfig` under `Metadata["config"]`. It decodes as the 4-second default, which is the only cadence it could have run under.
- **A cadence change is refused, not silently applied.** Resuming a data directory whose persisted slot duration disagrees with the config file fails with `GenesisMismatch::SlotDuration`. Its blocks are indexed against a different time grid, which makes it as foreign as another genesis, and the state cannot reveal this because the duration is deliberately absent from it.
- **Other clients ignore the key.** They hold the value at compile time (leanSpec's `SECONDS_PER_SLOT`) and their genesis parsers ignore unknown keys, so setting it only takes effect on a network where every node reads it. Until the spec adopts a config-file key, this is an ethlambda-only knob.

## Tests

- Config parsing: default, override, non-multiple rejected, zero rejected.
- `ChainConfig`: SSZ round trip, legacy blob decoding as the default cadence, garbage rejected.
- Storage: `from_db_state` rejects a slot-duration mismatch, and resumes a pre-`MILLISECONDS_PER_SLOT` data directory.
- Interval grid: offsets scale with the slot duration, conversions round-trip at 8 s, `ms_until_next_interval` at and around a boundary, the early window never exceeds an interval at any cadence.
- `store::on_tick` advances one interval per configured interval at 8 s, rolling the slot over after five.
- `GET /lean/v0/config/spec` reports `8000`/`1600` for a store built at 8 s.

### DIFF
--- a/crates/blockchain/src/lib.rs
+++ b/crates/blockchain/src/lib.rs
@@ -16,8 +16,8 @@ use ethlambda_types::{
 
 use crate::aggregation::{
     AggregateProduced, AggregationDeadline, AggregationDone, AggregationSession,
-    EarlyAggregationCheck, MAX_AGGREGATION_JOBS, PRIOR_WORKER_JOIN_TIMEOUT, aggregation_deadline,
-    early_aggregation_window, run_aggregation_worker,
+    EARLY_AGGREGATION_WINDOW, EarlyAggregationCheck, MAX_AGGREGATION_JOBS,
+    PRIOR_WORKER_JOIN_TIMEOUT, aggregation_deadline, run_aggregation_worker,
 };
 use crate::key_manager::ValidatorKeyPair;
 use crate::sync_status::SyncStatusTracker;
@@ -409,12 +409,11 @@ impl BlockChainServer {
 
                 // Schedule the early-aggregation window check. This tick is
                 // one interval before T2, so the timer fires right as the
-                // window opens at T2 - `early_aggregation_window`.
+                // window opens at T2 - EARLY_AGGREGATION_WINDOW.
                 if is_aggregator {
-                    let ms_per_interval = time_config.milliseconds_per_interval();
                     send_after(
-                        Duration::from_millis(ms_per_interval)
-                            - early_aggregation_window(ms_per_interval),
+                        Duration::from_millis(time_config.milliseconds_per_interval())
+                            - EARLY_AGGREGATION_WINDOW,
                         ctx.clone(),
                         EarlyAggregationCheck,
                     );
@@ -577,7 +576,7 @@ impl BlockChainServer {
     }
 
     /// Early-aggregation trigger: start the slot's session ahead of the
-    /// interval-2 tick when, inside the window `[T2 - early_aggregation_window, T2)`,
+    /// interval-2 tick when, inside the window `[T2 - EARLY_AGGREGATION_WINDOW, T2)`,
     /// a single attestation-data group already holds 2/3 of the signatures
     /// expected from this node's aggregation subnets. Called after every
     /// stored current-slot gossip signature and once at the window opening via
@@ -592,7 +591,7 @@ impl BlockChainServer {
             return;
         }
         // Only fire inside the early-aggregation window
-        // `[T2 - early_aggregation_window, T2)`, where T2 is the current
+        // `[T2 - EARLY_AGGREGATION_WINDOW, T2)`, where T2 is the current
         // slot's interval-2 boundary; the slot is derived from the wall clock.
         let time_config = *self.store.config();
         let Some(ms_since_genesis) = unix_now_ms().checked_sub(time_config.genesis_time_ms())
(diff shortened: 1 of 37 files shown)


crates/blockchain/src/lib.rs
```diff
@@ -718,8 +725,9 @@ impl BlockChainServer {
     async fn propose_block(&mut self, slot: u64, validator_id: u64) {
         info!(%slot, %validator_id, "We are the proposer for this slot");
 
-        let genesis_time_ms = self.store.config().genesis_time * 1000;
-        let slot_start_ms = genesis_time_ms + slot * MILLISECONDS_PER_SLOT;
+        let time_config = *self.store.config();
+        let slot_start_ms =
+            time_config.genesis_time_ms() + slot * time_config.milliseconds_per_slot;
```
pablodeymo: Tiny dedup opportunity: this hand-rolls the slot-start formula that `SlotInterval` already centralizes — `SlotInterval::BlockPublication.to_ms_since_genesis(slot, &time_config)` is a value-identical drop-in (interval index 0), and it's what the adjacent `t2_ms` computation already uses via `SlotInterval::Aggregation`. Same applies to `get_proposal_head` in `store.rs` (`store.config().genesis_time_ms() + slot * store.config().milliseconds_per_slot`), where it would also drop the doubled `store.config()` call. Keeps the time grid encoded in exactly one place if its shape ever changes.

pablodeymo:
Not introduced by this PR (the multiply pre-dates it), but since these exact lines are re-plumbed here, noting for a follow-up: `Handler<NewBlock>` / `Handler<NewAttestation>` call `observe_gossip_*_arrival` **before** `on_block` / `on_gossip_attestation` validation, so the gossip-supplied slot reaches `to_ms_since_genesis`'s `slot * config.milliseconds_per_slot` (`crates/blockchain/src/metrics.rs`) unbounded. A hostile slot near `u64::MAX` wraps in release (garbage delta + wrong `position` label — no panic, since the workspace profiles don't enable overflow-checks) and panics the BlockChain actor in debug builds. A `checked_mul` or an early slot-sanity bound before observing would close it. Happy to file this as a separate issue if you prefer.

pablodeymo (approved):

PR #600, opened 2026-08-31
feat(p2p): add a TCP (noise + yamux) transport alongside QUIC

MegaRedHand:
## What changes

A libp2p TCP transport (noise + yamux) is added alongside QUIC, bound to the same port number as the existing QUIC listener. TCP and UDP are separate namespaces, so one `--gossipsub-port` still names both.

The reason it matters: a peer that advertises a `quic` entry nothing answers on leaves a QUIC-only node with no second address to try. The dial times out and the peer is never reached. With both transports offered, libp2p races a peer's `quic` and `tcp` addresses within one dial attempt.

| Area | Before | After |
| --- | --- | --- |
| Swarm transports | QUIC only | `.with_tcp(nodelay, noise, yamux)` then `.with_quic()` |
| Listeners | `/udp/<port>/quic-v1` | that, plus `/tcp/<port>` |
| Admission | `quic` entry required (`NoQuicPort`) | `quic` or `tcp` (`NoDialableTransport`) |
| Dial target | one `Multiaddr` | `DialOpts` carrying the peer's full address list |
| `Bootnode` | `quic_port`, `udp_port` | plus `tcp_port`, same "port 0 is absent" rule |
| Local ENR | `ip`/`udp`/`quic`/`secp256k1`/`eth2`/`attnets` | plus `tcp` |
| `as_discovery_node` | hands ethrex `tcp_port: 0` | hands it the real advertised port |
| Bind failure | `.expect()` after bootnodes were dialed | typed `SwarmBuildError::Listen`, plus a CLI pre-flight |

`SwarmCommand::Dial` and `SwarmHandle::dial` now carry a `DialOpts` (peer id plus address list) rather than a single `Multiaddr`, since that list is what makes the per-dial race possible. Static bootnode dialing, bootnode redialing, and the discovery dial loop all go through it.

**The list order is not a preference.** libp2p pushes up to `dial_concurrency_factor` addresses into one `FuturesUnordered`, and the default factor exceeds the two a lean peer can offer, so both handshakes always run and the mesh lands on whichever finishes first. The cost is paid on every dial rather than only on a failure: two sockets and two handshakes per peer, on both ends, until one wins. That is the trade for never waiting out a dead-QUIC connect timeout, and the docs say so rather than claiming an ordering the swarm does not honour.

Static bootnodes are merged by `PeerId` before anything is dialed, not while dialing: two entries can name one peer (the same ENR pasted twice, or an old and a new record for one key), `parse_enrs` does not dedup, and `DialOpts::peer_id` refuses a second dial to a peer already being dialed. Merging first is what puts every address the duplicates offered into the one attempt the peer gets — the case that matters is a stale QUIC-only record listed ahead of a newer one with a live TCP port.

`--gossipsub-port` now shares a namespace with the HTTP servers, so `NodeOptions::validate_ports` rejects every clash before anything binds: against `--discovery.port` (both UDP), and against `--api-port`/`--metrics-port` (both TCP). Port `0` is skipped in each comparison, since two `0` binds always land on different OS-assigned ports and can never collide.

`lean_peer_connections_by_transport_total` (new, ethlambda-specific) counts established connections by the transport that carried them, read off the connection's own multiaddr rather than off which address we dialed. That is the production-visible answer to whether the fallback is doing work; `lean_peer_connection_events_total` is left alone because its `direction`/`result` label set is leanMetrics-specified.

Advertising `tcp` also clears lighthouse's discovery predicate, which requires `enr.tcp4().is_some() || enr.tcp6().is_some()` on top of the `fork_digest` comparison. That is only half the story: the lean fork digest is still the cross-client dummy `0x12345678`, so a beacon-chain client rejects us on that instead. The docs say so rather than claiming the gap is closed.

No manifest change. The pinned libp2p fork (`2f14d0ec`, `features = ["full"]`) already carries `libp2p-tcp`, `libp2p-noise` and `libp2p-yamux`.

## Accepted limitations, documented rather than fixed

- **Static bootnodes are not fork-digest gated, and the redial is unbounded.** `parse_enr` never reads `eth2`; that check lives on the discovered-peer path only. With `tcp`-only records now dialable, a beacon-chain bootnode list becomes a set of static dial targets whose handshake succeeds, occupies a `--discovery.target-peers` slot, contributes no subnets, and is redialed at a flat interval forever once the remote drops us. Bootnodes are operator-supplied, so pointing at another network's list is a configuration mistake; splitting the flag into initial peers (dial once) and bootnodes (discv5 seeds) is the real fix and is left to a follow-up. Written up under `docs/discovery.md`'s known limitations.
- **The ENR sequence number is not bumped for the new entry set.** The record still signs at ethrex's `INITIAL_ENR_SEQ`. A fixed local floor was tried and reverted: ethrex's discv5 re-signs at `seq + 1` whenever IP voting moves the advertised address, so a NAT'd node may already be serving a higher seq than any constant we could pick, and the floor would fail exactly where it was supposed to help. Doing this properly needs a persisted counter or a timestamp-derived seq, which is its own change. Written up under `docs/discovery.md`'s known limitations alongside the bootnode one.

## Tests

- Four admission cases: tcp-only, quic-only, both (asserting both addresses are offered), and each transport at port `0` with the other absent.
- `two_swarms_connect_over_tcp` builds two real swarms through `build_swarm` (port `0`, so it cannot collide with a running node), learns the listener's TCP address off its own `NewListenAddr`, dials it, and polls both until each reports `ConnectionEstablished` over `tcp`. A regression to QUIC-only hangs to the 10s timeout rather than passing.
- `merging_bootnodes_keeps_every_address_for_the_initial_dial` pins the merge on the case that distinguishes merge-then-dial from dial-per-entry: a QUIC-only record ahead of a TCP-only one for the same key. `a_bootnode_named_twice_is_dialed_once_with_both_addresses` builds a real swarm over that list, and since the merged map *is* the dial input, asserting the map asserts the dial.
- `port_zero_never_counts_as_a_clash` covers the `0 == 0` wrinkle for each flag `validate_ports` compares.
- `local_enr_advertises_udp_quic_and_tcp` and `local_enr_omits_a_zero_quic_and_tcp_port` pin the writer; the two `NoQuicPort` rejections are renamed to say "and no tcp".
- `From<&DialError> for DialOutcome` now matches exhaustively, so a variant added on a libp2p fork upgrade breaks the build instead of being absorbed as `Unreachable`.

`make lint` clean; `make test` green with fixtures in place: 122 forkchoice, 119 SSZ and 74 STF spec cases, 56 p2p unit tests, 56 in the binary crate.

## Deployment note

Nodes bind an extra **TCP** port at `--gossipsub-port` and advertise it in their ENR. The image `EXPOSE`s `9001/tcp` alongside `9001/udp`, but a devnet that maps only `-p 9001:9001/udp` will still publish a `tcp` entry that is unreachable from outside, and every peer racing the address list then burns a TCP connect timeout on a black-holed port. QUIC is dialed in the same attempt and still works, so this degrades rather than breaks, but adding the TCP mapping is what makes the fallback usable.

### DIFF
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -21,10 +21,15 @@ pub(crate) struct NodeOptions {
     /// Directory containing per-validator XMSS keys (e.g., hash-sig-keys/).
     #[arg(long)]
     pub(crate) hash_sig_keys_dir: PathBuf,
-    /// UDP port for the libp2p QUIC listener.
+    /// Port for the libp2p listeners: UDP for QUIC and TCP for the noise+yamux
+    /// fallback, both on this same number.
+    ///
+    /// TCP and UDP are separate namespaces, so one number names both. It must
+    /// still differ from every other port the node binds: `--discovery.port`
+    /// (also UDP), and `--api-port`/`--metrics-port` (also TCP).
     ///
     /// Defaults one above `--discovery.port` so that `--discovery.enable` works
-    /// on its own: both are UDP sockets and cannot share a port.
+    /// on its own.
     #[arg(long, default_value = "9001")]
     pub(crate) gossipsub_port: u16,
     #[arg(long, default_value = "127.0.0.1")]
@@ -160,11 +165,16 @@ pub(crate) struct DiscoveryConfig {
 }
 
 impl NodeOptions {
-    /// Reject a discovery port that collides with the QUIC port.
+    /// Reject port assignments that cannot all bind, before anything binds.
     ///
-    /// Both are UDP. Without this the collision surfaces at bind time as an
-    /// opaque `EADDRINUSE` on whichever socket loses the race.
-    pub(crate) fn validate_discovery(&self) -> eyre::Result<()> {
+    /// There are two clashes to catch, on two protocols. `--discovery.port` and
+    /// `--gossipsub-port` are both UDP. `--gossipsub-port` also binds TCP for
+    /// the noise+yamux listener, which puts it in the same namespace as the
+    /// HTTP servers: sharing that number with `--api-port` was legal while the
+    /// swarm bound UDP only, and is now a real collision. Without these checks
+    /// either surfaces at bind time as an opaque `EADDRINUSE` on whichever
+    /// socket loses the race.
+    pub(crate) fn validate_ports(&self) -> eyre::Result<()> {
         if self.discovery.enable && self.discovery.port == self.gossipsub_port {
             eyre::bail!(
                 "--discovery.port ({}) must differ from --gossipsub-port ({}): \
@@ -172,6 +182,28 @@ impl NodeOptions {
                 self.discovery.port,
                 self.gossipsub_port
             );
+        }
+        // A discovery-enabled node publishes `quic` and `tcp` entries naming
+        // this port. Port 0 asks the OS to pick, so the two listeners land on
+        // different real ports and the record advertises neither of them: a
+        // peer reading it finds nothing dialable.
+        if self.discovery.enable && self.gossipsub_port == 0 {
+            eyre::bail!(
+                "--gossipsub-port 0 cannot be used with --discovery.enable: the \
+                 advertised ENR would name port 0, which no peer can dial"
+            );
+        }
+        for (flag, port) in [
+            ("--api-port", self.api_port),
+            ("--metrics-port", self.metrics_port),
+        ] {
+            if port == self.gossipsub_port {
+                eyre::bail!(
+                    "{flag} ({port}) must differ from --gossipsub-port ({}): the \
+                     libp2p swarm binds TCP on that port as well as UDP",
+                    self.gossipsub_port
+                );
+            }
         }
         Ok(())
     }
@@ -255,16 +287,58 @@ mod tests {
         let options = parse(&["--discovery.enable"]);
 
         assert_ne!(options.discovery.port, options.gossipsub_port);
-        assert!(options.validate_discovery().is_ok());
+        assert!(options.validate_ports().is_ok());
     }
 
     #[test]
-    fn colliding_ports_are_rejected_only_when_discovery_is_enabled() {
+    fn colliding_udp_ports_are_rejected_only_when_discovery_is_enabled() {
         let ports = ["--gossipsub-port", "9000", "--discovery.port", "9000"];
         let mut enabled = ports.to_vec();
         enabled.push("--discovery.enable");
 
-        assert!(parse(&ports).validate_discovery().is_ok());
-        assert!(parse(&enabled).validate_discovery().is_err());
+        assert!(parse(&ports).validate_ports().is_ok());
+        assert!(parse(&enabled).validate_ports().is_err());
+    }
+
+    /// Unlike the UDP clash above, this one does not depend on discovery: the
+    /// swarm binds TCP either way, so an HTTP port sharing the number always
+    /// loses one of the two listeners.
+    #[test]
+    fn an_http_port_sharing_the_gossipsub_port_is_rejected() {
+        // A port no default claims, so only the flag under test collides and
+        // the message can be checked for naming it.
+        const SHARED: &str = "9100";
+
+        for flag in ["--api-port", "--metrics-port"] {
+            let err = parse(&["--gossipsub-port", SHARED, flag, SHARED])
+                .validate_ports()
+                .expect_err("a TCP clash with an HTTP port must be rejected");
+            assert!(
+                err.to_string().contains(flag),
+                "the message must name the offending flag, got: {err}"
+            );
+        }
+    }
+
+    /// `--api-port` and `--metrics-port` sharing one number is supported (the
+    /// RPC crate merges the routers onto a single listener), so the TCP check
+    /// must not sweep that up.
+    #[test]
+    fn api_and_metrics_may_share_a_port() {
+        let options = parse(&["--api-port", "5052", "--metrics-port", "5052"]);
+
+        assert!(options.validate_ports().is_ok());
+    }
+
+    /// Port 0 leaves the two listeners on different OS-assigned ports, so the
+    /// one number the ENR publishes describes neither.
+    #[test]
+    fn gossipsub_port_zero_is_rejected_with_discovery() {
+        assert!(parse(&["--gossipsub-port", "0"]).validate_ports().is_ok());
+        assert!(
+            parse(&["--gossipsub-port", "0", "--discovery.enable"])
+                .validate_ports()
+                .is_err()
+        );
     }
 }
--- a/bin/ethlambda/src/cli.rs
+++ b/bin/ethlambda/src/cli.rs
@@ -174,8 +174,17 @@ impl NodeOptions {
     /// swarm bound UDP only, and is now a real collision. Without these checks
     /// either surfaces at bind time as an opaque `EADDRINUSE` on whichever
     /// socket loses the race.
+    ///
+    /// Every comparison skips `0`, which is not a port but a request for one:
+    /// two `0` binds always land on different OS-assigned ports and can never
+    /// collide. Rejecting a pair of them would refuse the setup that exists to
+    /// avoid collisions, which test harnesses and several-nodes-per-host runs
+    /// rely on.
     pub(crate) fn validate_ports(&self) -> eyre::Result<()> {
-        if self.discovery.enable && self.discovery.port == self.gossipsub_port {
+        if self.discovery.enable
+            && self.gossipsub_port != 0
+            && self.discovery.port == self.gossipsub_port
+        {
             eyre::bail!(
                 "--discovery.port ({}) must differ from --gossipsub-port ({}): \
                  both bind UDP and cannot share a port",
@@ -197,7 +206,7 @@ impl NodeOptions {
             ("--api-port", self.api_port),
             ("--metrics-port", self.metrics_port),
         ] {
-            if port == self.gossipsub_port {
+            if self.gossipsub_port != 0 && port == self.gossipsub_port {
                 eyre::bail!(
                     "{flag} ({port}) must differ from --gossipsub-port ({}): the \
                      libp2p swarm binds TCP on that port as well as UDP",
@@ -340,5 +349,34 @@ mod tests {
                 .validate_ports()
                 .is_err()
         );
+    }
+
+    /// Two `0`s are two OS-assigned ports, so the equality checks must not read
+    /// them as a clash: without discovery, `--gossipsub-port 0` is a supported
+    /// configuration (the test above pins that), and it stays supported when an
+    /// HTTP port asks the OS to pick as well.
+    #[test]
+    fn port_zero_never_counts_as_a_clash() {
+        for flag in ["--api-port", "--metrics-port", "--discovery.port"] {
+            let mut argv = vec!["--gossipsub-port", "0", flag, "0"];
+            if flag == "--discovery.port" {
+                argv.push("--discovery.enable");
+                // Discovery rejects a `0` gossipsub port on its own grounds:
+                // the ENR would publish a port no peer can dial. What must not
+                // happen is the pair being rejected as a collision.
+                let err = parse(&argv)
+                    .validate_ports()
+                    .expect_err("gossipsub port 0 stays invalid under discovery");
+                assert!(
+                    !err.to_string().contains("must differ"),
+                    "0 == 0 must not be reported as a clash, got: {err}"
+                );
+                continue;
+            }
+            assert!(
+                parse(&argv).validate_ports().is_ok(),
+                "{flag} 0 alongside --gossipsub-port 0 must be accepted"
+            );
+        }
     }
 }
--- a/crates/net/p2p/src/discovery/admission.rs
+++ b/crates/net/p2p/src/discovery/admission.rs
@@ -27,15 +27,15 @@ use tracing::debug;
 
 use super::enr::{
     ATTNETS_ENR_KEY, ETH2_ENR_KEY, EnrForkId, read_ip, read_public_key, read_quic_port,
-    subnets_from_attnets,
+    read_tcp_port, subnets_from_attnets,
 };
-use crate::{quic_multiaddr, tcp_multiaddr};
+use crate::dial_addrs;
 
 /// A peer that passed admission and is ready to dial.
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) struct DiscoveredPeer {
     pub(crate) peer_id: PeerId,
-    /// Dial targets, QUIC first then TCP, built from whichever of the two ports
+    /// Every dial target for this peer, built from whichever of the two ports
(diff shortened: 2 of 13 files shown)


crates/net/p2p/src/lib.rs
```diff
@@ -303,41 +311,50 @@ pub fn build_swarm(
...
             continue;
-        };
-        let addr = quic_multiaddr(bootnode.ip, quic_port, peer_id);
-        bootnode_addrs.insert(peer_id, addr.clone());
-        swarm.dial(addr).unwrap();
+        }
+        bootnode_addrs.insert(peer_id, addrs.clone());
+        swarm
+            .dial(DialOpts::peer_id(peer_id).addresses(addrs).build())
+            .unwrap();
```
pablodeymo: This switch to `DialOpts::peer_id(...)` changes the dial condition from `Always` (the old `From<Multiaddr>` path) to the default `DisconnectedAndNotDialing`, but the `.unwrap()` stayed. A bootnodes list containing the same node twice (same ENR pasted twice, or an old+new ENR for one secp256k1 key — both decode to one `PeerId`, and `parse_enrs` doesn't dedup) now panics on the second entry with `DialError::DialPeerConditionFalse` before `listen_on` ever runs — a file that booted fine pre-PR aborts the node. Either dedup by `PeerId` in `parse_enrs` or treat `DialPeerConditionFalse` as non-fatal (log + continue).

crates/net/p2p/src/lib.rs
```diff
@@ -303,41 +311,50 @@ pub fn build_swarm(
...
+        .listen_on(quic_addr)
+        .expect("failed to bind gossipsub QUIC listening address");
+    // Same port number as the QUIC listener above: TCP and UDP are separate
+    // namespaces, so this cannot collide with it.
+    let tcp_addr = Multiaddr::empty()
+        .with(config.listening_socket.ip().into())
+        .with(Protocol::Tcp(config.listening_socket.port()));
+    swarm
+        .listen_on(tcp_addr)
+        .expect("failed to bind gossipsub TCP listening address");
```
pablodeymo: The new TCP bind is a hard `.expect()`, so any process already holding TCP on `--gossipsub-port` (e.g. sharing the port number with the API server, which was legal pre-PR since only UDP was bound; or a stale container on tcp/9001) now aborts the node at startup after QUIC bound and bootnodes were dialed, instead of degrading to today's QUIC-only behaviour. `build_swarm` already returns `Result`, so this could be a typed error — and the CLI side should follow: `validate_discovery` still only checks the UDP clash with `--discovery.port` ("both bind UDP"), and the `--gossipsub-port` help text still reads "UDP port for the libp2p QUIC listener".

crates/net/p2p/src/discovery/admission.rs
```diff
@@ -156,9 +168,19 @@ fn admit(
         .map(|bits| subnets_from_attnets(&bits, attestation_committee_count))
         .unwrap_or_default();
 
+    // QUIC first, so it stays the preferred path when a peer offers both and
```
pablodeymo: "QUIC first, so it stays the preferred path" doesn't hold: the pinned fork's `ConcurrentDial` (swarm `pool/concurrent_dial.rs`) pushes up to `dial_concurrency_factor` (default 8) addresses into a `FuturesUnordered` at once, so with two addresses list order confers no preference — both handshakes run on every dial (2x sockets and crypto on both ends) and the mesh nondeterministically lands on whichever finishes first. To make the ordering real and TCP a true fallback: `.override_dial_concurrency_factor(NonZeroU8::new(1).unwrap())` on the `DialOpts` builders, or dial TCP only after a QUIC failure. (The same ordering claim appears in `lib.rs`, `swarm_adapter.rs` and `docs/discovery.md`.)

crates/net/p2p/src/lib.rs
```diff
@@ -545,9 +562,10 @@ impl P2PServer {
             return;
         }
 
-        if let Some(addr) = self.bootnode_addrs.get(&peer_id) {
+        if let Some(addrs) = self.bootnode_addrs.get(&peer_id) {
             trace!(%peer_id, "Redialing disconnected bootnode");
-            self.swarm_handle.dial(addr.clone());
+            self.swarm_handle
+                .dial(DialOpts::peer_id(peer_id).addresses(addrs.clone()).build());
```
pablodeymo, line 595: With tcp-only records now dialable, a beacon-chain bootnode list (the stated use case for clearing lighthouse's predicate) becomes a set of static dial targets with no fork-digest gate — `parse_enr` never reads `eth2`; that check lives only on the discovered-peer path (`admission.rs`). The TCP+noise+yamux handshake to a lighthouse node succeeds, the peer occupies a `target_peers` slot contributing zero subnets, and when the remote drops us for having no shared protocols, `ConnectionClosed` re-arms `RetryPeerRedial` every 12s with no backoff or cap — permanent hammering of third-party mainnet infrastructure. Applying the same `eth2`/fork-digest admission to static bootnodes (or at least a redial backoff/cap) would close this.

pablodeymo:
Nice, careful port overall — admission, dial construction, ENR, and bootnode parsing move together, and `two_swarms_connect_over_tcp` is a real end-to-end check. I left inline comments on four issues I'd consider blocking (a startup panic on duplicated bootnode entries, a startup panic on an occupied TCP port, the dial-concurrency assumption behind "QUIC first", and ungated beacon-chain bootnodes + unbounded 12s redial). A few more, mostly non-blocking:

- **`Dockerfile` still `EXPOSE`s only `9001/udp`** — the new TCP listener on 9001 isn't exposed, so a containerized node advertises a `tcp` entry nothing outside can reach, and every peer racing the address list burns a TCP connect timeout on a black-holed port: the failure mode this PR exists to remove, inverted. The deployment note already says the TCP mapping is what makes the fallback usable — the image should ship it: `EXPOSE 9001/tcp`. (Blocking-adjacent, but it's a one-liner.)
- **ENR seq never bumps**: the record now differs across versions (new `tcp` entry) but is still signed at constant `INITIAL_ENR_SEQ = 1`. ethrex's WHOAREYOU responder omits the record when `enr_seq` matches and the peer table only accepts strictly-higher seq, so peers that stayed up across an in-place upgrade keep the tcp-less record forever — the lighthouse-predicate benefit is invisible to exactly the peers that already know us. The constant seq is pre-existing, but this PR is the first payload change that makes it bite.
- **Silently dropped dial candidate** (`discovery/dial.rs`): with the condition now `DisconnectedAndNotDialing`, a synchronous rejection can happen for any in-flight peer, not just `NoAddresses`/`Denied`. The candidate was already popped and marked tried, and the early return skips the `peer_attnets` insert, so `covered_subnets` under-counts permanently. The comment above the block still describes the old reality.
- **`tcp` written into the ENR unconditionally** (`discovery/enr.rs`): the advertised ports are copied from config, not derived from the bound listeners (contrast `discovery_port`, which uses `bound.port()`). With `--gossipsub-port 0` the ENR publishes `quic:0` and `tcp:0` — `Some(0)` satisfies lighthouse's `tcp4().is_some()` predicate, our own `admit()` would reject the record as `NoDialableTransport`, and the two OS-assigned real ports differ, breaking the "same port number" invariant. Readers filter `0` as absent; the writer has no matching guard.
- **Duplication that's now load-bearing**: the QUIC-first list builder is inlined byte-for-byte in both `admit()` and `bootnode_dial_addrs`, and the `.filter(|p| *p != 0)` tcp rule is open-coded twice with no `read_tcp_port` sibling to `read_quic_port`. Once the ordering is made real (concurrency = 1), divergence between the two sites changes behaviour silently — worth extracting one `dial_addrs(ip, quic, tcp, peer_id)` helper.
- **Test asserting its own inversion**: `parse_enrs_keeps_a_quic_less_record_as_a_discovery_seed`'s comment still says the record is kept "with `quic_port: None` telling build_swarm not to dial it", but the first embedded mainnet ENR carries `tcp=9000`, so post-PR it *is* dialed — and nothing asserts `Bootnode.tcp_port` or exercises `bootnode_dial_addrs`. Renaming it and asserting `bootnodes[0].tcp_port == Some(9000)` / `bootnodes[1].tcp_port == None` would pin the new behaviour.

One free win worth adding to the description: the `DialOpts` switch also fixes redial stacking — the dial condition now refuses a new dial while one is in flight.

crates/net/p2p/src/lib.rs
```diff
@@ -310,42 +328,89 @@ pub fn build_swarm(
...
+            }
+            Entry::Vacant(unknown) => {
+                unknown.insert(addrs.clone());
+            }
+        }
+        // A refused dial is not fatal: the entry stays in `bootnode_addrs`, so
+        // the redial path picks the peer up. Unwrapping here would abort a node
+        // over a bootnode file that is merely redundant.
+        swarm
+            .dial(DialOpts::peer_id(peer_id).addresses(addrs).build())
```
pablodeymo: The merge doesn't reach the initial dial: it fires from the `Entry::Vacant` arm with only the first entry's addresses, so addresses contributed by later duplicate entries are only ever used by the 12s redial path. In the exact stale-record case the comment above describes (an old QUIC-only ENR first, a newer ENR with a live TCP port second, QUIC dead), startup races only the dead QUIC address, burns the full connect timeout, and the live TCP address isn't tried until `OutgoingConnectionError` schedules the redial — timeout + 12s instead of connecting immediately. The comment ("dials once with everything the duplicate entries offered") and the test currently assert the intent, not the behaviour — the test only checks the `bootnode_addrs` map, not what was dialed. Fix: collect and merge first, then dial once per peer from the map's merged lists after the loop.

crates/net/p2p/src/discovery/enr.rs
```diff
@@ -160,14 +169,41 @@ impl LocalEnrParams {
...
+///
+/// ethrex re-signs at a higher seq of its own whenever discv5's IP voting moves
+/// the advertised address, so this is a floor rather than the only value a
+/// record is ever seen at.
+///
+/// | Value | Entry set |
+/// | --- | --- |
+/// | 1 | `id`, `ip`, `udp`, `quic`, `secp256k1`, `eth2`, `attnets` |
+/// | 2 | the above plus `tcp` |
+const LOCAL_ENR_SEQ: u64 = 2;
```
pablodeymo: A fixed `LOCAL_ENR_SEQ = 2` doesn't survive IP voting: ethrex's discv5 re-signs the record at `seq + 1` at runtime (`NodeRecord::edit`), so a pre-PR node running behind NAT with `--discovery.enable` may already be serving seq >= 2. After an in-place upgrade with the same `--node-key`, the new binary signs at seq 2, the WHOAREYOU responder omits the record when `enr_seq` matches, and the peer table only accepts strictly higher — so long-lived peers keep the tcp-less record indefinitely, which is the exact invisibility this constant exists to prevent (the doc comment even notes the IP-voting bumps, but the floor only pins the un-bumped case). A persisted counter (bump on every content change) or a timestamp-derived seq (e.g. seconds since a fixed epoch at startup) closes it; the latter is what several beacon clients do and needs no storage.

bin/ethlambda/src/cli.rs
```diff
@@ -173,6 +183,28 @@ impl NodeOptions {
...
+            eyre::bail!(
+                "--gossipsub-port 0 cannot be used with --discovery.enable: the \
+                 advertised ENR would name port 0, which no peer can dial"
+            );
+        }
+        for (flag, port) in [
+            ("--api-port", self.api_port),
+            ("--metrics-port", self.metrics_port),
+        ] {
+            if port == self.gossipsub_port {
```
pablodeymo: `validate_ports` compares raw port numbers, so `--gossipsub-port 0 --metrics-port 0` (or `--api-port 0`) is rejected with "must differ" even though two port-0 binds always receive distinct OS-assigned ports and can never collide — and port 0 for gossipsub without discovery is a configuration this PR's own `gossipsub_port_zero_is_rejected_with_discovery` test pins as otherwise supported. Test harnesses and multi-instance-on-one-host setups rely on 0 exactly to avoid collisions. Guard the comparison with `port != 0`. (The pre-existing `--discovery.port` check shares the same 0 == 0 wrinkle if you want to fix both.)

crates/net/p2p/src/lib.rs
pablodeymo: This one looks unaddressed in 39692cff — `parse_enr` still doesn't read `eth2`, and the redial is still a flat 12s with no backoff or cap. If the position is "bootnodes are operator-supplied, so pointing at a beacon-chain list is operator error," that's defensible — but then it deserves a line in `docs/discovery.md` and ideally a cap or backoff on the redial loop so a mistake degrades gracefully instead of hammering third-party infrastructure forever. Could you say which way you want to take it?

pablodeymo:
Round 2 looks much better — the dedup, typed bind errors + CLI pre-flight, the honest retraction of "QUIC first" (keeping the race with its cost documented), `dial_addrs`/`read_tcp_port`, `EXPOSE 9001/tcp`, and the renamed dial-over-tcp test all resolve their round-1 findings cleanly. I left three inline comments on new issues the fix introduced (the dedup merge never reaching the initial dial, `LOCAL_ENR_SEQ` vs. IP-voting bumps, and `validate_ports` rejecting port-0 pairs), plus a question on the still-open beacon-bootnode thread. Smaller, non-blocking:

- **`read_tcp_port`'s doc is false at merge time**: it claims to be "deliberately the only place" the port-0 rule is spelled and that "the ENR writer goes through it," but `local_pairs` open-codes its own filter, and the rule also appears in `read_quic_port`, `local_pairs`' quic guard, and `parse_enr`'s udp filter. One shared `fn dialable_port(port: u16) -> Option<u16>` used by readers, writer, and parser would make the doc true and the rule unforgettable.
- **`DiscoverySpawnConfig` / `LocalEnrParams` carry `quic_port` and `tcp_port` that can never legitimately differ** — `build_swarm` binds both from the single `listening_socket` port, and every construction site passes identical values, held equal only by comments. A single `p2p_port` field makes the invariant structural and deletes three copies of the "same number, separate namespaces" comment.
- **The `transport` field is trace-only**: the comment says it "answers whether the TCP fallback actually helped," but both consumers are `trace!` lines, invisible at default verbosity, and nothing in the repo sets `RUST_LOG` — so the PR's effectiveness signal is unobservable in production. One `transport` label on `lean_peer_connection_events_total` (computed right next to `notify_peer_connected`) would make it real.
- **`unwrap_or_else` + `warn!` on the initial bootnode dial** is side-effect-only error handling; repo CLAUDE.md prescribes `inspect_err` here (`let _ = swarm.dial(...).inspect_err(|err| warn!(...))`) — the same idiom this PR already uses in `swarm_adapter.rs`.
- Defensive, not a defect: `From<&DialError> for DialOutcome` classifies with a wildcard `_ => Unreachable`. `DialError` isn't `#[non_exhaustive]` in the pinned fork, so an exhaustive match is possible and would force a review of the mapping on fork upgrades instead of silently absorbing a new in-flight variant.

crates/net/p2p/src/lib.rs
MegaRedHand: The continuous redial is something we can accept. We can split these into initial peers and bootnodes in a following PR

pablodeymo (approved):

PR #603, opened 2026-09-02
refactor(blockchain): keep the aggregation worker running at all times

MegaRedHand:
## What

Aggregation stops being a per-slot session. One worker thread starts with the blockchain actor and lives as long as it does, holding its own `Store` clone: rank the pool's candidates, prove the best job, hand it to the actor, rank again. The actor buffers each aggregate and gossips the lot at the vote-aggregation interval.

```
before:  interval 2 (or up to 600ms early)  ->  snapshot pool, spawn worker,
                                                <= MAX_AGGREGATION_JOBS jobs,
                                                cancel at a soft deadline
after:   always                              ->  rank, prove best, send, repeat
         interval 2                          ->  publish what the worker produced
```

## Why

The session existed to fit XMSS proving into one interval. The prover sat idle for most of the slot and then had ~800ms to do everything, which is what the early-start window and the job caps were compensating for. Splitting *proving* from *publishing* removes the deadline without moving what the network sees: aggregates still only go out at interval 2.

## Changes

| Area | Change |
|------|--------|
| `aggregation.rs` | `spawn_aggregation_worker` + a loop over `select_best_job`, with a `JobPolicy` derived from the wall clock; sessions, session ids, `AGGREGATION_DEADLINE`, `EarlyAggregationCheck`, `AggregationDeadline`, `AggregationDone`, `MAX_AGGREGATION_JOBS` and the publish-alignment all gone. `EARLY_AGGREGATION_WINDOW` survives with a new job: it is now the stretch in which the worker keeps the prover free, not a licence to start early |
| `lib.rs` | `started()` spawns the worker, `stopped()` joins it; `pending_aggregates` buffer drained at interval 2; pause flag raised around `propose_block` |
| `storage/store.rs` | `max_gossip_group_count_for_slot` dropped with its only caller (the early-start check) |
| docs | `architecture.md`, `slots_and_intervals.md`, `spec_deviations.md` |

### What replaces the session machinery

What the worker may take up is a function of where the slot is (`JobPolicy`):

| phase | eligible |
|---|---|
| start of slot → T2 − `EARLY_AGGREGATION_WINDOW` | backlog work (stale groups, merges of proofs already held), plus a current-slot group at the two-thirds floor |
| the last `EARLY_AGGREGATION_WINDOW` before T2 | **only** a current-slot group at the floor |
| T2 onwards | everything, however few signatures back it |

The floor (`min_current_slot_group_sigs`) is the old early-start threshold: two thirds of the votes this node's own subnets are expected to carry. It keeps a slot's votes going out as one wide aggregate rather than several thin ones.

The window is the part worth arguing about. A backlog job is a recursive proof merge that can run well past the boundary, and the prover is single-threaded, so starting one there would delay the very aggregate the slot is waiting on. Idling costs little by comparison: the backlog is not going anywhere, and this is the window where the committee's signatures usually cross the floor anyway.

Separately, the actor raises a **pause flag** around its own block build, which is what the `max_jobs = 1` proposer cap used to buy. A proof already in flight is not interrupted.

### Two details worth a look

- **Apply on arrival, publish later.** The actor applies each aggregate to the store the moment the message lands, because the pool the worker re-reads has to account for it. Only the gossip publication waits for interval 2.
- **Worker-side coverage memory.** Between `send` and the actor's apply, the store still shows the job as pending, so the next round would prove it again. The worker remembers the coverage it emitted per slot (`EmittedCoverage`) to close that window without becoming a store writer. Coverage is recorded on *attempt*, so a failing job is not retried at full prover cost either.

### A plain thread, not `spawn_blocking`

The worker runs for the life of the process and spends it in leanVM proofs, so a blocking-pool thread would be parked permanently for nothing: the loop awaits nothing and reaches the actor through an unbounded channel that needs no reactor. Shutdown polls `is_finished()` instead of blocking a runtime thread on the proof in flight, and leaves the thread detached past `WORKER_JOIN_TIMEOUT`.

## Metrics

- Removed: `lean_aggregation_early_starts_total`, `lean_aggregation_early_start_lead_seconds` — the window they measured is gone. Neither was documented in `docs/metrics.md`.
- `lean_committee_signatures_aggregation_time_seconds` now times one aggregate's proof instead of a session's total.
- `lean_gossip_aggregation_arrival_*` samples our own aggregates at publication rather than production, keeping them comparable with the peers' aggregates they share a histogram with.
- `lean_aggregator_skipped_total{reason="other"}` now counts a job whose proof failed.

## Test status

- `make fmt`, `make lint` (clippy `-D warnings`): clean
- `cargo test --workspace --lib`: 250 passed, 0 failed
- Spec tests not run: they are red on `main` already (fixtures pull rolling latest)
- Local 3-node devnet, 1 aggregator: clean, finality trailing head by 3 slots, proving at t+1024ms and publication at t+1604ms ([results](https://github.com/lambdaclass/ethlambda/pull/603#issuecomment-5515748394))


MegaRedHand:
## Local devnet: 3 ethlambda nodes, 1 aggregator, 20 slots

`attestation_committee_count = 1`, so the single aggregator sees every vote and 2 of 3 validators are enough to justify.

| | ethlambda_0 (aggregator) | ethlambda_1 | ethlambda_2 |
|---|---|---|---|
| blocks imported | 11 | 11 | 12 |
| aggregates proved | 18 | 0 | 0 |
| aggregates gossiped | 16 | 0 | 0 |
| attestations | 19 | 18 | 18 |
| WARN | 0 | 0 | 0 |
| ERROR / panic | 0 | 0 | 0 |

Finality kept up: `finalized_slot=15` with `head_slot=18` at the end, so finalization trailed the head by 3 slots throughout.

### The split between proving and publishing is visible in the timing

Offsets into the slot, from the aggregator's log (interval boundaries at t+0 / 800 / 1600 / 2400 / 3200):

| duty | n | median | min | max |
|---|---|---|---|---|
| attestation published | 19 | t+815ms | t+806ms | t+859ms |
| **aggregate proved** | 18 | **t+1024ms** | t+951ms | t+3617ms |
| **aggregates gossiped** | 16 | **t+1604ms** | t+1602ms | t+1607ms |
| block published | 12 | t+232ms | t+194ms | t+328ms |

The proof lands ~200 ms after the votes do — as soon as the group crosses two thirds, well inside interval 1 — while publication sits on the interval-2 boundary to the millisecond. That is exactly the intent: proving off the grid, publication on it.

### The gate does what it is for

Every aggregate was `raw_sigs=3 children=0 participants=3`, one per slot, at 134-170 ms of prover time. One wide aggregate per slot rather than the several thin ones an ungated always-on worker would produce.

The single `t+3617ms` outlier is a backlog job taken late in a slot, which is the intended use of the otherwise-idle stretch.

<details>
<summary>How it was run</summary>

```bash
DOCKER_TAG=pr603 make docker-build
.claude/skills/devnet-runner/scripts/run-devnet-with-timeout.sh 140
```

Genesis keys came from `blockblaz/hash-sig-cli:latest` (52-byte pubkeys). The lean-quickstart checkout defaults to `ghcr.io/lambdaclass/hash-sig-cli:0.5.0`, which emits the 32-byte leanVM-main keys and makes any `main`-based image fail with `failed to parse genesis config`. Worth knowing before running a local devnet off `main`.
</details>

PR #606, opened 2026-09-03
feat(crypto)!: bump to new leanVM version

MegaRedHand:
## What

Brings ethlambda up to leanVM `main` at `5a4f55c1` (2026-09-03) and adapts to the
API rework that landed there.

This PR is **standalone against `main`**: it branches off the
`build/leanvm-track-main` line, so the diff here is the complete
ethlambda-on-leanVM-main migration, not only the latest bump. The last commit,
`build(deps): track leanVM main's unified aggregation API`, is the new work;
everything before it is the existing tracking branch (#539), which this
subsumes.

Two commits follow the bump: one enabling carryless multiply on aarch64, and one
adding `ethlambda keygen`. Both are described at the end.

leanVM `main` moved `a5909d18` (2026-07-30) → `5a4f55c1`, 718 commits. The whole
crypto stack now arrives through leanVM's own facade crate, `leanvm`, which
re-exports the aggregation API, the `xmss` module and the `rand` that signing
draws from: one dependency at one revision, in place of `xmss` +
`lean-multisig` + a direct `ethereum_ssz`.

## One aggregate type, grouped by epoch

`SingleMessageAggregateSignature` and `MultiMessageAggregateSignature` are gone.
There is one `AggregateSignature`, whose XMSS claims are grouped by epoch: one
`XmssGroup` per slot, carrying the one message signed at it and that group's
sorted, deduplicated keys. Type-1 and Type-2 are now the same object, one group
versus several, so the wrappers in `ethlambda-crypto` differ only in how many
groups they build.

| before | after |
| --- | --- |
| `aggregate_single_message_signatures(children, raw, msg, slot, rate)` | `aggregate(children, raw_xmss, raw_sphincs, declare, rate)` |
| `merge_single_message_aggregates(type_1s, rate)` | the same `aggregate`, children only |
| `split_multi_message_aggregate(t2, index, rate)` | the same `aggregate`, `declare` naming the group to keep |
| `verify_single_message_aggregate` / `verify_multi_message_aggregate` | `sig.verify()` |

## The binding left the wire

`to_bytes_without_pubkeys()` used to carry `(message, slot)` inside the proof, so
`verify_*` could compare it against what the caller expected and `split_*` could
search the decoded components for a message. Neither is possible now: the epoch
and the message live in the `XmssGroup`, which is exactly the part the
without-pubkeys form omits. Every decode has to rebuild the whole signer set from
the caller's own view of it, which is what the new `SignerSet` carries. Message,
slot and keys travel together because none of the three can be supplied without
the others.

Three wrapper signatures change as a result:

- `merge_type_1s_into_type_2` takes `(SignerSet, bytes)` per child.
- `verify_type_2_signature` takes `&[SignerSet]` in place of parallel pubkey and
  binding lists.
- `split_type_2_by_message` takes `&[SignerSet]` for the whole parent, since
  decoding it needs every claim even though one survives.

`VerificationError::BindingMismatch` goes with them. A wrong message or slot
changes the digest the proof is checked against, so it now fails inside the SNARK
verifier rather than at a field compare ahead of it. The check is not weaker, it
just costs a full verification to reach.

## Worth a look: a slot carries one message

`ConflictingMessages` is new upstream: within one aggregate the message is a
function of the epoch (`bind_message` in leanVM's `plan_coverage` keys purely on
`epoch → message`; signer sets never enter it). Two distinct `AttestationData` at
one slot therefore have no representation in a single proof, and `on_block`
rejects only duplicate `AttestationData`, not two entries at one slot. So a fork
that splits attesters at slot N yields a block this leanVM cannot prove, and
disjoint signers do not help.

The normal aggregator path is unaffected: same slot plus same message is the same
`AttestationData`, and partial coverage for it merges into one group with the
keys unioned, which is what `aggregate_proofs` / `aggregate_mixed` do.

Surfaced as `ethlambda_crypto::ConflictingMessages` rather than worked around,
since narrowing what a block may pack is a protocol decision. leanSpec `main`
still models a block's components as parallel `(root, slot)` lists, so the
question is open upstream too.

## XMSS

leanVM's `xmss` is BLAKE2s over binary fields now, not Poseidon over KoalaBear.
The wire sizes are unchanged by coincidence (32-byte keys, 1208-byte signatures,
both still static-asserted in `ethlambda-crypto`), but **every genesis key and
every stored proof from before the rewrite is invalid**: a devnet crossing this
boundary needs fresh genesis and a wiped DB.

The call surface moved too: `key_gen_from_seed` returns `(secret, public)` over an
inclusive epoch range, `sign` draws its randomness from a caller-supplied rng, and
`verify` reorders its arguments. The preparation window is gone, a key's signable
range being fixed at generation, and `XmssSecretKey::prepare` only warms the
cached bottom subtree. `KeyManager::advance_keys_to`, which looped an
`advance_preparation()` that had already become a no-op, is now
`prepare_keys_for`; an unsignable slot is named as exhaustion where it is reached.

## Testing

Every `ethlambda-crypto` test passes against the new backend under
`--include-ignored` (13 tests): the Type-2 round trip (aggregate two Type-1s,
merge, verify, narrow back to one claim, verify that), the three negative binding
tests, the arena path, and the XMSS sign/verify round trip. Two new unit tests pin
the signer-set construction every decode depends on.

`make lint` is clean. Fixture-driven spec tests still fail exactly as they did on
the tracking branch before this bump (122 forkchoice, 73 stf, 7 ssz, 3 signature),
all at deserialization: leanSpec's latest released fixtures are still on the
52-byte scheme, so they cannot load until leanSpec migrates and republishes.

## Not in scope

- **SPHINCS.** leanVM now aggregates SPHINCS alongside XMSS; every call site here
  passes an empty SPHINCS list.
- **The double deferred-claim recompute.** leanVM's decode runs
  `DeferredClaim::recompute`, and `verify()` repeats it to compare, so
  decode-then-verify pays that pass twice. Left upstream.
- **Fixture regeneration**, which needs leanSpec on leanVM `main` first.

## Carryless multiply on aarch64

`aarch64-unknown-linux-gnu` defaults to `neon` alone, and LLVM gates PMULL behind
the `aes` feature. leanVM main's binary-field tower multiplies via carryless
multiply, so a Linux container on an Apple-silicon host was the one build with no
hardware support for the prover's inner loop, and fell back to scalar code. The
x86 entry in `.cargo/config.toml` has had `+pclmulqdq` all along; there was no
aarch64 entry, because the devnet hosts are x86. `aarch64-apple-darwin` defaults
to a CPU that already has `aes`, which is why a native `cargo test` never showed
it.

The old Poseidon over KoalaBear prover worked in a 31-bit prime field and never
issued a carryless multiply at all, so this only became load-bearing here.
Measured on an M4 Max, one interval-2 aggregation of three XMSS signatures in a
three-node local devnet:

| build | aggregation |
| --- | --- |
| before | 27 s, then 56 s, then 83 s as sessions queued |
| after | 201-232 ms p50 |

Before, no node ever finished a block: a single uncontended node took 60.6 s to
build a zero-attestation block, against a 4 s slot.

## `ethlambda keygen`

Writes the key set and manifest `--hash-sig-keys-dir` reads, in the layout
`hash-sig-cli generate` produces, so `generate-genesis.sh` can call the client
instead of a separately pinned image.

```bash
ethlambda keygen --output-dir keys                       # one validator's pair
ethlambda keygen --num-validators 3 --output-dir keys    # a genesis
```

This is the same trap as the section above, one level up. An XMSS key file is
only usable by a client built against the same scheme, and no file size changes
when the scheme does: the BLAKE2s rewrite kept the public key at 32 SSZ bytes and
the secret key in postcard, so a key set from the wrong revision satisfies every
check a genesis generator makes and fails only once a signature is verified.
That reads as a consensus bug rather than a provisioning one, and it is what made
this devnet's first genesis unusable.

Generating in the client removes the second pin: the keys come from the same
`ethlambda-crypto` types the node loads them with, through two new constructors
on `ValidatorSecretKey`, so the generator and the loader cannot disagree about
the format. The manifest gains `leanvm_rev`, resolved from `Cargo.lock` at build
time; every other field is hardcoded or derived from parameters that came through
the rewrite unchanged, so `key_scheme` labels both formats identically.

`--force` gates overwriting an existing key set, because a key is one-time-use
material: replacing a set validators are still signing with makes each of them
sign twice at one slot, under a key someone else now holds.

**One caller-visible break:** there is no `--export-format`. `hash-sig-cli` took
one to choose between `ssz` and a `both` that additionally wrote serde-JSON dumps
its own help called legacy. Nothing reads those, so there is one output form and
no flag to pick it, and `generate-genesis.sh` has to drop the
`--export-format ssz` argument to call this.

## Devnet verification

Three ethlambda nodes, one subnet, one aggregator, on a genesis generated by
`ethlambda keygen`:

| | |
| --- | --- |
| Chain | 22 blocks, finalized slot 28 / justified 29 / head 31, identical on all three nodes |
| Finality | advancing every slot from slot 9 on |
| Aggregation | n=126, p50 201 ms, p95 350 ms, max 493 ms |
| Type-2 block-proof verify | p50 135 ms, max 270 ms |
| WARN / ERROR | 0 across all three nodes |

That exercises the whole migrated path with real proofs: keygen, XMSS sign,
interval-2 aggregate, proposer Type-1 wrap, Type-2 merge, gossip, decode against
rebuilt `SignerSet`s, verify.


.cargo/config.toml
```diff
@@ -1,9 +1,29 @@
...
+# scalar bit-twiddling: `+pclmulqdq` on x86, and on aarch64 the `aes` feature,
+# which is what LLVM gates PMULL behind.
+#
+# The aarch64 entry is not symmetry for its own sake. `aarch64-unknown-linux-gnu`
+# defaults to `neon` alone, while `aarch64-apple-darwin` defaults to a CPU that
+# already has `aes`, so a Linux container on an Apple-silicon host was the one
+# build with no carryless multiply. Measured on an M4 Max, one aggregation proof
+# went from ~30 s without this to well under a second with it, which is the
+# difference between a local devnet that cannot fill a slot and one that can.
+
```
MegaRedHand:
```suggestion
```

PR #607, opened 2026-09-04
build: add a 14-day publish-age cooldown to dependency resolution

pablodeymo:

Ports Commit-Boost/commit-boost-client#492. Cargo's unstable `min-publish-age` (rust-lang/cargo#17009) lets the resolver skip crate versions published less than N days ago, a cooldown against freshly compromised releases. This enables it at 14 days, adds the Makefile path for resolving under it, and makes every build `--locked` so stable cargo cannot silently re-resolve around the policy.

The feature is nightly-only; stable 1.97.1 ignores the tables silently. Only the pinned `nightly-2026-06-21` resolver enforces the cooldown, so `--locked` everywhere is what makes an unresolved manifest change fail loudly instead of resolving on stable.

## What Changed

| File | Change |
|------|--------|
| `.cargo/config.toml` | `[unstable] min-publish-age = true`, `[registry] global-min-publish-age = "14 days"` |
| `Makefile` | `make update` (resolve under the cooldown), `make update-allow PACKAGE= VERSION=` (escape hatch, bypasses the whole resolution), `make cooldown-check` (advisory). `make lint`/`make test` now `--locked` |
| `.github/workflows/ci.yml` | `cargo check`/`clippy` run `--locked`; new `cooldown` job annotates the run with a warning, never fails |
| `Dockerfile` | `cargo chef cook` honors `$LOCKED` like the final build; the shadow variant still sets it empty |
| `CONTRIBUTING.md`, `CLAUDE.md` | Document the workflow |

- No manifest or lockfile change. Locked builds are byte-for-byte what they were.
- A `cargo update --workspace --dry-run` does **not** flag too-new locked crates, so the check has to be the full `cargo update --dry-run -Z min-publish-age`, grepping `Downgrading|is too new`. That run also refreshes git dependencies (leanSig, leanVM, ethrex, rust-libp2p) to their branch heads with no cooldown; git revs have no publish age and must be reviewed by hand.
- **Known state today:** a full re-resolution under any window of 8+ days fails with a misleading `rand ^0.10` conflict. Real cause: rand 0.10 needs chacha20 0.10, whose 0.10.0/0.10.1 are yanked and whose only live release (0.10.2) was published 2026-08-27. `make update` errors and the CI job emits a "probe failed" warning until 2026-09-10; `make update-allow` covers an urgent bump before then. The committed lockfile already pins the yanked chacha20 0.10.0, which any future re-resolution will move.

## Tests Added / Run

- `make lint` passes with `--locked`.
- Stable `cargo check --locked` and `cargo update --workspace --dry-run --locked` with the new tables: no warnings, no changes.
- Temporarily pinned `smallvec 1.16.0` (3 days old) and ran the 7-day dry-run: ` Downgrading smallvec v1.16.0 -> v1.15.2 (...)`, confirming the grep pattern. Lockfile restored identical.
- Ran the CI step script locally and `make cooldown-check`: both report the chacha20-driven probe failure as a warning with exit 0.

## Related Issues / PRs

- Related to Commit-Boost/commit-boost-client#492

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean (no Rust changes)
- [ ] Ran `make test` — not run; no Rust or fixture changes, only the `--locked` flag was added

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,11 +85,10 @@ jobs:
             | jq -e '.schema_version == 1 and (.samples | length == 3)'
 
   # Stable cargo ignores the publish-age cooldown in .cargo/config.toml, so the
-  # lockfile can pin too-young crates (or deliberately via the
-  # CARGO_RESOLVER_INCOMPATIBLE_PUBLISH_AGE=allow escape hatch). Fail the build
-  # when either lockfile does. Infrastructure failures (toolchain download, a
-  # resolution that fails for reasons unrelated to age, e.g. a yanked upstream
-  # crate) only warn: they are outside the PR's control and would block every PR.
+  # lockfile can pin too-young crates. Fail the build when either lockfile does.
+  # Infrastructure failures (toolchain download, a resolution that fails for
+  # reasons unrelated to age, e.g. a yanked upstream crate) only warn: they are
+  # outside the PR's control and would block every PR.
   cooldown:
     name: Dependency cooldown
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,14 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
 
+      # `--locked` so the committed Cargo.lock is actually enforced (see the
+      # tooling steps below), and so a dependency bump cannot sneak past the
+      # publish-age cooldown in .cargo/config.toml by resolving on stable here.
       - name: Cargo check
-        run: cargo check --workspace --all-targets
+        run: cargo check --locked --workspace --all-targets
 
       - name: Clippy
-        run: cargo clippy --workspace --all-targets -- -D warnings
+        run: cargo clippy --locked --workspace --all-targets -- -D warnings
 
       # tooling/event-monitor declares its own [workspace] table, so every step
       # above stops at the root workspace members and never reaches it. Its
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -43,11 +43,14 @@ jobs:
       - name: Check formatting
         run: cargo fmt --all -- --check
 
+      # `--locked` so the committed Cargo.lock is actually enforced (see the
+      # tooling steps below), and so a dependency bump cannot sneak past the
+      # publish-age cooldown in .cargo/config.toml by resolving on stable here.
```
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -83,3 +86,31 @@ jobs:
           cargo run --profile release-fast --bin ethlambda -- benchmark synthetic --mock-crypto \
             --num-validators 4 --warmup-slots 4 --iterations 3 --format json \
             | jq -e '.schema_version == 1 and (.samples | length == 3)'
+
+  # Stable cargo ignores the publish-age cooldown in .cargo/config.toml, so the
+  # lockfile can pin too-young crates (or deliberately via `make update-allow`).
+  # Surface them as annotations; this job never fails the build.
```
MegaRedHand: If this doesn't fail the build it's useless

```suggestion
  # Stable cargo ignores the publish-age cooldown in .cargo/config.toml, so the
  # lockfile can pin too-young crates (or deliberately via `make update-allow`).
  # Surface them as annotations and fail the build.
```

Makefile
```diff
@@ -1,18 +1,55 @@
...
+
+# Escape hatch for an urgent update to a version younger than the cooldown,
+# e.g. `make update-allow PACKAGE=h2 VERSION=0.4.16`. The bypass applies to the
+# WHOLE resolution of this invocation (transitive picks included), so review the
+# resulting lockfile diff.
+update-allow: ## 🚨 Update one crate to a version younger than the cooldown (PACKAGE=... VERSION=...)
+	@test -n "$(PACKAGE)" -a -n "$(VERSION)" || { echo "usage: make update-allow PACKAGE=<crate> VERSION=<version>" >&2; exit 1; }
+	rustup toolchain install $(RESOLVER_TOOLCHAIN) --profile minimal > /dev/null 2>&1 && \
+	CARGO_RESOLVER_INCOMPATIBLE_PUBLISH_AGE=allow \
+	cargo +$(RESOLVER_TOOLCHAIN) update -Z min-publish-age -p $(PACKAGE) --precise $(VERSION)
```
MegaRedHand: I don't think this is necessary

Makefile
```diff
@@ -1,18 +1,55 @@
...
 
 help: ## 📚 Show help for each of the Makefile recipes
 	@grep -E '^[a-zA-Z0-9_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
 
 fmt: ## 🎨 Format all code using rustfmt
 	cargo fmt --all
 
+# `--locked` so the committed Cargo.lock is actually enforced: without it cargo
+# silently resolves and rewrites the lockfile, bypassing the publish-age cooldown
+# (see `update` below).
```
MegaRedHand:
```suggestion
```

MegaRedHand (approved):

.github/workflows/ci.yml
pablodeymo: Done: the job now exits 1 on a too-young pin in either lockfile, and I narrowed the grep to `Downgrading .*published` because bare downgrades for MSRV or tightened requirements show up too (event-monitor has three today). Infrastructure failures still only warn, so the root probe, which currently fails on yanked chacha20 releases upstream, doesn't block every PR; it starts gating on 2026-09-10 when chacha20 0.10.2 ages out. One consequence: an `allow`-bypassed bump keeps CI red until it ages, which I think is the right trade but wanted to flag.

Makefile
pablodeymo: Removed. The escape hatch is `CARGO_RESOLVER_INCOMPATIBLE_PUBLISH_AGE=allow make update UPDATE_ARGS="-p <crate> --precise <ver>"`, same as in the ethrex PR.

CLAUDE.md
```diff
@@ -90,6 +90,9 @@ make test                                    # All tests + forkchoice spec tests
 ### Common Operations
 ```bash
 rm -rf leanSpec && make leanSpec/fixtures                # Download latest released test fixtures
+make update UPDATE_ARGS="-p <crate>"                     # Bump deps under the 14-day publish-age cooldown (nightly resolver)
+CARGO_RESOLVER_INCOMPATIBLE_PUBLISH_AGE=allow make update UPDATE_ARGS="-p <crate> --precise <ver>"  # Escape hatch; CI stays red until it ages
```
MegaRedHand: We don't need the escape hatch

```suggestion
```

MegaRedHand (approved):

PR #608, opened 2026-09-04
fix(p2p): never leave a block root pending without an outcome

MegaRedHand:

A root sits in `pending_root_requests` from the moment a `BlocksByRoot` request is sent until something retires it, and the `FetchBlock` handler deduplicates every later fetch against that table. A root that is never retired is unreachable for the life of the process: the chain actor can ask for it any number of times, no request reaches the wire, and nothing is logged above `debug` to say so.

Two response paths ended an attempt without retiring the root. Neither needs a race; both are reachable with an ordinary peer on the other end.

**1. An error response for a root request.** The arm removed the id from `outbound_requests`, put it straight back, and returned:

```rust
Some(request @ PendingRequestKind::Root(_)) => {
    server.outbound_requests.insert(request_id, request);
}
```

`handle_fetch_failure` never reads `outbound_requests`, and an error response completes the exchange, so no `OutboundFailure` follows to pick the root up. Our own by-root responder never sends an error code, but other clients do.

**2. A response whose blocks all fail the requested-root check.** Every block hit `continue`, so `pending_root_requests.remove` was never reached and the function returned having retired nothing. `outbound_requests` had already been drained at the dispatch site, so no later event could recover it either.

Both leave the root pending with no retry scheduled, which is why the symptom is an indefinite silence rather than the `error!("...giving up")` you would expect from a fetch that ran out of attempts.

**A leak alongside them.** That re-insert appears twice on `main`: in the error arm above, and in the empty-response path, where a comment states the premise outright — *"Re-insert so failure handling can find it"*. `handle_fetch_failure` takes `root` and `peer` as arguments and has never read `outbound_requests`, so it bought nothing either time. It also leaked: the dispatch site has already removed the id, the exchange is over, and no later libp2p event carries that id, so nothing removes it a second time. Every empty or error `BlocksByRoot` response left one dead entry in the map for the life of the process. The empty-response path did fail the root correctly, so the leak was its only symptom.

## What Changed

All in `crates/net/p2p/src/req_resp/handlers.rs`.

- The error-response arm and the no-matching-block path both route into `handle_fetch_failure`, which is the single funnel that retires an attempt.
- The response path finds the one block that can answer a single-root request instead of looping. It subsumes the old empty-response special case, which is why the `request_id` parameter and the leaking re-insert it existed for are both gone.

- **The invariant:** an attempt only ends by retiring the root, either by delivering the block or by charging a failure that schedules a retry or gives up. Previously two paths ended an attempt while leaving the entry behind.
- Retry policy is unchanged: `MAX_FETCH_RETRIES` attempts with doubling backoff, then the entry is dropped and an `error!` is logged. Dropping the entry means the P2P side forgets the root, not that it is blacklisted; the chain starts a fresh cycle the next time an orphan referencing that root arrives.
- A failure for a root that is no longer pending stays a no-op, so a late or duplicate event cannot resurrect a root that already succeeded.
- `outbound_requests` is leak-free in the ordinary case: two insert sites at send, three remove sites covering every terminal event, and a retry inserts a fresh id only after the failure that triggered it removed the old one. That bound rests on every request eventually producing a terminal event, the same assumption behind dropping the watchdog below; if it is ever wrong, this map leaks alongside the stranded root.
- **Behavior change worth reviewing:** a peer that answers with an error code, or with blocks we did not ask for, now costs an attempt and triggers a backoff retry against a different peer, where before it silently ended the fetch.

## Tests Added / Run

No new tests. Both funnels sit behind an async handler taking `&mut P2PServer` and a live `Context`, which the crate has no harness for; earlier revisions of this branch reached the retry accounting by extracting it into a free function, and that indirection was not worth keeping for its own sake. Worth knowing that the existing suite did not catch either hole.

```
make fmt && make lint && make test
```

All green: 658 tests, 0 failures.

## A third path that was considered and rejected

Earlier revisions added a watchdog for a request libp2p never reports an outcome for, because `libp2p-request-response` drops a request whose dial the swarm rejects with `DialError::DialPeerConditionFalse` (upstream [PR 6000](https://github.com/libp2p/rust-libp2p/pull/6000), in our 0.30.0 pin). That case is not reachable here, so it was removed in 3f67c807:

- Rejection needs a concurrent dial or an existing connection. A concurrent dial resolves either way and both outcomes drain the peer's parked requests, via `preload_new_handler` or `on_dial_failure`. That is the invariant PR 6000 relies on.
- An existing connection would strand the request, but it requires the behaviour's connection map to disagree with the pool. `swarm_adapter.rs` runs one task selecting between swarm events and commands, so a command never observes a half-updated swarm.

The reported symptom agrees. The request-size histogram last moved an hour before the stall, so the stranding request did reach the wire; a dial-path strand never reaches the codec and would not have moved it at all.

## Not in scope

- **`BlocksByRange` has the same shape**, and a worse failure mode: a stranded range request leaves `range_sync_state.in_flight = true` forever, wedging sync rather than losing one block.
- **The beacon branches carry the same bug** in `crates/net/p2p/src/sync.rs` (`record_root_fetch_failure`, keyed by `RootKey`).

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean

### DIFF
--- a/crates/net/p2p/src/req_resp/handlers.rs
+++ b/crates/net/p2p/src/req_resp/handlers.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use ethlambda_network_api::BlockSource;
 use ethlambda_storage::Store;
@@ -19,7 +19,7 @@ use super::{
 };
 use crate::{
     BACKOFF_MULTIPLIER, INITIAL_BACKOFF_MS, MAX_FETCH_RETRIES, MAX_SYNC_RANGE, P2PServer,
-    PendingRequest, PendingRequestKind, RangeSyncState, p2p_protocol,
+    PendingRequest, PendingRequestKind, ROOT_FETCH_WATCHDOG, RangeSyncState, p2p_protocol,
     req_resp::RequestedBlockRoots,
 };
 
@@ -80,10 +80,8 @@ pub async fn handle_req_resp_message(
                                     .await;
                                 }
                                 Some(PendingRequestKind::Root(root)) => {
-                                    handle_blocks_by_root_response(
-                                        server, blocks, peer, request_id, root, ctx,
-                                    )
-                                    .await;
+                                    handle_blocks_by_root_response(server, blocks, peer, root, ctx)
+                                        .await;
                                 }
                                 None => {
                                     debug!(%peer, ?request_id, "Received blocks response for unknown request_id");
@@ -99,8 +97,12 @@ pub async fn handle_req_resp_message(
                             Some(PendingRequestKind::Range { .. }) => {
                                 fail_range_request(server, &peer);
                             }
-                            Some(request @ PendingRequestKind::Root(_)) => {
-                                server.outbound_requests.insert(request_id, request);
+                            Some(PendingRequestKind::Root(root)) => {
+                                // An error response completes the exchange, so
+                                // no `OutboundFailure` follows to retire the
+                                // root. Fail it here or it stays pending
+                                // forever and deduplicates every later fetch.
+                                handle_fetch_failure(server, root, peer, ctx).await;
                             }
                             None => {}
                         }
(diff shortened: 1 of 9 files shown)


crates/net/p2p/src/req_resp/handlers.rs
```diff
@@ -287,44 +289,40 @@ async fn handle_blocks_by_root_response(
...
+    let received = blocks.len();
+    trace!(%peer, count = received, "Received BlocksByRoot response");
+
+    // Requests carry a single root, so at most one block can answer one and
+    // anything else the peer sent is unsolicited.
+    //
+    // A response that answers nothing is a failed attempt, whether it was empty
+    // or carried only blocks we never asked for. Treating the latter as a
+    // no-op would leave the root pending forever, and the deduplication in the
+    // `FetchBlock` handler would swallow every later attempt to fetch it.
```
MegaRedHand:
```suggestion
    // Requests carry a single root, so at most one block can answer one and
    // anything else the peer sent is unsolicited.
```

pablodeymo (approved):

pablodeymo (approved):

PR #610, opened 2026-09-08
perf(blockchain): bound attestation ancestry walks by fork depth

MegaRedHand:
## Problem

`validate_attestation_data` makes three unbounded parent-chain walks per attestation, and every step of a walk is a fresh read view, a RocksDB point get, and an SSZ decode (`get_block_header` has no cache, unlike `get_state`).

A walk's length is the slot distance between its two checkpoints, so it grows without limit as finalization falls behind. On a node with finalized at slot 928 and the head near 64000:

| call site | walk | steps |
|---|---|---|
| source ← target | 7984 → ~64000 | ~56,000 |
| target ← head | ~0, early return | ~0 |
| finalized ← head | 928 → ~64000 | ~63,000 |

That is ~119,000 header reads for a single attestation, and it runs once per gossip attestation, once per aggregate, and for the node's own votes via self-delivery. At a few hundred attestations per slot it exceeds the slot duration by orders of magnitude.

The result is a feedback loop rather than merely slow validation: validation slows, attestations are dropped, finality falls further behind, and the walks grow longer still.

## Fix

`BlockRoots` already indexes the canonical chain by slot, written atomically with the head in `update_checkpoints`, so it always describes the branch ending at the stored head (it can lag a freshly imported block that fork choice has not selected, but never runs ahead).

Since the canonical chain is a single path, a walk that reaches any canonical block **above** the ancestor's slot can stop there: the ancestor lies on that chain exactly when the index names it at its own slot.

```
canonical ────●────●────●────●────●────●  head
               \
                ○────○  fork tip = descendant
                └──┬──┘
             walk only this far (fork depth)
                   ↓
      rejoin point is canonical and above ancestor.slot
                   ↓
    ancestor on chain? → one index lookup, O(1)
```

Cost is now bounded by how deep the descendant's branch has forked (a handful of slots) rather than by the distance to the ancestor. The `finalized ← head` check settles with **no walk at all**, since the finalized checkpoint is canonical by construction.

New storage API, the only addition:

```rust
pub fn canonical_root_at_slot(&self, slot: u64) -> Result<Option<H256>, Error>
```

The judgment about how far the index can be trusted stays inside the `Store`, so no caller can widen it.

## Two ways the short circuit can be wrong

Both are load-bearing and both have a regression test.

**1. An index miss must not mean "not an ancestor."** A `None` is either a slot the canonical chain skipped or a slot below this store's anchor, and the index cannot tell those apart. A miss therefore leaves the full walk as the answer. This is also what keeps the change correct if historical backfill ever lands.

**2. The short circuit must not fire at or below the ancestor's slot.** I had this wrong at first, and the entire existing suite passed:

```
canonical:  g(0) ── a(1) ── c(2)          index = {0:g, 1:a, 2:c}
fork:       g(0) ─────────────── d(3)      d's parent is g, skipping slots 1-2

is_ancestor(a@1, d@3):
  iter1  d@3   index[3]=None        → no fire; 3 > 1 → step to parent g@0
  iter2  g@0   index[0]=Some(g) ✓   → FIRES → returns true
                                              ↑ wrong: `a` is not on d's chain
```

A branch that skips the ancestor's slot and rejoins canonical below it does not contain the ancestor. The slot guards have to settle the walk before the index is consulted. `checkpoint_is_ancestor_rejects_ancestor_skipped_by_fork_branch` covers it — verified to fail on the broken ordering and pass on the fixed one.

## Check ordering left alone

I tried hoisting the four storage-free checks (source > target, head < target, slot < head.slot, and the time check) ahead of the three block lookups, so malformed and future-dated votes would cost no storage reads. It works, but it changes which reason a vote failing several checks at once is rejected for, and five leanSpec gossip-validation fixtures pin those reasons:

```
expected HEAD_SLOT_MISMATCH    got ATTESTATION_TOO_FAR_IN_FUTURE
expected TARGET_SLOT_MISMATCH  got ATTESTATION_SLOT_BEFORE_HEAD
expected UNKNOWN_SOURCE        got a moved check
```

Rejection reasons are cross-client observable, so the ordering is not ours to optimize. Reverted, with `validate_attestation_reports_availability_before_cheaper_failures` and a note on the function recording the constraint so it is not mistaken for an oversight later.

## Testing

- The four pre-existing ancestry tests are untouched and pass.
- Six added: canonical vote end-to-end, off-chain ancestor rejected via the index, index-miss fallback, fork branch skipping the ancestor's slot, block imported before `update_head` ran, and the rejection-precedence fence.
- `make lint` clean; `make test` green, including all 122 forkchoice spec fixtures.

## Not in scope

`get_live_chain` is rebuilt from a full `LiveChain` iteration on every head update, so at a 63k-slot finality gap it materializes a ~63,000-entry `HashMap` per call. Same underlying cause, separate hot path.

### DIFF
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -1706,47 +1706,6 @@ mod tests {
         ));
     }
 
-    /// A vote that fails several checks at once must be rejected for the reason
-    /// the spec's expected-failure fixtures name, which is the availability
-    /// failure here even though the cheaper time check would also have caught it.
-    ///
-    /// Hoisting the storage-free checks ahead of the block lookups to save reads
-    /// breaks exactly this: it turns five leanSpec gossip-validation fixtures
-    /// (`test_attestation_head_slot_mismatch_rejected` and friends) from
-    /// `HEAD_SLOT_MISMATCH` / `TARGET_SLOT_MISMATCH` / `UNKNOWN_SOURCE` into
-    /// `ATTESTATION_TOO_FAR_IN_FUTURE` / `ATTESTATION_SLOT_BEFORE_HEAD`. The
-    /// rejection reason is cross-client observable, so the order is not ours to
-    /// optimize.
-    #[test]
-    fn validate_attestation_reports_availability_before_cheaper_failures() {
-        let store = new_test_store();
-        let unknown = H256([9u8; 32]);
-
-        // Far-future slot and unknown roots: the time check would fire, but the
-        // availability check comes first and owns the rejection.
-        let data = AttestationData {
-            slot: 1_000,
-            source: Checkpoint {
-                root: unknown,
-                slot: 998,
-            },
-            target: Checkpoint {
-                root: unknown,
-                slot: 999,
-            },
-            head: Checkpoint {
-                root: unknown,
-                slot: 1_000,
-            },
-        };
-
-        let result = validate_attestation_data(&store, &data);
-        assert!(
-            matches!(result, Err(StoreError::UnknownSourceBlock(root)) if root == unknown),
-            "Expected UnknownSourceBlock, got: {result:?}"
-        );
-    }
-
     /// leanSpec #833: a vote whose head sits on a sibling fork of the target
     /// must be rejected by gossip validation, even though every slot and
     /// availability check passes.
--- a/crates/blockchain/src/store.rs
+++ b/crates/blockchain/src/store.rs
@@ -245,10 +245,7 @@ fn checkpoint_is_ancestor(
 
 /// Validate incoming attestation before processing.
 ///
-/// Ensures the vote respects the basic laws of time and topology. The order is
-/// fixed by the spec's expected-failure fixtures, which pin which reason a vote
-/// failing several checks at once is rejected for, so these must not be
-/// reordered to make a cheaper check run first:
+/// Ensures the vote respects the basic laws of time and topology:
 ///     1. The blocks voted for must exist in our store.
 ///     2. A vote cannot span backwards in time (source > target).
 ///     3. The head must be at least as recent as source and target.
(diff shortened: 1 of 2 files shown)


crates/blockchain/src/store.rs
```diff
@@ -187,27 +193,62 @@ fn checkpoint_is_ancestor(
...
+    }
 }
 
 /// Validate incoming attestation before processing.
 ///
-/// Ensures the vote respects the basic laws of time and topology:
+/// Ensures the vote respects the basic laws of time and topology. The order is
+/// fixed by the spec's expected-failure fixtures, which pin which reason a vote
+/// failing several checks at once is rejected for, so these must not be
+/// reordered to make a cheaper check run first:
```
```suggestion
/// Ensures the vote respects the basic laws of time and topology:
```

crates/blockchain/src/store.rs
```diff
@@ -187,22 +193,54 @@ fn checkpoint_is_ancestor(
...
-    while let Some(current_header) = store
-        .get_block_header(&current_root)
-        .expect("parent block exists")
-    {
-        if current_header.slot == ancestor.slot {
+    // Resolve the ancestor against the canonical index once. `None` means the
+    // index has nothing to say about that slot, so the parent walk stays the only
+    // sound answer: a miss must never be read as "not an ancestor". That is what
+    // keeps this correct for a slot below the anchor this store was bootstrapped
+    // from, where the index simply has no entry to offer.
```
```suggestion
    // Resolve the ancestor against the canonical index once. `None` means the
    // index has nothing to say about that slot, so the parent walk stays the only
    // sound answer: a miss must never be read as "not an ancestor".
```

crates/blockchain/src/store.rs
```diff
@@ -1511,6 +1547,206 @@ mod tests {
...
+                slot: 1_000,
+            },
+        };
+
+        let result = validate_attestation_data(&store, &data);
+        assert!(
+            matches!(result, Err(StoreError::UnknownSourceBlock(root)) if root == unknown),
+            "Expected UnknownSourceBlock, got: {result:?}"
+        );
+    }
```
```suggestion
```

pablodeymo (approved):

PR #611, opened 2026-09-08
feat(cli): benchmark block building with real XMSS and leanVM crypto

pablodeymo:

`ethlambda benchmark synthetic` only ran with `--mock-crypto`, so it measured selection, compaction and the state transition but none of the signing or aggregation that dominates a real proposal. This makes real crypto the default and measures the proposer's seal.

Without `--mock-crypto` the run derives an attestation and a proposal XMSS key per validator from the seed (sized to the slots the run signs, so keygen costs seconds; `--key-cache <dir>` stores them for reruns), puts the real pubkeys in the synthetic genesis, has every validator sign each slot's attestation data through the production `KeyManager`, aggregates each participant group into a real leanVM type-1 proof, and seals the built block with `seal_block`. The measured span is build plus seal, exactly the node's `lean_block_building_time_seconds` boundary.

## What Changed

| File | Change |
|---|---|
| `crates/blockchain/src/block_builder.rs` | `seal_block` + `SealError`, lifted verbatim from `propose_block`; observes `sign_proposer`, `wrap_proposer`, `merge_type2` |
| `crates/blockchain/src/lib.rs` | `propose_block` calls `seal_block`, same logging and failure counter on error |
| `crates/blockchain/src/metrics.rs` | `BLOCK_PROPOSAL_SEAL_PHASES`; the histogram now spans the same work as `lean_block_building_time_seconds` |
| `crates/common/crypto/src/signature.rs` | `ValidatorSecretKey::generate_from_seed` (deterministic keygen for tests and benchmarks) and `to_bytes` |
| `bin/ethlambda/src/benchmark/keys.rs` | Seed-derived XMSS keys sized to the run, generated in parallel, optional `--key-cache <dir>` keyed by leansig rev/seed/index/window |
| `bin/ethlambda/src/benchmark/corpus.rs` | `CryptoMode::{Mock, Real}` owns everything the mode decides: real genesis pubkeys, real type-1 proofs signed through the production `KeyManager`, seal, import (verifying in real mode), expected phases |
| `bin/ethlambda/src/benchmark/mod.rs` | Drop the mock requirement; seal in the measured span; one code path for both modes |
| `bin/ethlambda/src/benchmark/report.rs` | `aggregate` and `import` columns (outside the span) |
| `docs/benchmarking.md`, `docs/metrics.md` | Real-mode docs, sample output, new phase labels |

- `propose_block` behavior is unchanged; the seal is the same code behind a function, now taking the aggregates by value so proof bytes move into the merge instead of being copied.
- Mock mode is unchanged and still what CI and `make bench` run; it skips the seal since there are no keys, so its report carries only the build phases.
- Real mode imports every built block through `on_block`, so an invalid proof fails the run instead of producing a report about invalid blocks.
- Same seed and parameters reproduce the same roots: keys are seed-derived and XMSS signing is deterministic.
- Node metric change: three new `phase` label values on `lean_block_proposal_attestation_build_phase_seconds`.

## Tests Added / Run

- `keys.rs`: determinism per seed and role; cache round-trip that signs through the `KeyManager`.
- `corpus.rs`: `#[ignore = "too slow"]` test that a seeded proof passes `verify_aggregated_signature` (passes locally).
- Real run, 2 validators, 1 warmup, 2 iterations: keys in 6.8s, ~0.1s per slot of type-1 aggregation, 0.5–1.3s type-2 merges, 20ms verified imports; a rerun loaded the keys from cache and produced identical block roots. `--enable-proposer-aggregation --proofs-per-data 2` exercised a real recursive compaction (`compact` = 0.56s).
- Mock CI smoke contract still holds (`schema_version == 1`, one sample per iteration, build phases only).

```
ethlambda benchmark synthetic --num-validators 2 --warmup-slots 1 --iterations 2 --key-cache /tmp/bench-keys
```

## Related Issues / PRs

- Follows up #595 / #596 (benchmark harness and report), which left real crypto as a documented limitation.

## ✅ Verification Checklist

- [x] Ran `make fmt` — clean
- [x] Ran `make test` (`cargo test --workspace --profile release-fast`) — all passing on the latest leanSpec fixtures


MegaRedHand (approved):
LGTM

PR #613, opened 2026-09-10
feat(aggregation): score recursive aggregation through a per-aggregator subnet window

MegaRedHand:
## Summary

Every aggregator on the network currently does the same aggregation work. In a healthy slot all validators vote for the same head, so there is one hot `AttestationData`; candidates are scored deterministically from head state, and `select_proofs_greedily` picks the two highest-coverage children from a pool every aggregator sees, since aggregates gossip on one global topic. So all aggregators select the same two children and produce the same merged proof, and every copy of that leanVM work past the first is wasted.

If every aggregator anchors on the pool's best proof `r0`, then aggregator `i` publishes `out_i = r0 ∪ x_i`, and `out_i ∪ out_j = out_i ∪ x_j`. Merging two published proofs gains nothing over merging one of them with a raw pool proof.

This gives each aggregator a **duty subnet** and a **window** of subnets starting there, used as a scoring lens on child selection, so different aggregators merge different children.

## Design

- **Window**: the contiguous cyclic run of subnets `{s, s+1, ...}` an aggregator is responsible for, starting at its duty subnet.
- **Scoring lens, not a filter**: a child is valued by the validators it newly covers whose subnet is inside the window. A proof straddling the boundary stays usable for its in-window part; one lying wholly outside scores zero. A selected child still contributes *all* of its participants to `covered`, which keeps the marginal-coverage score honest across greedy rounds.
- **Width from the anchor**: `min(2 * reach(anchor), C)`, where the *anchor* is the largest-coverage proof in the candidate's pool that touches the aggregator's own duty subnet, and a proof's *reach* is how many distinct subnets it touches. Coverage rather than reach picks the anchor, so a sparse proof holding one validator in each of many subnets no longer sets the width for everybody. Requiring the anchor to touch the duty subnet makes "no anchor" mean "no peer has covered my subnet", which is exactly when this node's raw signatures are irreplaceable: the window then sits at its narrowest and the aggregator works on those instead of merging proofs it cannot add to. Deriving the width rather than choosing it is essential either way, since windows nest and "widest viable window" would collapse to the full committee set for everyone.
- **Duty subnet**: the first `--aggregate-subnet-ids` value, else the lowest subscribed subnet, else 0. Logged at startup so a collision is diagnosable. The node refuses to start on an `--aggregate-subnet-ids` value at or above the committee count: subscriptions use the raw id while the window reduces it, so an out-of-range value would subscribe to a topic no validator publishes on and aggregate as a subnet the node does not listen to.
- **`--skip-redundant-aggregation`** (opt-in): an aggregator sits out any candidate whose derived width it does not own this slot (`duty_subnet % w == slot % w`), and the freed job goes to the next-best `AttestationData` rather than to a narrower merge of the same one. Ownership rotates with the slot, so no duty subnet is permanently the one sitting out, and width 1 is owned by everyone, so a candidate with no anchor on this node's subnet is never skipped.
  - **Deployment precondition**: every subnet below the committee count needs an aggregator holding it as its duty subnet. Ownership rotates, but a width can have *no* owner in a slot on a sparser placement even with every node healthy: at duty subnets `{0, 2}` and committee count 4, nothing owns width 4 in an odd slot, and the fallback is off, so that merge level is dropped for the slot. Leave the flag unset on a placement that does not cover every subnet.

## Safety

- **No consensus impact.** Nothing in attestation processing, block building, or fork choice is touched. A windowed aggregate binds exactly `raw_ids ∪ accepted_child_ids` and its bits derive from that same set, so it is valid with fewer participants: less fork-choice weight, never wrong weight.
- **No coverage regression.** A window is contiguous but the pool need not be contiguous in subnet space, so a strided aggregator placement could leave a window holding one proof and drop a merge the unwindowed selection would have made. When a windowed selection is not viable it falls back to the full committee set, so the window can only improve on the old selection, never regress below it. The fallback is disabled under `--skip-redundant-aggregation`: every width below the committee count has several owners, so retrying there would rebuild exactly the duplication the flag buys away.
- **`attestation_committee_count = 1` is a provable no-op**: the window contains every validator, so in-window coverage equals total coverage and the break condition is identical to before, tie-break order included.
- **Mixed-network safe.** No wire-format, topic, or fork-digest change. A `main` node's wide proof can raise a branch node's anchor reach where it touches that node's duty subnet, which opens its window, so a partial rollout makes the feature weaker rather than inconsistent.

## Known trade-off

Because the anchor must touch the duty subnet, the derived width is no longer uniform across the network. Two aggregators reading one lopsided pool can derive different widths, so their windows nest rather than tile. That costs a round of climbing, not correctness, and it is the direct price of tying the window to work the aggregator can actually contribute to.

## Cadence, in practice

There is one aggregation session per slot, and the current slot's pool is empty at snapshot time since produced aggregates are held until the interval-2 boundary. So the window bites on the stale candidate, and a data root gets about one windowed merge rather than a multi-round climb. The widening matters across slots for a data root that stays live.

That emptiness is a timing expectation, not an invariant. A peer's aggregate for the current slot landing before this node's snapshot, under clock skew or a session started early via `EarlyAggregationCheck`, gives the current-slot candidate an anchor and a width of 2; under `--skip-redundant-aggregation` the duty subnets that do not own width 2 then sit that slot out and their raw signatures miss the next block. On the intended topology (distinct duty subnets, distinct subscriptions) a round-one peer proof never touches this node's subnet, so it holds anyway. Overlapping subscriptions are where it fails.

## Metrics

- `lean_aggregation_window_width` (Histogram): width derived per candidate. Climbs from 1 as the anchor climbs; pinned at the committee count means the window no longer restricts selection. Stuck at 1 while the network is aggregating means the pool holds nothing on this node's duty subnet, so it is only aggregating its own raw signatures.
- `lean_aggregation_skipped_redundant_total`: candidates handed to another duty subnet by the redundancy-skipping rotation. Only increments with the flag on.
- `lean_aggregation_window_fallback_total`: merges the window would have dropped, recovered by retrying selection at the full committee set. Counts recoveries, not attempts: a candidate no window could have made viable, a lone raw signature or a single-proof group, never reaches the retry. A persistently rising value therefore means the aggregator placement is too sparse for the committee count. Stays flat entirely under `--skip-redundant-aggregation`.

## Test Plan

- [x] `cargo test --workspace --profile release-fast`: 695 tests pass
- [x] `make lint` clean, `make fmt` clean, `make docs` builds
- [x] Four-aggregator reduction pinned end to end: round 1 produces four distinct proofs, round 2 reaches the full validator set
- [x] Mid-climb widening covered at committee count 8, where the window grows but stays a proper subset
- [x] The strided-placement regression has a test that fails without the fallback
- [x] Single-committee no-op pinned
- [x] Anchor selection pinned: a duty subnet the pool does not reach gets width 1, a sparse wide proof loses to a denser narrow one, a coverage tie falls to reach so pool order does not matter
- [x] A skipped candidate hands its job budget to the next-best `AttestationData`
- [x] The full-width retry is skipped where it is provably a no-op: a full-width window, an empty proof pool, a structureless committee
- [x] An `--aggregate-subnet-ids` value at or above the committee count is refused at startup, past the first id too
- [ ] Devnet: run all-ethlambda with `attestation_committee_count = 4` and aggregators on **distinct** duty subnets, confirm `lean_aggregation_window_width` climbs rather than pinning at 4, fleet-wide aggregation CPU drops against a control, and finality is unaffected
- [ ] Devnet: confirm `lean_aggregation_window_fallback_total` stays flat on the intended placement

## Notes for review

- The default proposer path (`keep_best_proof_per_data`) keeps one proof per `AttestationData` and drops the rest. Aggregators now emit proofs with distinct coverage rather than near-identical ones, so that path may drop more useful coverage than before. Bounded, since the window is sized to fit two children of the anchor's current reach, so a windowed proof is the same size as before; `--enable-proposer-aggregation` removes the exposure entirely. Declared out of scope here but worth measuring.
- With `--skip-redundant-aggregation` set and no explicit `--aggregate-subnet-ids`, every aggregator on a subnet-spanning topology derives duty subnet 0 and sits out in lockstep instead of taking turns. The node warns at startup when that combination is configured.
- The raw-signature guarantee is structural rather than a hard floor: "no anchor" means no peer covered our subnet. It is not airtight when two aggregators share a subnet and hold different slices of it, which no supported topology does today.


crates/blockchain/src/aggregation.rs
```diff
@@ -494,11 +637,56 @@ fn resolve_job(
...
+        validator_sigs,
+        new_proofs,
+        known_proofs,
+        validators,
+        window,
+    );
+    if primary.is_some() || config.skip_redundant {
+        return primary;
+    }
+    metrics::inc_aggregation_window_fallback();
```
pablodeymo: This retry runs, and the counter increments, whenever the first `resolve_job` returns `None`, but that also happens for reasons the window had nothing to do with: a gossip group holding a single raw signature and no children (every minority or late vote), or a candidate whose derived width already equals the committee count, where `SubnetWindow::full` scores identically and cannot recover anything.

So `lean_aggregation_window_fallback_total` will climb on any healthy deployment, which contradicts the metrics doc ("should stay at or near zero on a well-tiled deployment"), and the second selection pass runs on every non-viable candidate.

Suggestion:
- return early when `window.width() == config.committee_count` (the retry is a no-op there), and
- increment the counter only when the full-width retry actually yields a job, since that is the event the metric claims to count.

bin/ethlambda/src/main.rs
```diff
@@ -279,12 +279,34 @@ async fn run_node(options: NodeOptions) -> eyre::Result<()> {
     // receiver-count guard in `emit` makes every emission a no-op.
     let events = EventBus::default();
 
+    let aggregation_duty_subnet = resolve_aggregation_duty_subnet(
```
pablodeymo, line 288: The duty subnet is later reduced modulo `committee_count` in `window_for_candidate`, but the subscription set built at line 263 keeps the raw `--aggregate-subnet-ids` values and the swarm subscribes to those raw topics.

With `--attestation-committee-count 4 --aggregate-subnet-ids 5` this node subscribes to topic 5, which no validator publishes on, and aggregates as duty subnet 1, which it does not listen to. The startup log prints 5, so the mismatch is invisible to the operator.

Rather than reducing silently, reject any `--aggregate-subnet-ids` value at or above the committee count here, next to the existing `attestation_committee_count >= 1` check. That fixes the pre-existing dead subscription too, and the modulo in `window_for_candidate` plus `window_for_candidate_reduces_an_out_of_range_duty_subnet` can then go.

bin/ethlambda/src/cli.rs
```diff
@@ -83,8 +83,35 @@ pub(crate) struct NodeOptions {
...
+    /// aggregator skips a candidate whose width it does not own in the current
+    /// slot and spends that job on the next-best attestation data instead. The
+    /// owner rotates with the slot, so no node is permanently the one sitting
+    /// out, and the narrowest width is owned by everyone, so a candidate whose
+    /// pool holds nothing on this node's subnet, which is the raw-signature
+    /// case, is never skipped.
+    ///
+    /// Worth enabling when leanVM prover CPU is the bottleneck on co-located
+    /// aggregators. The cost is that a level in a given slot has few
+    /// producers, so a node that is down or late forfeits that slot's merge at
```
pablodeymo: The cost is larger than "a node that is down or late". Ownership is `duty_subnet % width == slot % width`, so a level can have zero owners while every configured node is healthy: with duty subnets {0, 2} at committee count 4, width 4 has no owner in odd slots, and since the full-width fallback is off under this flag, that merge level is dropped for the slot.

The rotation only guarantees an owner at every level when every subnet has an aggregator with that duty subnet. Worth stating that as a deployment precondition here and in the architecture doc, so an operator does not read the flag as safe on a sparse placement.

crates/blockchain/src/aggregation.rs
```diff
@@ -184,12 +184,106 @@ impl Message for EarlyAggregationCheck {
...
+///
+/// The width is therefore *not* uniform across the network: two aggregators
+/// on different duty subnets can derive different widths from one lopsided
+/// pool. That costs tiling precision (windows at different widths nest rather
+/// than tile) but no correctness, and it buys immunity to a single sparse
+/// wide proof, one validator in each of many subnets, collapsing every
+/// aggregator's window to the full committee set.
+///
+/// Only one session runs per slot (see [`snapshot_aggregation_inputs`]), and
+/// this candidate's own pool is still empty at that point: a produced
```
pablodeymo: This is an assumption about timing, not an invariant the code enforces. If a peer's aggregate for the current slot lands before this node's snapshot (clock skew, or a session that started early via `EarlyAggregationCheck`), the current-slot candidate gains an anchor and a width of 2. Under `--skip-redundant-aggregation` that means half the aggregators sit the current slot out, and their raw signatures miss the next block.

In the intended topology, distinct duty subnets with distinct subscriptions, a round-one peer proof never touches this node's subnet, so it does not bite there. It would with overlapping subscriptions. A sentence here (or in the PR description's cadence section) noting the assumption and when it fails would save the next reader the derivation.

crates/blockchain/src/aggregation.rs
MegaRedHand: Right on both counts, fixed in f33b308e.

The retry now returns early in the two cases where it is provably a no-op, and the counter only increments when the retry yields a job:

- `window.width() >= committee_count`: a window as wide as the committee admits every subnet, so `SubnetWindow::full` is the same lens. `>=` rather than `==` so a `committee_count` of 0 folds in too.
- empty proof pool: no child to admit at any width. This is the single-raw-signature case you named, and it is the ordinary shape of a minority or late vote.

What is left, a group whose pool is non-empty but whose windowed selection came up with one child, is exactly what the fallback exists for, so the retry there is genuine work. `resolve_job` can still decline it, hence counting recoveries rather than attempts.

Pulled the predicate out as `fallback_can_recover` so the three cases are unit-testable without a metrics registry. The metric's description and `docs/metrics.md` now say "recoveries, not attempts", which makes the "at or near zero on a well-tiled deployment" reading true.

bin/ethlambda/src/main.rs
MegaRedHand: Agreed, the silent reduction was the wrong call. 61e21ab9 adds `validate_aggregate_subnet_ids`, called from `run_node` right after the committee count resolves (it cannot live in clap, since the count comes from the flag *or* `validator-config.yaml`). Every id is checked, not just the first: only the first becomes the duty subnet, but all of them become subscriptions. That closes the pre-existing dead subscription too.

One deviation from your suggestion: I kept the modulo in `window_for_candidate` and its test. `AggregationWindowConfig` and `snapshot_aggregation_inputs` are `pub`, so the blockchain crate cannot lean on the binary's validation, and `SubnetWindow::new` reduces `start` regardless. Dropping the reduction only in `window_for_candidate` would leave the ownership test running on the raw value while the window ran on the reduced one, which is precisely the desync that test pins. So it is all-or-nothing, and three lines guarding a public API seemed the better side to land on. The comment there now says the binary is what actually enforces it.

bin/ethlambda/src/cli.rs
MegaRedHand: Confirmed your example: duty subnets {0, 2}, committee count 4, slot 1 gives `1 % 4 == 1`, which neither 0 nor 2 matches, so width 4 has no owner in odd slots and the fallback is off. The old wording framed that as a node being down or late, which is wrong.

8ac6dfc4 rewrites the flag's help and adds a paragraph to `docs/architecture.md`, stating the precondition explicitly: give every subnet below `--attestation-committee-count` an aggregator holding it as its duty subnet, otherwise leave the flag off. The loss is called structural rather than a consequence of an unhealthy node.

crates/blockchain/src/aggregation.rs
MegaRedHand: Fair, "always derives the narrowest width" claimed more than the code guarantees. 8ac6dfc4 softens it to "in practice" and adds the derivation you asked for: it is a timing expectation, a peer aggregate for the current slot landing before this node's snapshot (clock skew, or a session started early via `EarlyAggregationCheck`) gives the candidate an anchor and width 2, under `--skip-redundant-aggregation` the duty subnets that do not own width 2 sit the slot out and their raw signatures miss the next block, the intended topology keeps it true because a round-one peer proof never touches this node's subnet, and overlapping subscriptions are where it fails.

pablodeymo (approved):

